PR #48, opened 2019-08-29
Added the exclusion of WSAPlayer platform on Win32 plugin docs

pablomarcos:
I think there is an inconsistency on the documentation regarding the configuration of the Microsoft.MixedReality.WebRTC.Native.dll on the Unity inspector:

On Getting started->Unity tutorial->Importing MixedReality->WebRTC->Configuring the import settings the first picture shows on "Exclude Platforms" the WSAPlayer **checked ** for Win32/x86_64 library but later on the following table there is nothing written on "Exclude Platforms" for the same library:
Wrong_Doc


djee-ms:
Yes absolutely, thanks for the PR!

PR #49, opened 2019-08-29
Fixed 404 links in Readme Documentation section

adrianimajion:
The links to the Manual, Unity Tutorial, and API reference returned 404 pages so I updated them.


djee-ms:
Thanks, good catch @adrianimajion. But for me the last one is still wrong, should be https://microsoft.github.io/MixedReality-WebRTC/api/Microsoft.MixedReality.WebRTC.html no? Because https://microsoft.github.io/MixedReality-WebRTC/#api-documentation lands at the root of the docs instead.

adrianimajion:
Aha, I wasn’t sure where to point that and your URL makes much more sense. I’ll update the PR shortly.

adrianimajion:
Just updated with that link. Thanks for the catch!

djee-ms:
Merged. Thanks @adrianimajion!

Issue #36, opened 2019-08-14
Distorted video in TestAppUWP


EddieTheHead:
Remote video stream from hololens looks broken in TestAppUWP. The same stream works fine in Unity Editor. One can recognize the image, but it's replicated multiple overlapping times in the preview window. The attached image shows a logitech mouse. 

![2019-08-14 15_46_52-Microsoft MixedReality WebRTC (Running) - Microsoft Visual Studio](https://user-images.githubusercontent.com/7189283/63027052-836b2400-beac-11e9-9eac-787d7b65c516.png)

djee-ms:
Hi @EddieTheHead and thanks for logging this. Yes I observed that recently, this is a known issue I forgot to log. The most likely culprit is that the MediaPlayerSource used for the remote feed was randomly initialized to a default resolution of 640 x 480, and I did not update this code yet.

I incidentally wrote a change today (on a feature branch with other unrelated changes) that I believe would fix that. I tried with different resolutions and that seems to work in all cases for me:
https://github.com/microsoft/MixedReality-WebRTC/commit/25180a05c56b6a596bcae4fbf6be753e939cc844#diff-c1eacadf530048e585bc6e41ee0a887e

I am in the middle of some other change, could you please try to apply that fix and see if that solves it for you? If that's the case I'll cherry pick the change on `master`.

djee-ms:
@EddieTheHead I don't have access to a HoloLens at the moment but I pushed 545d6e6f0dd6c5ae5c34c3e259b175660d8a9524 which I am hoping is fixing that distorsion.

djee-ms:
Closed by mistake, waiting for your feedback.

djee-ms:
Confirmed this is also an issue for the local video feed, as the code currently assumes a resolution of 640x480 instead of checking the actual frame size of the incoming local video. There's even a TODO there 😀 
https://github.com/microsoft/MixedReality-WebRTC/blob/777396acc620d3cac7784882ca10e7d4e4cb6ac8/examples/TestAppUwp/MainPage.xaml.cs#L430

EddieTheHead:
Everything works great now. Thanks for quick fix :)

djee-ms:
Ok thanks for confirming @EddieTheHead. I am leaving that issue open until it is fixed also on the local side, which shall come with the video profile support soon.

PR #58, opened 2019-09-03
Initialize MF sources from actual frame format

djee-ms:
In TestAppUWP, ensure that the Media Foundation stream sources are
initialized, both for the local and the remote video, based on the
actual format of the video stream and not from a hard-coded resolution.
This prevent visual artifacts when resolution mismatches.

Bug: #36


djee-ms:
Need help on that one -- I am not familiar enough with Media Foundation, and this change although fixing #36 also introduces some 1~2 seconds black frames when re-opening the local video capture after the first start/stop cycle. It seems some MF pipeline resources are not cleaned-up properly on stop, and incur some delay on next start.

djee-ms:
This is already fixed on master. Abandoning.

PR #60, opened 2019-09-04
Add codec parameter override to SDP filtering

### LINKED ISSUE #51: More options for SDP and video settings
Hi,

I'm currently initiating calls with the SDP i receive from the LocalSdpReadytoSend, but as I understand, it has already been set as local description at that time. 
It would be beneficial if I could e.g. modify specific OPUS parameters as _Max Average Bitrate_, _Max Capture Rate_, _Max Playback Rate_ and so forth. Same goes for video parameters such as resolution and framerate, but as I understand, that's already [in the works](#46). 👍

Maybe CreateOffer() could have an optional input variable for specific SDP options?

djee-ms:
Allow the caller to specify a semicolon-separated list of key=value
pairs of codec arguments passed directly to the preferred codec during
SDP filtering, to configure it.

This is a low-level interface for advanced use only, will no validation
checks whatsoever. This fixes #51 at least short term.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/api.h
@@ -348,6 +348,19 @@ mrsPeerConnectionClose(PeerConnectionHandle* peerHandle) noexcept;
 // SDP utilities
 //
 
+/// Codec arguments for SDP filtering, to allow selecting a preferred codec and
+/// overriding some of its parameters.
+struct SdpFilter {
+  /// SDP name of a preferred codec, which is to be retained alone if present in
+  /// the SDP offer message, discarding all others.
+  const char* codec_name = nullptr;
+
+  /// Semicolon-separated list of "key=value" pairs of codec parameters to pass
+  /// to the codec. Arguments are passed as is without validation of their name
+  /// nor value.
+  const char* params = nullptr;
+};
+
 /// Force audio and video codecs when advertizing capabilities in an SDP offer.#
 ///
 /// This is a workaround for the lack of access to codec selection. Instead of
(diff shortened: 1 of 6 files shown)


djee-ms:
@eirikhollis how does that change look? I tested that the SDP offer message contains all arguments passed, provided that you are careful to pass the correct parameter name and valid values. This is a short-term low-level fix to allow configuring if you know what you're doing. I plan to add some higher level layer that simplifies this and doesn't force you to know the internals of SDP, but for now I think that should sort you out.

djee-ms:
Here's an example of SDP filtered message for OPUS, passing the extra parameters string `"ptime=40;maxaveragebitrate=8000"` :
```
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 maxaveragebitrate=8000;minptime=5;useinbandfec=1
a=ptime:40
```

Note that this only works with the one codec you select with `PreferredAudioCodec`, so this won't allow configuring multiple codecs in case you want to leave two choices to the receiver.

eirikhollis:
Looks good as an intermediate solution. 
I think you can place the ptime in the fmtp attribute together with the other parameters, e.g.:
```
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 minptime=20; ptime=40; cbr=0; maxplaybackrate=48000; sprop-maxcapturerate=48000; maxaveragebitrate=16000; stereo=0; sprop-stereo=0; useinbandfec=1
```

djee-ms:
The RFC [says no](https://tools.ietf.org/html/draft-spittka-payload-rtp-opus-03#section-6.2):
> The OPTIONAL media type parameters "ptime" and "maxptime" are mapped to "a=ptime" and "a=maxptime" attributes, respectively, in the SDP.

In any case the SDP message is generated by Google's code. The filtering I do is:
1. Parse the string message into in-memory data structures (Google)
2. Filter on the data structures <- my code
3. Re-generate a new string message from the data structures (Google)

So I don't generate the strings myself, I use the same code that generated the SDP message in the first place, only I feed it with different inputs.

eirikhollis:
Sounds awesome. Looking forward to testing it

PR #65, opened 2019-09-06
The name of the demo scene was wrong so I fixed it.

pablomarcos:
For me it was confusing the first time I tried the example scene so I just fixed the wrong scene name :)


djee-ms:
Thanks @pablomarcos!

PR #99, opened 2019-10-14
Remote audio feature

stephenatwork:
Remote audio can now be redirected away from the default audio device. 

RemoteAudioSource on Unity now plays audio using the OnAudioFilterRead API.

This PR also adds some a new high-level API which manages buffering and provides a way to consume the audio as a stream, resampling and/or changing the number of channels on the fly.

Open issues:
- Add proper interface to enable/disable individual tracks
- Refactor AudioTrackReadBuffer to buffer an individual track, and tidy up interface according to comments.
- Tidy up Unity integration.


eanders-MS:
Did you consider using [OnAudioFilterRead](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnAudioFilterRead.html)? It allows you to write the streaming audio data (after decoding it to WAVE_FORMAT_IEEE_FLOAT) directly to an `AudioSource`, bypassing the need for an `AudioClip`.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+        /// Internal adaptor from webrtc frame data to unity audio clip.
+        /// </summary>
+        class AudioData
+        {
+            float[] _buffer = null;
+            bool _isFull = false;
+            int _readPos = 0; // AudioClip fetches start from here
+            int _writePos = 0; // WebRTC data gets appended here
+            int _channelCount = 0;
+            int _sampleRate = 0;
```
djee-ms: Use an explicit `private` keyword please by convention, for clarity. And same below.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+            public Action<int, int> AudioFormatChanged = null;
```
djee-ms: I got caught already by not defining an explicit delegate type when the parameters are similar. I can guarantee people getting caught and swapping the sample rate and channel count in their handler. An explicit delegate doesn't provide compile-time guarantee, but at least will name the argument on auto-complete, which lowers the risk for users.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+            {
+                if (frame.sampleRate != _sampleRate
+                    || frame.channelCount != _channelCount
+                    || _buffer == null
+                    || _buffer.Length != (frame.sampleRate * frame.channelCount))
+                {
+                    // If anything changes, let's drop the buffer for simplicity (i.e. we won't resample or do extra work)
+                    _sampleRate = (int)frame.sampleRate;
+                    _channelCount = (int)frame.channelCount;
+                    _buffer = new float[frame.sampleRate * frame.channelCount];
```
djee-ms: So this implicitly buffers 1 second, right? Maybe add a comment about that?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+                    int srcSamples = (int)(frame.frameCount * frame.channelCount);
+                    // write1 is from _writePos forwards. write2 is the wraparound part (very often 0)
+                    int write1 = Math.Min(spaceAtEnd, srcSamples);
+                    for (int i = 0; i < write1; ++i)
+                    {
+                        _buffer[i + _writePos] = src[i] / 32768.0f; // 16 bit data is signed. Remap to -1,1
+                    }
+
+                    // Adjust the _writePos. Note that this may also update *_readPos*, if the reader is slow.
+                    // We will always enque ALL the data on the basis that newer is better.
```
djee-ms: "enqueue"

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+                    // Sometimes, if the reader isn't consuming fast enough, we may advance 'through' the read position.
+                    // This will give bad artifacts, so instead we'll drop some of the input instead.
```
djee-ms: Isn't that comment directly contradicting the one just above? Which is it, we discard new input or overwrite old non-consumed one?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -40,6 +40,163 @@ public class RemoteAudioSource : AudioSource
...
+                    for (int i = totalRead; i < data.Length; ++i)
+                    {
+                        data[i] = 0;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Local storage of audio data to be fed to _audioClip;
```
djee-ms: Use `<see cref="_audioClip">` for docs linking. VS auto-completes that.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+                }
+            });
+        }
+
+        private void OnAudioFormatChanged(int sampleRate, int channelCount)
+        {
+            lock (this)
+            {
+                _audioClip = null;
+                // TODO Race condition here. AudioSource.clip still refers to the old clip. We can't modify it (not main thread).
```
djee-ms: What's the plan to fix that?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+                int lengthSamples = sampleRate / 10; // Buffer up to 100ms of data.
+                _audioClip = AudioClip.Create("RemoteClip", sampleRate, channelCount, sampleRate, true, _audioData.OnClipRead);
+                _audioClipEnqueuedCreate = false;
+                var aud = GetComponent<UnityEngine.AudioSource>();
+                if (aud)
+                {
+                    aud.clip = _audioClip;
+                    aud.volume = 1.0f;
+                    aud.loop = true;
+                    aud.bypassEffects = true;
```
djee-ms: Arguable, no? What's the rationale for preventing the user from adding audio effects?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+            _mainThreadWorkQueue.Enqueue(() =>
+            {
+                int lengthSamples = sampleRate / 10; // Buffer up to 100ms of data.
+                _audioClip = AudioClip.Create("RemoteClip", sampleRate, channelCount, sampleRate, true, _audioData.OnClipRead);
+                _audioClipEnqueuedCreate = false;
+                var aud = GetComponent<UnityEngine.AudioSource>();
+                if (aud)
+                {
+                    aud.clip = _audioClip;
+                    aud.volume = 1.0f;
```
djee-ms: Good for testing, but we shouldn't modify this for the user.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+                    aud.loop = true;
+                    aud.bypassEffects = true;
+                    aud.Play();
```
djee-ms: Even that one I am not sure if we need to do it automatically. Maybe the user wants an `AudioClip` ready, but wants to control the timing of actually starting playback (based on some animation, or anything else...). Thoughts?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+            // _audioData buffers the audioframes which arrive via WebRTC callbacks.
+            // WebRTC will sometimes change the format midstream, and if so, we update to match the new
+            // format immediately and will need to create a new AudioClip instance with the new parameters.
+            // Note we can't change AudioClip immediately since this is not the unity main thread.
+            //
+            // _audioClip is the unity object which can be attached to various audio emitters.
+            // The audioclip in this class is created with callbacks which read the data from _audioBuffer.
+            if (_audioClipEnqueuedCreate)
+            {
+                return; // waiting for creation.
```
djee-ms: Can't we start buffering inside a pure C# object while waiting to create the Unity wrapper (`AudioClip`)? It seems we will take some latency on start here that we could avoid.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -173,9 +332,73 @@ private void TrackRemoved(WebRTC.PeerConnection.TrackKind trackKind)
...
+                return;
+            }
+            lock (_audioData)
+            {
+                switch (frame.bitsPerSample)
+                {
+                    case 16:
+                        _audioData.AppendAudio16(frame);
+                        break;
+                    case 8:
```
djee-ms: Add TODO?

djee-ms:
> Did you consider using [OnAudioFilterRead](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnAudioFilterRead.html)? It allows you to write the streaming audio data (after decoding it to WAVE_FORMAT_IEEE_FLOAT) directly to an `AudioSource`, bypassing the need for an `AudioClip`.

I saw that technique yesterday in [that Gamasutra article](https://www.gamasutra.com/blogs/JoeStrout/20170223/292317/Procedural_Audio_in_Unity.php) and was wondering about the difference with the streaming `AudioClip`. @eanders-MS do you have any experience with this? How do the two compare? It sounds (pun intended) like `OnAudioFilterRead` might be better for performance and/or latency, no? And that avoids the issue with Unity object access when changing channel count and sample rate I assume, at the expense of re-sampling since you seem to be limited to the internal sampling rate of the Unity DSP (but anyway Unity will do that resampling on the `AudioSource` anyway I am pretty sure).

Ping @stephenatwork, what do you think? The docs explicitly mention the technique for procedural audio:
> If this is the first filter in the chain and a clip isn't attached to the audio source, this filter will be played as the audio source. In this way you can use the filter as the audio clip, procedurally generating audio.

eanders-MS:
I was hoping to sign up for this work but @stephenatwork beat me to it :)

>  @eanders-MS do you have any experience with this? How do the two compare?

My prior experience is one-sided in that I've only used the `OnAudioFilterRead` approach. On a previous project I incorporated a streaming mp3 player into Unity. It would read output from ffmpeg and pipe it to an `AudioSource` via `OnAudioFilterRead`. There was some timing code to ensure the video stream stayed in sync to the audio. I think the same approach is applicable here, given the similarities.

That said, if the solution in this PR works and it's a good Unity integration, then fantastic.

stephenatwork:
> Did you consider using [OnAudioFilterRead](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnAudioFilterRead.html)? It allows you to write the streaming audio data (after decoding it to WAVE_FORMAT_IEEE_FLOAT) directly to an `AudioSource`, bypassing the need for an `AudioClip`.

Thanks for the tip. That looks like a much better choice than the one I found. I did a few local tests and it seems to have very little delay (which is an issue in the AudioClip version). It shouldn't be much work to port the implementation, except to add the resampling.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
stephenatwork: New API hides this detail

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
stephenatwork: New API hides this detail

stephenatwork:
I've updated this PR to use the OnAudioFilterRead API.

Also, managing buffers and resampling in C# seems inefficient. So this PR adds some a new high-level API which manages buffering and provides a way to consume the audio as a stream, resampling and/or changing the number of channels on the fly.

libs/Microsoft.MixedReality.WebRTC.Native/src/api.cpp
```diff
@@ -1075,6 +1075,34 @@ void MRS_CALL mrsPeerConnectionRemoveLocalAudioTrack(
   }
 }
 
+
+
+void MRS_CALL mrsAudioReadStreamCreate(PeerConnectionHandle peerHandle,
```
djee-ms: API functions like this generally return an `mrsResult` so that we can transform errors into exceptions in C# for example.

libs/Microsoft.MixedReality.WebRTC.Native/src/api.cpp
```diff
@@ -1075,6 +1075,34 @@ void MRS_CALL mrsPeerConnectionRemoveLocalAudioTrack(
...
+                                       int bufferMs,
+                                       AudioReadStreamHandle* audioBufferOut) {
+  *audioBufferOut = nullptr;
+  if (auto peer = static_cast<PeerConnection*>(peerHandle)) {
+    *audioBufferOut = new AudioReadStream(peer, bufferMs);
```
djee-ms: So far in the API all objects are referenced counted and held alive by C++. The alternative here is to not reference count and use raw pointers. Both rely on the wrapper layer (C#) calling `mrsAudioReadStreamDestroy()` but the latter makes sharing and debugging much easier. Do we think that it's not worth the complication here because the audio read stream is uniquely owned by a single C# object?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
```diff
@@ -508,4 +508,36 @@ class PeerConnection : public webrtc::PeerConnectionObserver,
   PeerConnection& operator=(const PeerConnection&) = delete;
 };
 
+class AudioReadStream {
+ public:
+  AudioReadStream(PeerConnection*, int bufferMs);
+
+  int Read(int sampleRate, float data[], int dataLen, int numChannels) noexcept;
```
djee-ms: +docs (in particular : what is the `int` return value?)

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -981,6 +981,36 @@ public void RemoveLocalAudioTrack()
...
+            }
+
+            public void ReadAudio(int sampleRate, float[] data, int channels)
+            {
+                AudioReadStreamInterop.Read(_nativeStreamHandle, sampleRate, data, data.Length, channels);
+            }
+
+            public void Dispose()
+            {
+                AudioReadStreamInterop.Destroy(_nativeStreamHandle);
```
djee-ms: Add `GC.SuppressFinalize(this);` and add some finalizer (dtor) calling `Destroy()` too, to avoid any leak? I'm still unclear in C# if this is better practice or if the user not calling `Dispose()` is something that should not be handled.

libs/Microsoft.MixedReality.WebRTC.Native/src/api.cpp
stephenatwork: Yes, given that reads are inherently destructive, I don't expect these objects to be shared. I can ensure the handle is nulled in the C# object also to avoid double delete.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
stephenatwork: TBD - some kind of status such as: 
* normal read (success)
* underflow (audio didn't arrive from webrtc in time, )
* overflow (reads happening too slowly)

stephenatwork: I'll remove the return value for now. I think it may be useful to revisit and add a way for the caller to get a status somewhat like below. Perhaps with millisecond values instead of number of samples. This would let the app show some UI if the connection was going bad in some way. 

struct Status {
 int samplesRead;
 int samplesDropped;
 int samplesMissing;
}

stephenatwork:
All the scaffolding is in place, we just need to choose and wire up a resampler. Luckily webrtc seems to include at least 2 which I'll need to evaluate:  
webrtc\xplatform\opus\silk
webrtc\xplatform\webrtc\common_audio\resampler\include

libs/Microsoft.MixedReality.WebRTC/Interop/AudioReadStreamInterop.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.MixedReality.WebRTC.Interop
+{
+    internal class AudioReadStreamInterop
+    {
+        [DllImport(Utils.dllPath, CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Ansi,
+            EntryPoint = "mrsAudioReadStreamCreate")]
+        public static extern uint Create(PeerConnectionHandle peerHandle, int bufferMs, ref IntPtr audioReadStream);
```
stephenatwork: @djee-ms is this the correct pattern? Accept a safehandle & the native method has void*?

djee-ms: Yes. But you probably want to mark `audioReadStream` as `out`, not `ref`, no? It's the returned handle to the newly allocated object, isn't it? So it's not valid until returned.

stephenatwork:
The current branch supports (only) spatial audio. Objects with a RemoteAudioSource component, can have an AudioSource attached. Remote audio tracks are not piped to the output device. When there is no remote audio data, either because of underrun or not yet connected, the RemoteAudioSource outputs a low buzz for debugging.
TODO: remove the dialtone, allow per track routing of output to speakers or AudioSource.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -453,6 +453,10 @@ MRS_API void MRS_CALL mrsPeerConnectionRegisterRemoteAudioFrameCallback(
     PeerConnectionAudioFrameCallback callback,
     void* user_data) noexcept;
 
+// Experimental. Play or not remote audio tracks on the audio device.
+MRS_API void MRS_CALL
+mrsPeerConnectionPlayRemoteAudio(PeerConnectionHandle peerHandle, bool play);
```
stephenatwork: Should we call this 'RenderRemoteAudio' (sending data to device), not 'Play' (possible confusion vs paused)?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -479,6 +496,13 @@ class PeerConnectionImpl : public PeerConnection,
   /// https://stackoverflow.com/questions/43788872/how-are-data-channels-negotiated-between-two-peers-with-webrtc
   bool sctp_negotiated_ = true;
 
+  // TODO
```
stephenatwork: nit: description, not just bare todo

stephenatwork:
I would approve but ...
image

drejx:
Hi @stephenatwork,

These look like some great changes!   Funny enough I'm looking at an audio problem where all audio volume is severely reduced when a webrtc session is initiated (all audio except for the webrtc audio).  What I've found is that the Google webrtc implementation selects the default audio _communications_ device for audio output and when that device gets used then Windows ducks/reduces all other audio by 80% (for anyone else reading this, it's a Windows feature to ensure speech is heard above everything else).  Problem is if you're building something like a game then all the game audio is suppressed by 80% as well 😞 

So this feature you've worked on looks to actually solve that for Unity, correct?  Could you help me understand some aspects of how it works, specifically:

1. With your changes will the incoming/remote webrtc audio get piped into Unity via your new RemoteAudioSource Component?
2. Does that mean all the webrtc audio will not go through Windows Core audio directly, but instead be piped to the Unity audio pipeline and get output to the Windows default audio device (i.e. **not** the Windows default _communications_ device)?
3. With this change will we be able to independently change the webrtc audio volume via Unity's AudioSource.volume property, but through your RemoteAudioSource?
4. Is there an ETA when it will be merged to the master branch?  :)
5. My understanding was that since Google's libwebrtc Windows implementation sends audio directly to the default audio communications device that the only way to change that was by modifying the libwebrtc code, recompiling and re-integrating into MixedReality.  How were you able to "circumvent" that paradigm?

Thank you for all your hard work!

Cheers,
Andrej

fibann:
Hi @drejx, I had noticed the same behavior but hadn't looked into it, thanks for investigating!

This change uses a custom AudioMixer but doesn't change the AudioDeviceModule, so I am afraid it won't solve the problem (I suppose that even if no audio is played on it, the default comm device will still be in use). I'll open a new issue for it.

I'll try to reply to your questions with what I know:
1. Yes
2. Yes (AFAIK), but this won't avoid the volume drop, as said above.
3. I am not sure what you're asking - I think you will be able to change the volume through AudioSource.volume but I haven't tried. @stephenatwork can you confirm?
4. Not sure - in order to work on UWP this relies on some other changes that are still in flux. Would it still be useful for you if it only worked on Desktop?
5. I am not sure what you mean by "the MixedReality-SDK from 2019-12-16". If you mean "is it possible to integrate these changes to a previous master/release", it's not really recommended, but I can't think to any obvious blocker so feel free to try.
6. We weren't, as said above.

stephenatwork:
Thanks the interest @drejx - @fibann has taken over this feature.

> 3. I am not sure what you're asking - I think you will be able to change the volume through AudioSource.volume but I haven't tried. @stephenatwork can you confirm?

My understanding is that the unity.audiosource is the correct (indeed, only!) place to change the volume.

Because the reduction appears to be a system setting (see below), I expect there may be no way to bypass the setting. I agree we'd want a custom ADM. So an option to use the default output device as well as an option for a custom device id sounds like a great idea.
image

djee-ms:
Yes @stephenatwork but I think @drejx is right that at least in theory opening the audio endpoint as a non-communication device should prevent this behavior. Now I have no idea if/how easy it is to do that. @fibann had a look and from what I understand from above it requires some changes to the Google implementation, either by changing their code or by injecting a custom audio device module that replaces the built-in one. Either way this requires some work.

drejx:
Thanks for the detailed responses everyone :)  In my case I am using it with Unity Desktop/Standalone and hence wouldn't need it to work for UWP.

Correct me if I'm wrong, but as I understand it even with the Remote Audio Feature the voice audio will still be piped through the Windows default communications device (due to Google's webrtc implementation) and also piped to your new RemoteAudioComponent to Unity.  So in effect it would kinda negate the Remote Audio feature. For example, changing the volume/muting in Unity won't affect the audio going through the communications device and all audio will still be ducked (including RemoteAudio).  I'm wondering if it will make the experience worse because you'll be hearing the voice audio coming from 2 sources.  In your tests have you noticed this beahviour?  I don't mean to be a drag, but want to highlight possible pitfalls before you get 300 issues/bugs titled "audio doesn't work!"  :P 

It would be great for the Google webrtc logic to be able to select a device, but I understand that is tough to do in your situation as it would require a new fork of their code.  @fibann thanks for creating #263.  I'll post there what I've found in my digging of the code, but the short answer is in `audio_device.h` there is this enum for which device to select and `kDefaultCommunicationDevice` is what the `AudioDeviceWindowsCore()` is initialized with:

```
    enum WindowsDeviceType {
        kDefaultCommunicationDevice = -1,
        kDefaultDevice = -2
      };
```

Keep on truckin'!
Andrej

fibann:
@drejx
> even with the Remote Audio Feature the voice audio will still be piped through the Windows default communications device (due to Google's webrtc implementation) and also piped to your new RemoteAudioComponent to Unity.

Not really, with this change the audio data from remote tracks is "redirected away" from the WebRTC audio device (i.e. the Windows default comm device) so you won't hear the audio twice (for the sake of clarity, note that this behavior needs to be chosen explicitly by calling `RenderRemoteAudioTrack(false)`). On the other hand, even if all tracks are redirected away, the device will still be opened and empty frames (silence) will be sent to it, so volume from all the other apps will still be reduced when the session begins. I hope this clarifies the issue.

If you think this is a useful change even if it only works on Desktop I think we can push it as it is (only minor changes should be left) so hopefully it can go in in the next few days.

drejx:
Hi @fibann,

Got it, thank you.  I think it's quite useful for Unity/Desktop, even though it will still duck other sounds.  A major plus is being able to change the volume so at least the user can adjust voice chat volume to match the other ducked game/app audio :)

I'm exploring changing the Google webrtc code to switch from `kDefaultCommunicationDevice ` -> `kDefaultDevice`.  In your opinion what would cause the least friction to get a custom built webrtc.lib into the MixedReality-WebRTC SDK?  I believe the MixedReality project uses the output of the [webrtc-uwp-sdk](https://github.com/webrtc-uwp/webrtc-uwp-sdk) project for it's webrtc library and I can build my own webrtc.lib by following the instructions here:

https://github.com/webrtc-uwp/webrtc-uwp-sdk#building-the-sdk

Once built I can replace the MixedReality webrtc.lib in here:

MixedReality-WebRTC\packages\Microsoft.MixedReality.WebRTC.Native.Core.Desktop.x64.Release.1.0.2\lib\native\lib\win10\x64\release\webrtc.lib

Would that be sufficient or would Org.WebRtc.WrapperGlue.lib also need to be copied?

Oops, didn't meant to hijack this issue thread!  I'd be happy to open a new one if you wish.

Cheers,
Andrej

djee-ms:
@drejx if you build locally the "core dependencies" (the Google code) following [this doc](https://microsoft.github.io/MixedReality-WebRTC/manual/building-core.html#build-steps) then it automatically takes precedence over the `webrtc.lib` from the NuGet packages when you use the Visual Studio solution, no need to copy anything or overwrite any file. There's already a `build.ps1` script also to help compiling, you "only" need to make sure the environment and all tools/dependencies are setup on your dev machine, as explained in that doc.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1061,6 +1061,50 @@ public void RemoveLocalAudioTrack()
...
+                if (this._nativeStreamHandle != IntPtr.Zero)
+                {
+                    AudioReadStreamInterop.Destroy(_nativeStreamHandle);
+                    this._nativeStreamHandle = IntPtr.Zero;
+                }
+            }
+        }
+
+        /// <summary>
+        /// High level interface for consuming WebRTC audio streams.
```
djee-ms, line 1098: This doesn't say much about why one would use that and what features it enables. Can we add an example of feature to show why/how to use an audio read stream?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1202,6 +1230,29 @@ void PeerConnectionImpl::OnAddTrack(
   const std::string& trackKindStr = track->kind();
   if (trackKindStr == webrtc::MediaStreamTrackInterface::kAudioKind) {
     trackKind = TrackKind::kAudioTrack;
+
+    if (custom_audio_mixer_) {
+      // We need to get the ssrc of the receiver in order to match the track to
```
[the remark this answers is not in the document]
fibann: Yeah I forgot this part actually - I tried to call `GetSources` (here and also at the point where an audio frame is received) and it seems to always return an empty vector. I tried to dig in the code a bit but couldn't figure out if it's broken or I am misusing it, so just went for getting the stats instead. I'll clarify this in the comment.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -622,6 +630,21 @@ MRS_API mrsResult MRS_CALL mrsPeerConnectionRemoveLocalVideoTrack(
 MRS_API void MRS_CALL mrsPeerConnectionRemoveLocalAudioTrack(
     PeerConnectionHandle peerHandle) noexcept;
 
+MRS_API mrsResult MRS_CALL
+mrsAudioReadStreamCreate(PeerConnectionHandle peerHandle,
+                         int bufferMs,
```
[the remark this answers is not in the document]
stephenatwork: I'm not sure bufferMs should even be in the API. (I added it originally because I needed to test different values).

While on the topic of buffering, it's worth mentioning an issue if there's a hiccup (delay) in unity. There is no logic for 'catching up' and so in the worst case, and even with a great connection, the buffer can be completely full and all it's doing is adding latency. Half a second by default!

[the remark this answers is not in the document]
fibann: > There is no logic for 'catching up'

This can be solved in `AudioReadStream::Read` by taking the newest n frames that fill the requested buffer, rather than taking the oldest ones, right? It seems that should be easy to do - or am i missing some fundamental issue?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1366,4 +1417,184 @@ void PeerConnection::GetStats(webrtc::RTCStatsCollectorCallback* callback) {
...
+AudioReadStream::~AudioReadStream() {
+  peer_->RegisterRemoteAudioFrameCallback(AudioFrameReadyCallback{});
+}
+
+AudioReadStream::Buffer::Buffer() {
+  resampler_ = std::make_unique<webrtc::Resampler>();
+}
+AudioReadStream::Buffer::~Buffer() {}
+
+void AudioReadStream::Buffer::addFrame(const Frame& frame,
```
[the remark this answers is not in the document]
fibann: @stephenatwork IIRC you were saying that the first passes are probably dwarfed by resampling (and all of this by video processing)?

Anyway I agree there is a lot of space for optimization here (and that it doesn't need to be tackled now).

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
```diff
@@ -398,4 +405,74 @@ class PeerConnection : public TrackedObject {
...
+ public:
+  /// Create a new stream which buffers 'bufferMs' milliseconds of audio.
+  /// WebRTC delivers audio at 10ms intervals so pass a multiple of 10.
+  /// Or pass -1 for default buffer size (currently 0.5 seconds).
+  AudioReadStream(PeerConnection* peer, int bufferMs);
+  ~AudioReadStream();
+
+  /// Fill data with samples at the given sampleRate and number of channels.
+  /// If the internal buffer overruns, the oldest data will be dropped.
+  /// If the internal buffer is exhausted, the data is padded with white noise.
```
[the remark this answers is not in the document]
fibann: From https://github.com/microsoft/MixedReality-WebRTC/pull/99#issuecomment-567525122 this should be debugging only.

@stephenatwork any reason to pad with noise rather than simply silence here?

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.MixedReality.WebRTC
+{
+    /// <summary>
+    /// High level interface for consuming WebRTC audio streams.
```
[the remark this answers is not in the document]
fibann: +1. What about AudioTrackBuffer?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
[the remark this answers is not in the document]
stephenatwork: Yes, these passes are a tiny fraction of frame time. The existing code is pretty dumb because it handles some unlikely paths.

One easy optimization to make is that for headsets, the most common transform is mono16 -> stereo f32.
So a dedicated code path for mono -> stereo could avoid some copies and half the resampling.

One other thing is that these simple loops tend to auto-vectorize very well if the compiler can be given enough hints (e.g. round buffer sizes/loop counts up to 8 or 16).

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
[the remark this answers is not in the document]
stephenatwork: I think the issue is that if you always take the latest, then you run the risk of dropouts which is what the buffering is trying to avoid! So the idea is that you're willing to trade N ms of delay for reduced probability of dropouts.

You can do this by skipping some part of the buffer, but it's likely (?) that creates pops/artifacts. Maybe there's something better? Perhaps temporarily speed up the audio by messing with the sample rates, or something more complex which doesn't change the pitch.

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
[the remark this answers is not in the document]
djee-ms: Not too bad. Do we want to keep the `Read` term, so `AudioTrackReadBuffer`?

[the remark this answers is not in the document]
fibann: Yeah I guess we might have a WriteBuffer too at some point 😄

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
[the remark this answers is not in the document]
stephenatwork: It was @djee-ms suggestion to help debugging, since there are many states which can lead to silence. It was moderately useful to experimentally choose a buffer size. I suppose it may be more useful to have underrun/overrun status available programmatically.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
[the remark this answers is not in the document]
fibann: Yes, you're right (also if frames are pushed more frequently than when they are pulled that wont' work at all, but it shouldn't be our case).

Are we sure this is a problem though? I would expect OnAudioFilterRead to request more data after a hiccup to get up to speed.

In any case I have looked at how WebRTC handles audio packets buffering, doing it properly (and generically) is complicated unsurprisingly (see modules/audio_coding/net_eq/net_eq.cc). Though they have an `Accelerate` class that we might use directly (modules/audio_coding/net_eq/accelerate.h). Looks like some work so I'd be for logging this separately (if it is indeed an issue that can happen).

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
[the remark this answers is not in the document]
fibann: We can have `Read` return the actual consumed samples then.

Also I'd say we can make sine padding opt-in (easier than leave it to higher layers and not too cumbersome).

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -3,9 +3,9 @@
 
 #pragma once
 
-#include "audio_frame.h"
 #include "export.h"
 #include "result.h"
+#include "audio_frame.h"
```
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -450,6 +453,11 @@ MRS_API void MRS_CALL mrsPeerConnectionRegisterRemoteAudioFrameCallback(
     PeerConnectionAudioFrameCallback callback,
     void* user_data) noexcept;
 
+// Experimental. Render or not remote audio tracks on the audio device.
```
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1366,4 +1417,184 @@ void PeerConnection::GetStats(webrtc::RTCStatsCollectorCallback* callback) {
...
+    const AudioFrame& frame) {
+  auto ars = static_cast<AudioReadStream*>(user_data);
+  ars->audioFrameCallback(frame.data_, frame.bits_per_sample_, frame.sampling_rate_hz_,
+                          frame.channel_count_, frame.sample_count_);
+}
+
+AudioReadStream::AudioReadStream(PeerConnection* peer, int bufferMs)
+    : peer_(peer),
+      buffer_ms_(bufferMs >= 10 ? bufferMs : 500 /*TODO good value?*/) {
+  peer->RegisterRemoteAudioFrameCallback(
```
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// The implementation builds on top of the low-level AudioFrame callbacks
+    /// and handles all buffering and resampling.
+    /// </summary>
+    public interface IAudioReadStream : IDisposable
```
fibann: @stephenatwork @djee-ms it's a bit inconsistent that we use an interface here when the rest of the public API uses classes. Using more interfaces is not necessarily a bad thing _if_ we plan to move the whole API in that direction, otherwise I'd stay consistent and just use a class with public+internal methods. Thoughts?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1061,6 +1061,50 @@ public void RemoveLocalAudioTrack()
             PeerConnectionInterop.PeerConnection_RemoveLocalAudioTrack(_nativePeerhandle);
         }
 
+        class AudioReadStream : IAudioReadStream
+        {
+            IntPtr _nativeStreamHandle = IntPtr.Zero;
```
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -58,10 +68,16 @@ public void Play()
             if (!IsPlaying)
             {
                 IsPlaying = true;
-                //PeerConnection.Peer.RemoteAudioFrameReady += RemoteAudioFrameReady;
+                OnAudioConfigurationChanged(false);
```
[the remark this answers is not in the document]
fibann: Fixed

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
stephenatwork: I think clarity of intent trumps consistency. Thus would suggest using interfaces wherever possible, even if it's not everywhere.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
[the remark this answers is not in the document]
stephenatwork: I didn't look in to the pipeline of buffers, but would expect OnAudioFilterRead to be much more dumb/realtime and always ask for the same number of samples. I think it's only a short hop from there to handing to the output device.

Absolutely, it's complex and can be deferred. Perhaps there is some simple adaptive scheme we can use to choose the buffer size based on network conditions. Or perhaps webrtc is already doing all the smart stuff and we can reduce bufferMs to 20 or 30.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h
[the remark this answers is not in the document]
djee-ms: I'd rather have it opt-out. Better have strong indications that something is wrong, and let user ignore them (opt out and get silence instead), than hide a potential issue from the user without them knowing.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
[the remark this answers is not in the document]
djee-ms: I don't see any change here.

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
djee-ms: I kind of agree with @fibann, although inconsistency bothers me less than the fact that we are using an interface for hiding the implementation, without any intent of having more than 1 of them. C# is already pretty good at separating public/private API and exposing the public symbols of an assembly alone to the developers (unlike C/C++ headers), so there's no good reason to abstract that further with an interface, unless there's an intent to abstract part of the `AudioTrackReadBuffer` implementation only (if it had multiple functions; it doesn't) or expect another implementation (I don't think we do).

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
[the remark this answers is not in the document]
djee-ms: Ok that refreshed, sorry.

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
fibann: > C# is already pretty good at separating public/private API and exposing the public symbols of an assembly alone to the developers

Yes, for this reason I think that using an interface as opposed to a class doesn't add much in terms of clarity of intent from the point of view of external users.

> there's no good reason to abstract that further with an interface

IMHO there are good reasons, namely you avoid dependencies on instances of a specific class in the user code, which makes refactoring easier if later you want to split it in multiple classes or merge it with another class. It's a way to write the API that has its merits - OTOH I don't think it's suited to this class more than to the rest of the existing API so I wouldn't use it just here.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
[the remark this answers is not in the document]
fibann: I'll mark the API as experimental then and log an issue to investigate this more.

libs/Microsoft.MixedReality.WebRTC/IAudioReadStream.cs
fibann: It is now a public class.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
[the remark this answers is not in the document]
fibann: Frames are going to be relatively small (960B for 10ms of 16bit 48KHz), so multiple passes shouldn't be an issue. I have reordered the code a bit so that we do the resampling on just one channel when possible, that should be good enough.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
```diff
@@ -58,14 +69,20 @@ public void Play()
             if (!IsPlaying)
             {
                 IsPlaying = true;
-                //PeerConnection.Peer.RemoteAudioFrameReady += RemoteAudioFrameReady;
+                OnAudioConfigurationChanged(deviceWasChanged: false);
+                _audioTrackReadBuffer = PeerConnection.Peer.CreateAudioTrackReadBuffer();
             }
         }
```
fibann, line 75: @stephenatwork @djee-ms it feels weird that we can control play state on the linked AudioSource _and_ on this object independently. Looks like an easy source of mistakes. Also we need to decide a sane behavior for when the AudioSource plays but this doesn't (at the moment you get beeping).

I propose to
1. create/destroy the read buffer automatically when a track is added/removed
2. return empty frames in OnAudioFilterRead if there is no track
3. remove Play/Stop/IsPlaying from this class, or make them a shortcut for `_linkedAudioSource.(Play|Stop|IsPlaying)`

Thoughts?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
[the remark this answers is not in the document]
fibann: > I didn't look in to the pipeline of buffers, but would expect OnAudioFilterRead to be much more dumb/realtime and always ask for the same number of samples.

That seems what's happening actually 😞

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
stephenatwork: Sounds good to me. I suppose that there is a small overhead even when paused via the unity AudioSource, because IIRC the  buffering/transcoding still happens internally, even if we're not consuming the data via OnAudioFilterRead. Not sure if it's worth doing anything about that.

fibann: At the moment we transcode only on read so we don't have this issue*. For the sake of correctness though I suppose we can override OnEnable/OnDisable to turn the buffer on/off.

*Maybe we _shouldn't_ wait for a Read() call (I am not sure if the processing adds any meaningful latency) but we can deal with this later.

djee-ms: Yes, we should continue to transcode on read because this means transcoding inside the audio thread, which is the right place to do so, instead of some other thread like the WebRTC signaling thread or the main Unity app thread, which are busy doing other stuffs. And yes we should return silence when there's no track, to be consistent with the WebRTC behavior of sending silence when there's no track on a transceiver. I also agree on removing the playing state from here and use the audio source one only to avoid confusion.

libs/mrwebrtc/src/interop/peer_connection_interop.cpp
```diff
@@ -69,3 +70,52 @@ mrsPeerConnectionAddTransceiver(mrsPeerConnectionHandle peer_handle,
...
+  *audioBufferOut = nullptr;
+  if (auto peer = static_cast<PeerConnection*>(peerHandle)) {
+    *audioBufferOut = new AudioTrackReadBuffer(peer, bufferMs);
+    return Result::kSuccess;
+  }
+  return Result::kInvalidNativeHandle;
+}
+
+mrsResult MRS_CALL
+mrsAudioTrackReadBufferRead(AudioTrackReadBufferHandle readStream,
```
stephenatwork, line 103: nit: readStream -> readBuffer or readHandle

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/RemoteAudioSource.cs
stephenatwork: Filippo - when I say on read I mean OnAudioFilterRead. Do you mean something else? IIRC we're transcoding when the frame arrives from webrtc (addFrame), not when requested from OnAudioFilterRead (that part is a memcpy). The transcoding is the extra work I mean if the unity source is paused, but data continues to arrive from webrtc.

Issue #107, opened 2019-10-22
H.264 video quality very low for a long time after startup


fibann:
By default the H264 encoder code uses the min bitrate as starting bitrate, which leads to very poor image quality at the beginning of the stream.  The problem is compounded by the fact that:
1. the bitrate is changed through stopping the stream, resetting the whole Media Foundation pipeline and restarting the stream; this takes a long time and causes visible stuttering
2. the min interval between successive resets is set to 15s (supposedly to avoid a lot of stuttering in one go) which has the side effect of making significant rate changes take a very long time.

There are several ways we can address this:
- avoiding a full pipeline reset by setting https://docs.microsoft.com/windows/desktop/DirectShow/avenccommonmeanbitrate-property instead when only a bitrate change is needed (which is most times)
- initialize the new pipeline while the old one is still encoding (not sure if this is possible or there are limitations, especially with hardware encoders)
- expose the start/min bitrate through the native API and have the user/high-level API set it to a higher value (depending on e.g. the resolution).

eirikhollis:
@djee-ms @fibann 
I've been trying to modify the min/start/max bitrate to varying values, and can't seem to fix the blcokiness of the video sent from HL. I'm referring to the `PeerConnection.SetBitrate` method which addresses the following sentence mentioned above
```
expose the start/min bitrate through the native API and have the user/high-level API set it to a higher value (depending on e.g. the resolution).
```
I notice a significant lack in quality when setting min/start/max to 0.1mpbs, and some increase when setting min to 0.1mbps, start to 1mbps and max to 10mbps. 
I've uploaded a screenshot from a 720p30 stream with min/start/max all set to 10mbps:
![image](https://user-images.githubusercontent.com/30655189/68123734-0df07c00-ff0e-11e9-850f-4c5249113fcf.png)

Is this normal? Or am I experiencing something a lot more severe when what the issue is?
Both connection endpoints are located with 3 meters of a powerful router and I've tried varying webcameras. The issue is not present in Unity using VP8. Only H264 on UWP.

djee-ms:
Yes H.264 hardware encoding still has issues on UWP and we're not sure yet why the encoder sometimes seems to struggle, and some other times works fine. Although 0.1mpbs is probably too low, with 10Mbps your blockiness seems more severe than what we usually see. What device are you using? Is it some PC/Laptop or HoloLens 1? How is the CPU load on your app? Can you reproduce the same issue with TestAppUWP, or only with your app?

[...]

PR #109, opened 2019-10-23
 Expose PeerConnectionInterface::SetBitrate 

fibann:
Add usage example in TestAppUWP.
Mitigates #107.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
@@ -601,6 +601,20 @@ mrsPeerConnectionCreateOffer(PeerConnectionHandle peerHandle) noexcept;
 MRS_API mrsResult MRS_CALL
 mrsPeerConnectionCreateAnswer(PeerConnectionHandle peerHandle) noexcept;
 
+/// Set the bitrate allocated to all RTP streams sent by this connection.
+/// Other limitations might affect these limits and are respected (for example
+/// "b=AS" in SDP).
+///
+/// Setting |start_bitrate_bps| will reset the current bitrate estimate to the
+/// provided value.
+///
+/// If any of the arguments has a negative value, it will be ignored.
+MRS_API mrsResult MRS_CALL
+mrsPeerConnectionSetBitrate(PeerConnectionHandle peer_handle,
+                            int min_bitrate_bps,
+                            int start_bitrate_bps,
+                            int max_bitrate_bps) noexcept;
+
 /// Set a remote description received from a remote peer via the signaling
 /// service.
 MRS_API mrsResult MRS_CALL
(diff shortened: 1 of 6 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
```diff
@@ -601,6 +601,20 @@ mrsPeerConnectionCreateOffer(PeerConnectionHandle peerHandle) noexcept;
...
 mrsPeerConnectionCreateAnswer(PeerConnectionHandle peerHandle) noexcept;
 
+/// Set the bitrate allocated to all RTP streams sent by this connection.
+/// Other limitations might affect these limits and are respected (for example
+/// "b=AS" in SDP).
+///
+/// Setting |start_bitrate_bps| will reset the current bitrate estimate to the
+/// provided value.
+///
+/// If any of the arguments has a negative value, it will be ignored.
```
djee-ms: Please add the unit of those (_i.e._ expand "bps" to "bits-per-second").

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1159,10 +1159,31 @@ public bool CreateAnswer()
...
         }
 
+        /// <summary>
+        /// Set the bitrate allocated to all RTP streams sent by this connection.
+        /// Other limitations might affect these limits and are respected (for example
+        /// "b=AS" in SDP).
+        /// </summary>
+        /// <param name="minBitrateBps">Minimum bitrate.</param>
+        /// <param name="startBitrateBps">Start/current target bitrate.</param>
+        /// <param name="maxBitrateBps">Maximum bitrate.</param>
```
djee-ms: Same: "Minimum bitrate, in bits per second.", etc.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1159,10 +1159,31 @@ public bool CreateAnswer()
...
+        {
+            ThrowIfConnectionNotOpen();
+            int signedMinBitrateBps = minBitrateBps.HasValue ? (int)minBitrateBps.Value : -1;
+            int signedStartBitrateBps = startBitrateBps.HasValue ? (int)startBitrateBps.Value : -1;
+            int signedMaxBitrateBps = maxBitrateBps.HasValue ? (int)maxBitrateBps.Value : -1;
+            if (PeerConnectionInterop.PeerConnection_SetBitrate(_nativePeerhandle,
+                signedMinBitrateBps, signedStartBitrateBps, signedMaxBitrateBps) != Utils.MRS_SUCCESS)
+            {
+                throw new InvalidOperationException();
+            }
```
djee-ms:
```cs
uint res = PeerConnectionInterop.PeerConnection_SetBitrate(...);
Utils.ThrowOnErrorCode(res);
```
This makes customizing the mapping {error code} <-> {exception} must easier, for example if we add new exception types in the future, and keeps it consistent.

PR #132, opened 2019-11-22
Expose workaround for H.264 Hololens encoder bug

fibann:
Mitigates https://github.com/webrtc-uwp/webrtc-uwp-sdk/issues/210.

TODO this is dependent on https://github.com/webrtc-uwp/webrtc-windows/pull/73

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
@@ -656,11 +656,12 @@ MRS_API mrsResult MRS_CALL mrsSdpForceCodecs(const char* message,
                                              char* buffer,
                                              uint64_t* buffer_size) noexcept;
 
-enum class HololensH264EncoderWorkaround : int32_t { CROP = 0, PAD = 1};
+/// Must be the same as PeerConnection::FrameHeightRoundMode.
+enum class FrameHeightRoundMode : int32_t { NONE = 0, CROP = 1, PAD = 2};
 
-/// See PeerConnection.SetHololensH264EncoderWorkaround for details.
+/// See PeerConnection::SetFrameHeightRoundMode.
 MRS_API void MRS_CALL
-mrsSetHololensH264EncoderWorkaround(HololensH264EncoderWorkaround value);
+mrsSetFrameHeightRoundMode(FrameHeightRoundMode value);
 
 //
 // Generic utilities
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
@@ -668,7 +668,7 @@ MRS_API mrsResult MRS_CALL mrsSdpForceCodecs(const char* message,
                                              char* buffer,
                                              uint64_t* buffer_size) noexcept;
 
-enum class HololensH264EncoderWorkaround : uint32_t { CROP = 0, PAD = 1};
+enum class HololensH264EncoderWorkaround : int32_t { CROP = 0, PAD = 1};
 
 /// See PeerConnection.SetHololensH264EncoderWorkaround for details.
 MRS_API void MRS_CALL
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
@@ -668,6 +668,12 @@ MRS_API mrsResult MRS_CALL mrsSdpForceCodecs(const char* message,
                                              char* buffer,
                                              uint64_t* buffer_size) noexcept;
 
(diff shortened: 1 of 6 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
```diff
@@ -668,6 +668,12 @@ MRS_API mrsResult MRS_CALL mrsSdpForceCodecs(const char* message,
                                              char* buffer,
                                              uint64_t* buffer_size) noexcept;
 
+enum class HololensH264EncoderWorkaround : uint32_t { CROP = 0, PAD = 1};
```
djee-ms: This is unsigned here but signed in C#...

djee-ms:
webrtc-uwp/webrtc-windows#73 has been merged to `master` and `release/1.0`

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
```diff
@@ -172,5 +172,12 @@ public static void ThrowOnErrorCode(uint res)
                     throw new ArgumentOutOfRangeException("Invalid ID passed to AddDataChannelAsync().");
             }
         }
+
+        /// <summary>
+        /// See <see cref="PeerConnection.SetFrameHeightRoundMode(PeerConnection.FrameHeightRoundMode)"/>.
+        /// </summary>
+        /// <param name="value"></param>
+        [DllImport(dllPath, CallingConvention = CallingConvention.StdCall, EntryPoint = "mrsSetFrameHeightRoundMode")]
+        public static unsafe extern void SetFrameHeightRoundMode(int value);
```
djee-ms: We can also use a C# enum here, to make it more robust.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1333,6 +1333,39 @@ public static Task<List<VideoCaptureFormat>> GetVideoCaptureFormatsAsync(string
...
+            /// <summary>
+            /// Leave frames unchanged.
+            /// </summary>
+            NONE = 0,
+            /// <summary>
+            /// Crop frame height to the nearest multiple of 16.
+            /// </summary>
+            CROP = 1,
+            /// <summary>
+            /// Pad frame height to the nearest multiple of 16.
```
djee-ms: Add comment on padding method? (mirrored bottom lines I think)

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1333,6 +1333,39 @@ public static Task<List<VideoCaptureFormat>> GetVideoCaptureFormatsAsync(string
...
+            /// </summary>
+            PAD = 2
+        }
+
+        /// <summary>
+        /// Use this function to select whether resolutions where height is not multiple of 16
+        /// should be cropped, padded or left unchanged.
+        /// Default is <see cref="FrameHeightRoundMode.NONE"/>, except on Hololens 1 where it is
+        /// <see cref="FrameHeightRoundMode.CROP"/> to avoid severe artifacts produced by
+        /// the H.264 hardware encoder.
```
djee-ms: That doesn't seem to be the case currently. `InitHololensH264Workaround()`, which overrides the default no-op behavior, is only called from within `SetFrameHeightRoundMode()`. So by default the no-op behavior is still active.

fibann: Good catch. There was a call in the PeerConnection factory method, but it seems it has been merged out 😕

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1188,4 +1271,9 @@ ErrorOr<RefPtr<PeerConnection>> PeerConnection::create(
   return RefPtr<PeerConnection>(peer);
 }
 
+void PeerConnection::SetFrameHeightRoundMode(FrameHeightRoundMode value) {
+  InitHololensH264Workaround();
```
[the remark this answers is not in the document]
fibann: See other comment, there is supposed to be another call in PeerConnection::create. Calling it here initializes the static bool in InitHololensH264Workaround so that a following call to PeerConnection::create won't reset the value.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
```diff
@@ -656,6 +656,13 @@ MRS_API mrsResult MRS_CALL mrsSdpForceCodecs(const char* message,
                                              char* buffer,
                                              uint64_t* buffer_size) noexcept;
 
+/// Must be the same as PeerConnection::FrameHeightRoundMode.
+enum class FrameHeightRoundMode : int32_t { NONE = 0, CROP = 1, PAD = 2};
```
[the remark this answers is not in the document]
fibann: There seems to be include order problems, will leave it alone.

PR #135, opened 2019-11-27
Fix crash on delete on x86

fibann:


djee-ms:
For context, instantiating the implementation type of a wrapper requires using its `::create()` factory method OR setting its `thisWeak_` member (which are equivalent), none of which were done here.

PR #137, opened 2019-12-04
Backport #132 HL1 frame height workaround

djee-ms:
This backport changes the original code due to some link error on Desktop.
Making a PR for it, then that change will be forward-ported to `master` to fix the same issue there.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/peer_connection.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/peer_connection.h
@@ -244,6 +244,30 @@ class PeerConnection : public webrtc::PeerConnectionObserver,
   /// connection.
   bool IsLocalVideoTrackEnabled() const noexcept;
 
+  enum class FrameHeightRoundMode {
+    /// Leave frames unchanged.
+    NONE = 0,
+
+    /// Crop frame height to the nearest multiple of 16.
+    /// ((height - nearestLowerMultipleOf16) / 2) rows are cropped from the top
+    /// and (height - nearestLowerMultipleOf16 - croppedRowsTop) rows are
+    /// cropped from the bottom.
+    CROP = 1,
+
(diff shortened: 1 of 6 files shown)


djee-ms:
!azp

libs/Microsoft.MixedReality.WebRTC.Native/include/peer_connection.h
```diff
@@ -244,6 +244,31 @@ class PeerConnection : public webrtc::PeerConnectionObserver,
...
+    /// are added symmetrically at the bottom.
+    PAD = 2
+  };
+
+  /// [HoloLens 1 only]
+  /// Use this function to select whether resolutions where height is not multiple of 16
+  /// should be cropped, padded or left unchanged.
+  /// Default is FrameHeightRoundMode::NONE, except on Hololens 1 where it is
+  /// FrameHeightRoundMode::CROP to avoid severe artifacts produced by
+  /// the H.264 hardware encoder.
```
fibann: Should be changed, now the setting doesn't exist at all on other platforms.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1301,6 +1301,45 @@ public static Task<List<VideoCaptureFormat>> GetVideoCaptureFormatsAsync(string
...
+            /// </summary>
+            PAD = 2
+        }
+
+        /// <summary>
+        /// Use this function to select whether resolutions where height is not multiple of 16
+        /// should be cropped, padded or left unchanged.
+        /// Default is <see cref="FrameHeightRoundMode.NONE"/>, except on Hololens 1 where it is
+        /// <see cref="FrameHeightRoundMode.CROP"/> to avoid severe artifacts produced by
+        /// the H.264 hardware encoder.
```
fibann: Should be same as C++ API.

PR #138, opened 2019-12-04
Add simple statistics gathering

fibann:
Made a simplified wrapper around the WebRTC stats API that contains only a few stats parameters considered of common use.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
@@ -1122,13 +1122,6 @@ T& FindOrInsert(std::vector<std::pair<std::string, T>>& vec,
   return vec.emplace_back(id, T{}).second;
 }
 
-class ReportWrapper : public Microsoft::MixedReality::WebRTC::RefCountedBase {
- public:
-  ReportWrapper(const rtc::scoped_refptr<const webrtc::RTCStatsReport>& report)
-      : report_(report) {}
-  rtc::scoped_refptr<const webrtc::RTCStatsReport> report_;
-};
-
 }  // namespace
 
 mrsResult MRS_CALL
@@ -1145,7 +1138,9 @@ mrsPeerConnectionGetSimpleStats(PeerConnectionHandle peerHandle,
       void OnStatsDelivered(
           const rtc::scoped_refptr<const webrtc::RTCStatsReport>& report)
           override {
-        (*callback_)(user_data_, new ReportWrapper(report));
+        report->AddRef();
+
+        (*callback_)(user_data_, report.get());
       }
     };
     rtc::scoped_refptr<Collector> collector =
@@ -1165,7 +1160,7 @@ mrsStatsReportGetObjects(mrsStatsReportHandle report_handle,
   if (!report_handle || !user_data) {
     return Result::kInvalidNativeHandle;
   }
-  auto& report = static_cast<ReportWrapper*>(report_handle)->report_;
+  auto report = static_cast<const webrtc::RTCStatsReport*>(report_handle);
 
   if (!strcmp(stats_type, "DataChannelStats")) {
     for (auto&& stats : *report) {
@@ -1219,4 +1214,13 @@ mrsStatsReportGetObjects(mrsStatsReportHandle report_handle,
   }
 
   return Result::kSuccess;
+}
+
+MRS_API mrsResult MRS_CALL
+mrsStatsReportRemoveRef(mrsStatsReportHandle stats_report) {
+  if (auto rep = static_cast<const webrtc::RTCStatsReport*>(stats_report)) {
+    rep->Release();
+    return Result::kSuccess;
+  }
+  return Result::kInvalidNativeHandle;
 }
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
@@ -1126,14 +1126,14 @@ T& FindOrInsert(std::vector<std::pair<std::string, T>>& vec,
 
 mrsResult MRS_CALL
 mrsPeerConnectionGetSimpleStats(PeerConnectionHandle peerHandle,
-                                PeerConnectionSimpleStatsCallback callback,
+                                PeerConnectionGetSimpleStatsCallback callback,
                                 void* user_data) {
   if (auto peer = static_cast<PeerConnection*>(peerHandle)) {
     struct Collector : webrtc::RTCStatsCollectorCallback {
-      Collector(PeerConnectionSimpleStatsCallback callback, void* user_data)
+      Collector(PeerConnectionGetSimpleStatsCallback callback, void* user_data)
           : callback_(callback), user_data_(user_data) {}
 
-      PeerConnectionSimpleStatsCallback callback_;
+      PeerConnectionGetSimpleStatsCallback callback_;
       void* user_data_;
       void OnStatsDelivered(
           const rtc::scoped_refptr<const webrtc::RTCStatsReport>& report)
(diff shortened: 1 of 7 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
```diff
@@ -5,6 +5,7 @@
 // line, to prevent clang-format from reordering it with other headers.
 #include "pch.h"
 
+#include "api/stats/rtcstats_objects.h"
```
djee-ms: Can you please separate core headers from mr-webrtc ones with a blank line?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1214,6 +1214,117 @@ public void SetRemoteDescription(string type, string sdp)
...
+            void Handle(in VideoReceiverStats stats);
+            void Handle(in TransportStats stats);
+        }
+
+        public void GetSimpleStats(StatsCollector collector)
+        {
+            var collectorHandle = GCHandle.Alloc(collector, GCHandleType.Normal);
+            PeerConnectionInterop.PeerConnection_GetSimpleStats(_nativePeerhandle,
+                PeerConnectionInterop.SimpleStatsCallback, GCHandle.ToIntPtr(collectorHandle));
+        }
```
djee-ms: Is that staying as it or changing to something more user friendly like `PeerConnection.GetStats(out TransportStats)` or `LocalVideoTrack.GetStats(out VideoSenderStats)`?

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
```diff
@@ -1061,66 +1069,78 @@ mrsPeerConnectionGetSimpleStats(PeerConnectionHandle peerHandle,
...
-                // TODO
-              }
-            }
-          }
-        }
-
-        for (auto&& stats : pending_audio_sender_stats) {
-          (*callback_)(user_data_, "AudioSenderStats", &stats.second);
-        }
+        (*callback_)(user_data_, new ReportWrapper(report));
```
djee-ms: Use a `Callback<>` object? This is what it's for, and will avoid the NULL dereferencing.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
```diff
@@ -1061,66 +1069,78 @@ mrsPeerConnectionGetSimpleStats(PeerConnectionHandle peerHandle,
...
   return Result::kInvalidNativeHandle;
 }
+
+mrsResult MRS_CALL
+mrsStatsReportGetObjects(mrsStatsReportHandle report_handle,
+                         const char* stats_type,
+                         mrsStatsReportGetObjectCallback callback,
+                         void* user_data) {
+  if (!report_handle || !user_data) {
+    return Result::kInvalidNativeHandle;
```
djee-ms: Return `kInvalidParameter` for `!user_data`.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
```diff
@@ -1061,66 +1069,78 @@ mrsPeerConnectionGetSimpleStats(PeerConnectionHandle peerHandle,
...
+  }
+  auto& report = static_cast<ReportWrapper*>(report_handle)->report_;
+
+  if (!strcmp(stats_type, "DataChannelStats")) {
+    for (auto&& stats : *report) {
+      if (!strcmp(stats.type(), "data-channel")) {
+        const auto& dc_stats = stats.cast_to<webrtc::RTCDataChannelStats>();
+        mrsDataChannelStats simple_stats{
+            dc_stats.timestamp_us(),     *dc_stats.datachannelid,
+            *dc_stats.messages_sent,     *dc_stats.bytes_sent,
```
djee-ms: The `operator*` will assert if "not defined", what does that mean? Is there a risk to assert here?

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
```diff
@@ -694,10 +694,18 @@ struct mrsAudioSenderStats {
...
+using PeerConnectionSimpleStatsCallback = void (MRS_CALL*)(void* user_data, mrsStatsReportHandle stats_report);
+using mrsStatsReportGetObjectCallback =
+    void(MRS_CALL*)(void* user_data, void* stats_object);
 
 MRS_API mrsResult MRS_CALL mrsPeerConnectionGetSimpleStats(
-    PeerConnectionHandle peerHandle, PeerConnectionSimpleStatsCallback callback,
+    PeerConnectionHandle peer_handle, PeerConnectionSimpleStatsCallback callback,
                                 void* user_data);
 
+MRS_API mrsResult MRS_CALL
```
djee-ms: TODO: documenting comments

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
```diff
@@ -269,31 +271,24 @@ public static void RemoteAudioFrameCallback(IntPtr userData, AudioFrame frame)
...
-                case "VideoSenderStats":
-                    collector.Handle(Marshal.PtrToStructure<PeerConnection.VideoSenderStats>(statsObject));
-                    break;
-                case "VideoReceiverStats":
-                    collector.Handle(Marshal.PtrToStructure<PeerConnection.VideoReceiverStats>(statsObject));
-                    break;
-                case "TransportStats":
-                    collector.Handle(*(PeerConnection.TransportStats*)statsObject);
-                    break;
+                audioSenderStatsList.Add(Marshal.PtrToStructure<PeerConnection.AudioSenderStats>(statsObject));
```
djee-ms: Did you figure out the difference between direct pointer cast and `PtrToStructure`?

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
```diff
@@ -433,8 +428,10 @@ internal struct VideoDeviceConfiguration
         public delegate void PeerConnectionAudioFrameCallback(IntPtr userData, AudioFrame frame);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall, CharSet = CharSet.Ansi)]
-        public delegate void PeerConnectionSimpleStatsCallback(IntPtr userData, string statsType, IntPtr statsObject);
+        public delegate void PeerConnectionSimpleStatsCallback(IntPtr userData, IntPtr statsReport);
```
djee-ms: Use `StatsReport(Handle)` instead of `IntPtr`?

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
fibann: Pointer cast only compiles if the struct is blittable (makes sense).

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
fibann: As said IRL this is for reverse PInvoke so needs to be IntPtr

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
fibann: Discussed IRL. I think we can assume that the members mandated by the spec will always be defined. I'll add checks to the others.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
fibann: Changed to an IEnumerable based interface.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
fibann: This is in the C interop so I guess `void*` is the only option.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
fibann: Actually we don't care about user_data here, we are supposed to pass it through.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.h
fibann: Done

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp
fibann: Done

PR #177, opened 2020-02-01
Typo in method name.

sipsorcery:


djee-ms:
Hi @sipsorcery and thank you for taking the time to make a PR. However I'm afraid there is no mistake here, the event name is correct on the `master` branch:
https://github.com/microsoft/MixedReality-WebRTC/blob/0ad9423ac856d1426cb043404ecedf93e85406c6/libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs#L663

Are you by any chance confusing it with the old name on the `release/1.0` branch? The event was renamed on `master` from `I420` to `I420A` for naming consistency (see 73b69d7da6c74cb02b97e5ec1a627e7ce07eba55).

sipsorcery:
Ah ok got it. I was using the nuget package.

PR #184, opened 2020-02-07
New test program to render audio and video streams from a browser peer

sipsorcery:
The goal is to provide the simplest possible test program for interop between a browser based WebRTC peer and the dotnet WebRTC library.

testrun


djee-ms:
That sounds like a good idea, thanks @sipsorcery!
Since this is more than just a few lines of fixes, I have to validate a few things internally I think to make sure we're OK to take the change. In particular I am wondering about that `Program.cs` header and whether that should be the standard Microsoft header instead, and your name added to the CONTRIBUTORS list. Also I didn't find the time to test it yet, I'd like to have a look and see if that fits well with the upcoming changes. But I don't foresee any major issue either way. Please bear with me :wink:

examples/TestReceiveVideo/Program.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows.Forms;
+using Microsoft.MixedReality.WebRTC;
+using WebSocketSharp;
+using WebSocketSharp.Net.WebSockets;
+using WebSocketSharp.Server;
+
+namespace TestNetCoreConsole
```
djee-ms: Should that namespace be `TestReceiveVideo` instead?

examples/TestReceiveVideo/Program.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        protected override void OnMessage(MessageEventArgs e)
+        {
+            MessageReceived(this.Context, e.Data);
+        }
+    }
+
+    class Program
+    {
+        private const string WEBSOCKET_CERTIFICATE_PATH = "certs/localhost.pfx";
```
djee-ms: This certificate file is not included. How does a user get/create it? Is it created by the code just below in `Main()`, or on the opposite does that code needs to load that certificate?

examples/TestReceiveVideo/TestReceiveVideo.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
djee-ms: Is it preferable to have a separate solution, or use the main one at the root of the repository? I am genuinely asking, I don't know what's best. Why did you go with a separate solution in the first place?

sipsorcery:
Addressed concerns and improved example. Example now able to deal with multiple simultaneous peers.

mr_webrtc_demo

examples/TestReceiveAV/README.md
```diff
@@ -0,0 +1,44 @@
...
+ - Build and run the solution file,
+ - Open the `webrtc.html` page in a WebRTC enabled browser.
+ - Click the `Start` button in the Web Browser.
+ - If successful a new Window should pop-up displaying the Video stream received from the browser and the audio should be played on the default system speaker.
+
+ #Notes
+
+ A web socket connection is used to exchange the SDP offer and answer as well as ICE candidates. The web socket connection will only work if listening port on the test application can be accessed by the browser. It should work correctly if both are on the same development machine or local network.
+
+ A PKCS12 archive file is included which contains an X509 certificate with a common name of local host as well as an arbitrarily generated private key. This archive is suitable for use when the web socket connection is between a browser and the test application running on the same machine and since the private key is freely accessible it should not be used outside of a development environment. The instructions use to generate the certificate using openssl are below:
```
djee-ms: I don't see anything included. In any case I don't think we should include a private certificate into the repository, and that might not even be possible for policy reasons. The instructions below to generate one should be enough.

sipsorcery: I've updated the readme to reflect the need to generate the pfx file (the exact same pfx file can be found here https://github.com/sipsorcery/sipsorcery-media/tree/master/examples/WebRtcDaemon/Certs if it's ever needed for testing).

examples/TestReceiveAV/README.md
```diff
@@ -0,0 +1,45 @@
+#Usage
+
+ - Generate an X509 certificate and key for the web socket server using the instructions below,
+ - Build and run the solution file,
```
fibann:
```suggestion
 - Build and run the TestReceiveAV project
```

examples/TestReceiveAV/README.md
```diff
@@ -0,0 +1,45 @@
...
+ - If successful a new Window should pop-up displaying the Video stream received from the browser and the audio should be played on the default system speaker.
+
+ #Certificate Instructions
+
+ A web socket connection is used to exchange the SDP offer and answer as well as ICE candidates. The web socket connection will only work if listening port on the test application can be accessed by the browser. It should work correctly if both are on the same development machine or local network.
+
+ A PKCS12 archive file which contains an X509 certificate with a common name of "localhost" and a private key is required. The archive file is needed for the web socket server and is suitable for use when the web socket connection is between a browser and the test application running on the same machine. The instructions use to generate the certificate using openssl are below:
+
+ ````
+openssl req -config req.conf -x509 -newkey rsa:4096 -keyout private/localhost.pem -out localhost.pem -nodes -days 3650
```
fibann: I tried running this locally passing the req.conf example below, I got
```
Error Loading extension section x509_ext
19992:error:2207507C:X509 V3 routines:v2i_GENERAL_NAME_ex:missing value:crypto/x509v3/v3_alt.c:512:
19992:error:22098080:X509 V3 routines:X509V3_EXT_nconf:error in extension:crypto/x509v3/v3_conf.c:47:name=subjectAltName, value=localhost, IP:127.0.0.1, IP:::1
```

Maybe there is something missing/wrong in the sample?

examples/TestReceiveAV/TestReceiveAV.csproj
```diff
@@ -0,0 +1,26 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.MixedReality.WebRTC" Version="1.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="SIPSorcery.WebSocketSharp" Version="0.0.1" />
```
fibann: We'd rather not take dependencies on third party components unless absolutely necessary. Any chance you can use the [.NET WebSocket classes](https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.websocket) instead?

sipsorcery: I wish I could use the in built .NET classes but to the best of my knowledge there's no WebSocket server support outside of ASP.NET and running Kestrel or IIS Express is overkill for a sample like this. I'd be very happy to be corrected but the popularity of projects like https://github.com/sta/websocket-sharp (of which the `SIPSorcery.WebSocketSharp` is a minor fork) and others points to the same fact.

fibann: HttpListener seems to be able to act as a WebSocket server: https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistenercontext.acceptwebsocketasync. It seems to me that it should be reasonably easy to write the same logic around it, unless I am missing something. Seems to be working only on Windows 8 onward but this shouldn't be an issue for us - also it might be less performant than a bespoke server implementation but since this is a sample app we don't care too much.

examples/TestReceiveAV/README.md
fibann: It worked after rewriting the subjectAltName line as in [this sample](https://gist.github.com/andrewconnell/6c2c14e80ef45b232d11e2f4706489b5) - I am not familiar with the syntax so I am not sure why, but it might give you an idea.

Btw the readme should mention that you need to assign an empty password to the PKCS12 file, or the server program won't load it as it is right now.

examples/TestReceiveAV/TestReceiveAV.csproj
sipsorcery: That's relying on `http.sys` which is Windows specific. The current web socket server logic used in the example is not platform specific. Is it worth changing just to avoid a dependency in an example?

fibann: I have validated the dependency internally, sounds like it's fine! You're right, it's nicer if the sample works on all platforms.

examples/TestReceiveAV/README.md
sipsorcery: I've updated the openssl config and instructions for generating the localhost self signed certificate.

djee-ms:
We should add a CONTRIBUTORS file to add @sipsorcery to it.
@sipsorcery do you contribute as an individual or as a company with this change?

sipsorcery:
@djee-ms very kind thanks. As an individual would be good, my name and and email address are shown on my GitHub profile.

PR #186, opened 2020-02-12
[Experimental] Add interface and samples for a signaling service

fibann:
Starting point for discussion.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
@@ -1,47 +1,221 @@
 using System;
 using System.Collections.Generic;
+using System.Net;
 using System.Text;
 
 namespace Microsoft.MixedReality.WebRTC
 {
+    public enum SignalingType
+    {
+        SdpOffer,
+        SdpAnswer,
+        IceCandidate
+    }
+
+    public class SignalingMessage
+    {
+        public string SessionId;
+        public SignalingType Type;
+        public string Payload;
+    }
+
     interface ISignalingService : IDisposable
     {
         /// <summary>
-        /// Get the currently available remote endpoints. These can be passed to
-        /// <see cref="ConnectToRemoteEndpoint(string, PeerConnection)"/>.
+        /// Start the service and make the local process visible as `localEndPointId`.
+        /// </summary>
+        void Start(string localEndPointId);
+
+        /// <summary>
+        /// Stop the service.
+        /// </summary>
+        void Stop();
+
+        /// <summary>
+        /// Get the endpoints that are currently active.
         /// </summary>
         IEnumerable<string> RemoteEndPoints { get; }
 
         /// <summary>
-        /// Make a local <see cref="PeerConnection"/> visible to remote endpoints under the given ID.
+        /// Raised every time there is a change in the list of remote endpoints.
+        /// Every delegate added to this is called once even if there are no updates.
         /// </summary>
-        /// <remarks>
-        /// SDP/ICE messages sent to the passed ID will be forwarded to `pc`.
-        /// ??? Replies?
-        /// </remarks>
-        void PublishLocalEndPoint(string id, PeerConnection pc);
+        event Action RemoteEndPointsChanged;
 
         /// <summary>
-        /// Remove a previously published local endpoint.
+        /// Send a message to the remote endpoint.
         /// </summary>
-        void RemoveLocalEndPoint(string id);
+        void SendToRemoteEndpoint(string targetEndPointId, SignalingMessage message);
 
         /// <summary>
-        /// Connect the passed <see cref="PeerConnection"/> to the endpoint with the passed ID.
+        /// Raised when a message is received.
         /// </summary>
-        /// <remarks>
-        /// The method calls <see cref="PeerConnection.CreateOffer"/> on the passed connection.
-        /// The service forwards any following message between the connection and the remote endpoint.
-        /// </remarks>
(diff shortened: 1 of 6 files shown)


libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public class SignalingMessage
+    {
+        public string SessionId;
+        public SignalingType Type;
+        public string Payload;
+    }
+
+    interface ISignalingService : IDisposable
```
djee-ms: The `IDisposable` should be on the implementation.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
...
+        /// </summary>
+        void SendToRemoteEndpoint(string targetEndPointId, SignalingMessage message);
+
+        /// <summary>
+        /// Raised when a message is received.
+        /// </summary>
+        event Action<string, SignalingMessage> MessageReceived;
+    }
+
+    public class PeerSignalingService : ISignalingService
```
djee-ms: TODO: Move to separate file.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        /// <summary>
+        /// Stop the service.
+        /// </summary>
+        void Stop();
+
+        /// <summary>
+        /// Get the endpoints that are currently active.
+        /// </summary>
+        IEnumerable<string> RemoteEndPoints { get; }
```
djee-ms: This is a bit confusing. WebRTC signaling is about a peer-to-peer connection, there are never multiple peers. Is this a discovery/matchmaking concept rather? Is this interface really about signaling (the process of connecting 2 known peers), or about discovery/matchmaking (the process of selecting the 2 peers before signaling)?

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
...
+        public void Stop()
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+    static class SignalingExamples
+    {
+        // Use the service to connect a local PeerConnection to another chosen by the user among the available ones.
+        public static void ChooseAndConnect(PeerConnection pc)
```
djee-ms: This looks like a useful utility, if taking an `ISignalingService` as parameter.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
...
+                                pc.SetRemoteDescription("answer", message.Payload);
+                            }
+                            else if (message.Type == SignalingType.IceCandidate)
+                            {
+                                var parts = message.Payload.Split('|');
+                                pc.AddIceCandidate(parts[2], int.Parse(parts[1]), parts[0]);
+                            }
+                        }
+                    };
+                pc.CreateOffer();
```
djee-ms: I think either `CreateOffer()` should be removed, or you should go to the full length and see the connection through by also waiting for the answer and the connected event. This alone starts connecting but terminates the method while the connection is pending.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
```diff
@@ -0,0 +1,221 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Text;
+
+namespace Microsoft.MixedReality.WebRTC
+{
+    public enum SignalingType
```
djee-ms: Note: we should add that enum to the API instead of carrying around strings which are error-prone.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
fibann: Good point - this is about both. Maybe we want to split it? It doesn't seem trivial as finding endpoints and communicating with them are related even if not the same thing. Ideas?

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
fibann: Yes, will try to turn it into a utility.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
fibann: Yes, I omitted it but that's necessary. I assume this should also listen to `RenegotiationNeeded` to create new offers.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
fibann: Logged as #188, can be done independently - it's an API break so it should be timed carefully.

libs/Microsoft.MixedReality.WebRTC/ISignalingService.cs
fibann: I have taken the responsibility of creating offers out of the utility, now the utility only forwards outgoing/incoming messages, and the responsibility of initiating/reinitiating the negotiation is on the caller. It seems a bit cleaner this way, but I am willing to discuss this.

fibann:
Closing as no work on this is scheduled for the foreseeable future.

PR #193, opened 2020-02-19
Android support

eanders-MS:
**Add support for Android platform**

New folders below `/tools/build` contain per-platform scripts to build MR-WebRTC, and to build the core Chromium libwebrtc base library:
* `/tools/build/libwebrtc`: Scripts to build the Chromium base library for a given platform.
* `/tools/build/android`: Contains a Gradle+CMake configuration to build MR-WebRTC for Android/arm64-v8a platform.
* In the future possibly add more, even all, platform builds here? iOS build is similar to Android. 
* Interested to hear what you think of this organization. Happy to change it.

**Status**
* libwebrtc build works for Android, with placeholder codepaths for future platforms. Build artifact for Android is `libwebrtc.aar`. It is automatically copied to the Unity sample project upon build success.
* Android build works, though README instructions are incomplete. Build artifact is `Microsoft.MixedReality.WebRTC.Native.aar`, and is also copied to the Unity sample project automatically.
* In testing, the DLL successfully loads and initializes on Android (Quest).
* In a two-client test between Desktop/Editor and Quest, Desktop crashes after receiving the SDP answer.

**TODO**
- [ ] Finish adding Android build instructions to READMEs.
- [x] Fix: VideoChatDemo crashes on Oculus Quest unless Local Video and Scene Capture are disabled.
- [ ] Revert: **Unsure how to fix this one**: The change list shows I changed some things in webrtc-uwp-sdk. I should revert those changes.


eanders-MS:
Updated the PR description (not reflected in email below).

libs/Microsoft.MixedReality.WebRTC.Native/src/pch.h
```diff
@@ -25,13 +25,17 @@
...
+#define MRS_NODISCARD [[nodiscard]]
+
 #include <windows.h>
 
 #elif defined(MR_SHARING_ANDROID)
 
 #define WEBRTC_POSIX 1
 #define WEBRTC_ANDROID 1
 
+#define MRS_NODISCARD
```
djee-ms: This is because of C++14 I assume? I think we will need to remove it entirely, even on Windows, because libwebrtc doesn't support C++17. So far we were compiling it with C++17 anyway but I'd rather be compliant with the Google repo. This can stay as is for now though.

libs/Microsoft.MixedReality.WebRTC.Native/src/sdp_utils.h
```diff
@@ -55,4 +55,9 @@ std::string EncodeIceServers(const std::string& url,
                              const std::string& username,
                              const std::string& password);
 
+// Copied from src/pc/jsepsessiondescription.cc so that we don't rely on a binary
+// representation of absl::optional as per the Abseil Compatibility Guidelines:
+// https://abseil.io/about/compatibility
+std::optional<webrtc::SdpType> SdpTypeFromString(const std::string& type_str);
```
djee-ms: We should revert to C++14 instead, which solves the issue with `std::optional`. That's what I did on the [experimental/undock](https://github.com/microsoft/MixedReality-WebRTC/tree/experimental/undock) branch, and it works fine, though this requires several changes to downgrade the Native projects to remove C++17 features. Not sure what's the best way to integrate those changes though.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Plugins/arm64-v8a.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
djee-ms: I think the folder name should be `ARM64` according to Unity conventions, no? I can't find again where I saw those though.

In any case those should move inside `Microsoft.MixedReality.WebRTC.Unity/` instead of being inside the `Plugins/` folder, which is the per-project folder for user plugins. I saw other products with native plugins (_e.g._ Photon) are putting all their files including DLLs inside a single folder. But I can do that part later.

libs/Microsoft.MixedReality.WebRTC.Unity/Packages/manifest.json
```diff
@@ -1,8 +1,20 @@
...
+    "com.unity.2d.tilemap": "1.0.0",
+    "com.unity.ext.nunit": "1.0.0",
+    "com.unity.ide.rider": "1.1.0",
+    "com.unity.ide.vscode": "1.1.2",
+    "com.unity.multiplayer-hlapi": "1.0.2",
+    "com.unity.package-manager-ui": "2.2.0",
+    "com.unity.test-framework": "1.0.13",
+    "com.unity.textmeshpro": "2.0.1",
+    "com.unity.timeline": "1.1.0",
+    "com.unity.ugui": "1.0.0",
```
djee-ms: Probably don't need most of the new ones here (aside from `nunit`)

libs/Microsoft.MixedReality.WebRTC.Unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -158,7 +166,7 @@ PlayerSettings:
   applicationIdentifier: {}
   buildNumber: {}
   AndroidBundleVersionCode: 1
-  AndroidMinSdkVersion: 16
+  AndroidMinSdkVersion: 19
```
djee-ms: What's the rationale here? I remember something about version 16 though I forgot why.

projects/README.md
```diff
@@ -0,0 +1,9 @@
+# Instructions for per-platform builds
```
djee-ms: According to [pitchfork layout](https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs) (mentioned in README.md) this whole folder should be `<root>/tools/build` rather than `<root>/projects`.

projects/android/.gitignore
```diff
@@ -0,0 +1,9 @@
+*.iml
+.gradle
+/local.properties
+/.idea
+.DS_Store
```
djee-ms: `.DS_Store` should probably be in the top-level `.gitignore`.

projects/android/.gitignore
```diff
@@ -0,0 +1,9 @@
...
+build/
+/captures
+.externalNativeBuild
+.cxx
```
djee-ms: Are we ignoring all C++ files?

projects/android/build.gradle
```diff
@@ -0,0 +1,35 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
```
djee-ms: This and other new files need the Microsoft license header.

projects/android/webrtc-native/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
djee-ms: Can we exclude that from the `.gitignore` in `projects/` instead of having a separate file just for it?

projects/android/webrtc-native/src/main/cpp/jni_onload.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// See LICENSE in the project root for license information.
```
djee-ms: New shorter license header for open-source projects is:
> // Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

projects/libwebrtc/.gitignore
```diff
@@ -0,0 +1,4 @@
+.config.sh
```
djee-ms: Same, can we merge all `.gitignore`?

projects/libwebrtc/ENV-UNIX.md
```diff
@@ -0,0 +1,45 @@
...
+
+    WSL2 installation instructions: https://docs.microsoft.com/en-us/windows/wsl/wsl2-install
+
+    If building on a Windows-accessible drive (USB or main disk), ensure the drive is mounted with "-o metadata", e.g.:
+
+    `$> sudo mount -t drvfs D: /mnt/d -o metadata`
+  
+    This enables support for Linux-style file permissions (chown, chmod, etc).
+
+    The filesystem must be case sensitive. There are three ways to enable this. I recommend trying all of them.
```
djee-ms: Style: avoid "I"? 😉

projects/libwebrtc/build.sh
```diff
@@ -0,0 +1,175 @@
...
+        local arch=""
+        case "$TARGET_CPU" in
+        "arm") arch="armeabi-v7a" ;;
+        "arm64") arch="arm64-v8a" ;;
+        "x86") arch="x86" ;;
+        "x64") arch="x86_64" ;;
+        esac
+        local src="$SRC_DIR/src/$outdir/libwebrtc.aar"
+        local dst="../../libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Plugins/$arch/"
+        echo "Copying libwebrtc.aar to Unity sample scene."
```
djee-ms: "...sample **project**" (not scene)

projects/libwebrtc/lib.sh
```diff
@@ -0,0 +1,270 @@
...
+    echo -e "\e[39mRevision: \e[96m$REVISION\e[39m"
+}
+
+#-----------------------------------------------------------------------------
+function verify-webrtc-deps() {
+    echo -e "\e[39mVerifying WebRTC dependencies\e[39m"
+    case $HOST_OS in
+    "linux")
+        # Automatically accepts ttf-mscorefonts EULA
+        echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
```
djee-ms: Not sure we can ship some script which auto-accept some EULA for the user...

projects/android/gradlew
```diff
@@ -0,0 +1,172 @@
...
+##  Gradle start up script for UN*X
+##
+##############################################################################
+
+# Attempt to set APP_HOME
+# Resolve links: $0 may be a link
+PRG="$0"
+# Need this for relative symlinks.
+while [ -h "$PRG" ] ; do
+    ls=`ls -ld "$PRG"`
```
stephenatwork: nit: modern style is to use $()
```
ls=$(ls -ld "$PRG")
```

stephenatwork: Hmm, were these scripts written from scratch or based on some existing ones?

djee-ms: Doesn't look like just a style issue, apparently backtick is more error-prone:
http://mywiki.wooledge.org/BashFAQ/082

djee-ms: Also `gradlew` is generated by Graddle.

eanders-MS: That's correct, this script is auto-generated by Gradle.

libs/Microsoft.MixedReality.WebRTC.Unity/Packages/manifest.json
eanders-MS: These must have been auto-added by Unity. Likely due to my use of Unity 2019.2.12f1. I will revert this, and install 2018 to avoid such side effects.

libs/Microsoft.MixedReality.WebRTC.Unity/ProjectSettings/ProjectSettings.asset
eanders-MS: I was testing on Oculus Quest, and Oculus requires a minimum API level of 19. This doesn't need to be committed to the repo necessarily.

projects/README.md
eanders-MS: I will update the layout to follow pitchfork guidelines.

projects/android/.gitignore
eanders-MS: This file was generated by Gradle, but I agree .DS_Store should be in the root .gitignore. I will move it.

projects/android/.gitignore
eanders-MS: Interesting. I'm unsure why that's there. This file was generated by Gradle.

projects/android/webrtc-native/.gitignore
eanders-MS: I can remove this .gitignore file. This build/ folder is excluded at the outer level.

projects/libwebrtc/.gitignore
eanders-MS: It isn't safe to assume that files matching these filenames should be excluded everywhere across the project. Do you prefer root-level entries like:
```
tools/build/libwebrtc/.config.sh
```
?

projects/libwebrtc/ENV-UNIX.md
eanders-MS: Good catch.

projects/libwebrtc/lib.sh
eanders-MS: Good catch. Will remove.

projects/README.md
eanders-MS: The existing `<root>/tools/build` folder contains a script: `build.ps1`. With the addition of my subfolders, I think this file should be moved to its own subfolder. What do you think? Additionally, the project root contains `Microsoft.MixedReality.WebRTC.sln`, seems like it should exist under `tools` as well?

libs/Microsoft.MixedReality.WebRTC.Native/src/pch.h
eanders-MS: Interesting, I was able to put [[nodiscard]] back without compilation errors. As I was working through the build config issues it was generating compiler errors at one point. Android is being compiled with C++17.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Plugins/arm64-v8a.meta
eanders-MS: I think the convention might be `Plugins/Android` for Android native plugins, but not finding the docs for that now. I checked in the associated .meta files for each .aar, so they should be imported correctly in the build.

libs/Microsoft.MixedReality.WebRTC.Native/src/sdp_utils.h
eanders-MS: Your undock build instructions could integrate nicely into the /tools/build folder as another platform/target leveraging the new libwebrtc scripts.

projects/android/gradlew
stephenatwork: OK, so no need to review this file then! BTW I highly recommend https://www.shellcheck.net/ for checking shell scripts. It's available as a standalone program also.

libs/Microsoft.MixedReality.WebRTC.Unity/Packages/manifest.json
djee-ms: They're auto-added, can't remember under which circumstances, but can be safely removed manually from the Package Manager UI.

libs/Microsoft.MixedReality.WebRTC.Unity/ProjectSettings/ProjectSettings.asset
djee-ms: Not sure what's best. I don't know what device wouldn't support API level 19 (KitKat 4.4 from 2013). Note also that [from Unity 2019.3 this is the minimum level version supported](https://unity3d.com/unity/whats-new/2019.3.0), so anyway API level 16 won't work anymore.

projects/libwebrtc/.gitignore
djee-ms: Yes, I think it's easier to understand when there's a single root `.gitignore`. And marginally faster for git to read a single file.

projects/README.md
djee-ms: Yes agreed. Although I am still unclear on what approach to take for multi-platform build scripts. I would rather rely on a maintained solution like `CMake` or `gn` rather than roll our own scripts for all platforms.

The Visual Studio solution was at the root for discoverability, because (and these points should ideally be fixed):
- we do not ship binaries/packages for everything (Unity integration ships as source only for now, we should work on that to make it more user friendly)
- there was a single platform
- I don't think it is very intuitive to go look into the `tools/build` folder to find a Visual Studio solution for the project

Not sure how we can maintain good discoverability while moving everything into `tools/build`. Maybe have a single `build.sh`/`build.ps1` script at root if we go that way. Or maybe have a single `CMakeLists.txt` at root if we use CMake, and then have everything else inside `tools/build`. I'm open to suggestions.

libs/Microsoft.MixedReality.WebRTC.Native/src/pch.h
djee-ms: Ok good to know. I do like the change though, because I want to roll back native to C++14 to be in line with the official repo, since this will be internal code from now (the C++ library will now build on top of it, so can do whatever it wants, and will be C++17). So let's leave that `MRS_NODISCARD` macro.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Plugins/arm64-v8a.meta
djee-ms: I think `Plugins/Android` is for user's plugins. Probably Unity used to support only that, but now it seems you can put your plugins anywhere in `Assets`, so for third-party libraries it's best I think we keep all our files inside a single folder, that makes it much easier and cleaner for users importing multiple third-party libraries into their project and potentially also having their own plugins.

libs/Microsoft.MixedReality.WebRTC.Native/src/pch.h
djee-ms: Argh sorry I saw you reverted already. My initial comment was not clear. You can leave it as is.

eanders-MS: Oh sorry about that.

projects/libwebrtc/.gitignore
eanders-MS: .gitignores merged into the root

projects/libwebrtc/build.sh
eanders-MS: Fixed

projects/libwebrtc/lib.sh
eanders-MS: Removed

projects/android/.gitignore
eanders-MS: Removed that line (and merged this .gitignore into the root file)

projects/README.md
eanders-MS: `gn` might be an appropriate build system for this repo, given the growing number of platforms and options you plan to support. I am new to `gn` and `Ninja`, but the system seems very flexible with regard to build configuration and conditional build of components. I don't know how easily it can be coerced into building Visual Studio projects.

djee-ms: It does support generating some VS solution via the `--ide` flag. But no idea how "good" that is (can we regenerate the exact same solution we hand-crafted so far?). My other concern is that I have yet to understand how to integrate an existing `gn` project (libwebrtc) into another one (mr-webrtc), as ideally, we'd take advantage of that to have a single unified solution for building.

libs/Microsoft.MixedReality.WebRTC.Native/src/pch.h
djee-ms: Thanks

astaikos316:
I have been trying to run through the build process and came across that there is no copy.sh script in the android build folder and the README describes after going through the build process.

tools/build/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Fri Nov 08 10:58:59 PST 2019
+distributionBase=GRADLE_USER_HOME
```
stephenatwork: nit: line endings?

libs/Microsoft.MixedReality.WebRTC.Native/src/sdp_utils.h
stephenatwork: Is it worthwhile to add an 'invalid sdptype' so we can avoid the optional altogether?

djee-ms: No, this needs to return a `webrtc::SdpType` to pass the value to `webrtc::CreateSessionDescription()` and that's a Google type we cannot change.

Also:
- on `master` we compile everything with C++17 so there is no discrepancy and `absl::optional` is an alias for `std::optional`
- on `experimental/undock` we reverted to C++14 for the interop DLL to be compatible with the `libwebrtc` code, and we use `absl::optional` only

...so we can then directly use the implementation in `jsepsessiondescription.cc` instead of making a copy of it. But I think @eanders-MS is possibly compiling some fork of `master` in C++14 (`libwebrtc`'s default) hence hitting this issue. This is a temporary problem.

tools/build/android/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. Run `$./gradlew[.bat] assembleRelease`
+
+### Artifacts produced by the build
+
+* `./webrtc-native/build/outputs/aar/Microsoft.MixedReality.WebRTC.Native.aar`
+
+### Next Steps
+
+Run `./copy.sh` to copy the build artifacts to the Unity sample project.
```
djee-ms: As noted by @astaikos316 the `copy.sh` script is missing in this PR.

astaikos316:
When I try to run a build on an oculus quest to a pc, once I hit send offer from the pc to the quest, the quest nodedss stops polling but the app does not crash. No audio or video is passed at all. I haven’t set up any debug methods for the quest yet so I cannot provide any additional info yet. Has anyone else seen this behavior?

tools/build/android/gradle/wrapper/gradle-wrapper.properties
eanders-MS: This script is ultimately owned by Gradle. I didn't author it. It's part of Gradle's tooling.

tools/build/android/README.md
eanders-MS: Good catch. README was out of date. That script no longer exists (copy happens automatically now).

eanders-MS:
@astaikos316, A PC/Quest test of the Unity sample project is working ok for me. Are you testing with the Unity sample project?

Here's what I did, in case it helps:
* On a command prompt, run node-dss: `set DEBUG=dss*`, `npm start`.
* On another command prompt, run ngrok (https://ngrok.com) like this: `ngrok http 3000`. Take note of the ngrok forwarding URL. It will look something like `http://xxxxxxxx.ngrok.io`. ngrok lets the Quest hit node-dss, even though it will be listening on the PC's localhost.
* Build MR-WebRTC for Android and Windows (if needed).
* Open the Unity sample project.
    * Open the VideoChatDemo scene.
    * In VideoChatDemo/PeerConnection/NodeDssSignaler settings:
        * Set Http Server Address to the ngrok forwarding address.
        * Set Local Peer Id to 1.
        * Set Remote Peer Id to 2.
* Build a standalone Windows player, then launch it.
* In Unity, go back to VideoChatDemo/PeerConnection/NodeDssSignaler settings and switch the peer ids:
    * Set Local Peer Id to 2.
    * Set Remote Peer Id to 1.
* Switch to Android platform.
* Build and Run. This will build and deploy to your USB-attached Quest.
* Once the app is deployed and running on Quest, click "Create Offer" in the Windows standalone.
* You should start to see signals getting sent and received, and eventually see the PC feed on the Quest remote panel.

astaikos316:
@eanders-MS from what I can see from node-dss and logcat, the quest never sends an answer message and the dss polling stops after an offer from the windows standalone is sent.  No errors stand out in logcat when attached to the quest.

djee-ms:
My 2c (didn't try) : network/signaling issues aside, not sending an answer is generally a sign of the offer failing to apply, either because of unsupported options (_e.g._ unsupported video codec), or because the offer arrives at the wrong time or the answer tries to be created too soon, both of which when the peer connection is in an invalid state (must be in "idle" to transition to "have-remote-offer" when applying the offer, and then be able to transition to "have-local-answer" when creating the answer). Typically as an example the latter fails in C# when forgetting to `await` the `SetRemoteDescriptionAsync()` call, and calling `CreateAnswer()` too soon.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/LocalAudioSource.cs
```diff
@@ -85,7 +86,13 @@ private void OnPeerInitialized()
         }
 
         private async void DoAutoStartActions(WebRTC.PeerConnection nativePeer)
-        {
+        {
+            // If a headset is active then do not capture local streams.
+            if (XRDevice.isPresent)
```
eanders-MS: I don't think this is the right check. Will research.

tools/build/libwebrtc/config.sh
```diff
@@ -0,0 +1,84 @@
...
+
+BUILD_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source "$BUILD_DIR/lib.sh"
+
+#=============================================================================
+# Functions
+
+#-----------------------------------------------------------------------------
+function usage() {
```
djee-ms: I've just spent 4 hours setting up an Ubuntu VM with Hyper-V (╯°□°）╯︵ ┻━┻ and resizing the virtual disk to accommodate the gigantic repository (currently at 37GB total partition size) (╯°□°）╯︵ ┻━┻, to be told by the Google setup script that:
```
ERROR: The only supported distros are
 	Ubuntu 14.04 LTS (trusty)
 	Ubuntu 16.04 LTS (xenial)
 	Ubuntu 17.10 (artful)
 	Ubuntu 18.04 LTS (bionic)
 	Debian 8 (jessie) or later
``` 
Yes, I used Ubuntu 19.10 as suggested by default by Hyper-V. 😡

So I really think we should test ASAP in those scripts, preferably during `config.sh`:
1. The user runs a valid distribution (see list above)
2. The user has at least 40GB of disk space available

The actual check is in [`webrtc/build/install-build-deps.sh`](https://chromium.googlesource.com/chromium/src/+/master/build/install-build-deps.sh#110) which is outdated on `branch-heads/71` compared to recent Ubuntu releases (as linked on `master` it actually supports `eoan` 19.10). Ideally we'd grab that file and extract the supported distros from it, but it might be a bit complex to reach it via the chain of dependency (git branch => sub-repo chromium/build => script => grep on `supported_codenames`).

tools/build/libwebrtc/lib.sh
```diff
@@ -0,0 +1,275 @@
...
+    fi
+    echo -e "\e[39mRevision: \e[96m$REVISION\e[39m"
+}
+
+#-----------------------------------------------------------------------------
+function verify-webrtc-deps() {
+    echo -e "\e[39mVerifying WebRTC dependencies. You may be prompted to accept software licenses.\e[39m"
+    case $HOST_OS in
+    "linux")
+        echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections
```
djee-ms: This needs to be changed, we can't accept an EULA for the user.

tools/build/android/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Prerequisites
+
+1. Build the Chromium WebRTC base library. Instructions [here](../libwebrtc/README.md).
+
+2. Install prerequisites:
+    * Android Studio (https://developer.android.com/studio/index.html). Once installed, use the System Settings/Android SDK dialog to install the following packages:
+        * SDKs 9.0 (Pie), 8.1 (Oreo)
+        * CMake
+        * NDK 16.1.4479499
```
djee-ms: There's no option to select a version of the NDK as I know. I can only see "NDK (side by side)", which as of today resolves to 21.0.6113669. I _think_ that's OK but I can't remember the details about NDK versions.

djee-ms: Actually that's not even enough, I had to uncheck "Hide obsolete packages", then install "NDK (Obsolete)" version 21.0.6113669, then click "Sync Project with Gradle Files" before I was finally allowed to Build > Make Project.

tools/build/android/README.md
```diff
@@ -0,0 +1,36 @@
...
+    * JRE 1.8 (https://www.java.com/download/)
+
+3. Add environment variable JAVA_HOME pointing to your JRE install directory. e.g.: `JAVA_HOME="C:\Program Files\Android\Android Studio\jre"`
```
djee-ms: On Linux that's `apt install openjdk-8-jre` and I think already installed by the Google dependency script. Shall we mention it for completeness?

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
```diff
@@ -39,11 +39,7 @@ sealed class MonoPInvokeCallbackAttribute : Attribute
 
     internal static class Utils
     {
-#if MR_SHARING_WIN
-        internal const string dllPath = "Microsoft.MixedReality.WebRTC.Native.dll";
-#elif MR_SHARING_ANDROID
-        internal const string dllPath = "Microsoft.MixedReality.WebRTC.Native.so";
-#endif
+        internal const string dllPath = "Microsoft.MixedReality.WebRTC.Native";
```
djee-ms: Unfortunately this needs to be reverted, this makes DLL loading fail on Windows because the filename has dots in its name. See https://github.com/dotnet/runtime/issues/7223. I think we talked about it offline, or you mentioned it to me, but I just hit it and couldn't run `TestAppUWP.exe` anymore to test the connection.

tools/build/android/README.md
djee-ms: I noted later that this is also needed on Unity, although Unity will strongly suggest using the version shipped with the Unity Editor. Maybe we should recommend that too, or at least point it out.

tools/build/libwebrtc/lib.sh
djee-ms: Actually this was supposed to be fixed, looking through this PR comments above. Did that cropped up again later?

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
eanders-MS: This change made it possible to include a single copy of the DLL in Unity and it would load for multiple platforms (Windows Desktop and Android), but if that doesn't work in UWP then it has to be reverted. The consequence is that Unity projects will need to have multiple copies of this DLL present - one for the .dll extension (Windows/UWP) and another for the .so extension (Android) - each configured to load for that platform only.

tools/build/libwebrtc/lib.sh
eanders-MS: Oh I thought I got rid of that already.

tools/build/libwebrtc/config.sh
eanders-MS: That is terrible! Yes, we need to verify platform earlier.

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
djee-ms: Ah, right. But since they deploy to difference platforms, that has no impact on the final app size on each device, correct? So I guess it's just a matter of having 2 files instead of 1 in your Unity project. Am I missing something here?

eanders-MS: That sounds right to me.

tools/build/android/README.md
eanders-MS: This section is light on detail, due entirely to the fact that I wasn't sure what the best practice is here. I wanted to provide steps that could be done entirely on the command line and not rely on Android Studio. I haven't gone back and worked out what those steps are yet.

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
djee-ms: Fixed by renaming `Microsoft.MixedReality.WebRTC.Native.dll` into `mrwebrtc.dll` / `libmrwebrtc.so`, which both work with P/Invoke as `[DllImport("mrwebrtc")]`.

tools/build/android/README.md
djee-ms: We can polish later. I want to commit to `master` an experimental support version, then we refine the instructions.

tools/build/libwebrtc/lib.sh
djee-ms: Deleted.

tools/build/libwebrtc/config.sh
djee-ms: I pushed some basic checks as discussed (see d4255487ea4217ce1545d0d5950cf508e8580b7a). Can you please have a look @eanders-MS? My bash is quite rusty... :wink:

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/LocalAudioSource.cs
djee-ms: What's this check for? Why prevent local audio capture from XR devices?

eanders-MS: There's no problem capturing local audio from the device. This check is unnecessary.

djee-ms: Ok will delete, thanks!

tools/build/android/README.md
astaikos316: Trying to build in Android Studio and getting this build error.  any suggestions?

Build command failed.
Error while executing process /home/astaikos/Android/Sdk/cmake/3.10.2.4988404/bin/cmake with arguments {--build /home/astaikos/eanders/MixedReality-WebRTC/tools/build/android/webrtc-native/.externalNativeBuild/cmake/debug/arm64-v8a --target mrwebrtc}
[1/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/local_video_track_interop.cpp.o
[2/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/data_channel.cpp.o
[3/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/str.cpp.o
[4/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp.o
[5/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/mrs_errors.cpp.o
[6/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/audio_frame_observer.cpp.o
[7/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/external_video_track_source_interop.cpp.o
[8/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/peer_connection_interop.cpp.o
[9/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/pch.cpp.o
[10/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp.o
[11/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp.o
[12/17] Building CXX object CMakeFiles/mrwebrtc.dir/jni_onload.cpp.o
[13/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/sdp_utils.cpp.o
[14/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp.o
[15/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp.o
[16/17] Building CXX object CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/video_frame_observer.cpp.o
[17/17] Linking CXX shared library /home/astaikos/eanders/MixedReality-WebRTC/tools/build/android/webrtc-native/build/intermediates/cmake/debug/obj/arm64-v8a/libmrwebrtc.so
FAILED: /home/astaikos/eanders/MixedReality-WebRTC/tools/build/android/webrtc-native/build/intermediates/cmake/debug/obj/arm64-v8a/libmrwebrtc.so 
: && /home/astaikos/Android/Sdk/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64/bin/clang++ --target=aarch64-none-linux-android24 --gcc-toolchain=/home/astaikos/Android/Sdk/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64 --sysroot=/home/astaikos/Android/Sdk/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64/sysroot -fPIC -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security  -std=c++17 -DANDROID_STL=c++_shared -DMRS_USE_STR_WRAPPER -DMR_SHARING_ANDROID -DWEBRTC_POSIX -DWEBRTC_ANDROID -O0 -fno-limit-debug-info  -Wl,--exclude-libs,libgcc_real.a -Wl,--exclude-libs,libatomic.a -static-libstdc++ -Wl,--build-id -Wl,--fatal-warnings -Wl,--no-undefined -Qunused-arguments -shared -Wl,-soname,libmrwebrtc.so -o /home/astaikos/eanders/MixedReality-WebRTC/tools/build/android/webrtc-native/build/intermediates/cmake/debug/obj/arm64-v8a/libmrwebrtc.so CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/external_video_track_source_interop.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/local_video_track_interop.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/peer_connection_interop.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/interop_api.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/audio_frame_observer.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/data_channel.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/mrs_errors.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/pch.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/sdp_utils.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/str.cpp.o CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/video_frame_observer.cpp.o CMakeFiles/mrwebrtc.dir/jni_onload.cpp.o  -llog -landroid -ldl -lOpenSLES /home/astaikos/eanders/build_webrtc/webrtc/src/out/android/arm64/Release/obj/libwebrtc.a -latomic -lm && :
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp.o: In function `PeerConnection':
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp:1370: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::TrackedObject(Microsoft::MixedReality::WebRTC::RefPtr<Microsoft::MixedReality::WebRTC::GlobalFactory>, Microsoft::MixedReality::WebRTC::ObjectType)'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp.o: In function `~PeerConnection':
/home/astaikos/eanders/MixedReality-WebRTC/tools/build/android/webrtc-native/src/main/cpp/../../../../../../../libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.h:65: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::~TrackedObject()'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp.o:(.data.rel.ro._ZTIN9Microsoft12MixedReality6WebRTC14PeerConnectionE[_ZTIN9Microsoft12MixedReality6WebRTC14PeerConnectionE]+0x10): undefined reference to `typeinfo for Microsoft::MixedReality::WebRTC::TrackedObject'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp.o: In function `ExternalVideoTrackSource':
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp:287: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::TrackedObject(Microsoft::MixedReality::WebRTC::RefPtr<Microsoft::MixedReality::WebRTC::GlobalFactory>, Microsoft::MixedReality::WebRTC::ObjectType)'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp.o: In function `~ExternalVideoTrackSource':
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.h:79: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::~TrackedObject()'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp.o:(.data.rel.ro._ZTIN9Microsoft12MixedReality6WebRTC24ExternalVideoTrackSourceE[_ZTIN9Microsoft12MixedReality6WebRTC24ExternalVideoTrackSourceE]+0x10): undefined reference to `typeinfo for Microsoft::MixedReality::WebRTC::TrackedObject'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp.o: In function `LocalVideoTrack':
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp:18: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::TrackedObject(Microsoft::MixedReality::WebRTC::RefPtr<Microsoft::MixedReality::WebRTC::GlobalFactory>, Microsoft::MixedReality::WebRTC::ObjectType)'
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp:27: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::~TrackedObject()'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp.o: In function `~LocalVideoTrack':
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp:35: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::~TrackedObject()'
/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp:35: undefined reference to `Microsoft::MixedReality::WebRTC::TrackedObject::~TrackedObject()'
CMakeFiles/mrwebrtc.dir/home/astaikos/eanders/MixedReality-WebRTC/libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.cpp.o:(.data.rel.ro._ZTIN9Microsoft12MixedReality6WebRTC15LocalVideoTrackE+0x28): undefined reference to `typeinfo for Microsoft::MixedReality::WebRTC::TrackedObject'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.

djee-ms: I fixed that like 5 min ago, this was due to the merge of `master`, you're too fast... 😉

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/LocalAudioSource.cs
djee-ms: Deleted.

tools/build/android/README.md
HyperLethalVector: I'm also attempting to build for android, via android studio. but I'm coming across the following error:
In file included from C:\MixedRealityRTCLib\MixedReality-WebRTC\libs\Microsoft.MixedReality.WebRTC.Native\src\media\external_video_track_source.cpp:4:
C:\MixedRealityRTCLib\MixedReality-WebRTC\libs\Microsoft.MixedReality.WebRTC.Native\src\pch.h:44:10: fatal error: 'api/audio_codecs/builtin_audio_decoder_factory.h' file not found

is there some include directive that is missing? I've pulled from the 'master'  branch.

PR #194, opened 2020-02-19
NodeDssSignaler: CO_GetAndProcessFromServer(): set timeSincePollMs = …

simonchiarel:
…PollTimeMs - 1f after receiving a valid message so that the next message is received asap.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/NodeDssSignaler.cs
+++ b/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/NodeDssSignaler.cs
@@ -249,7 +249,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity
                             break;
                     }
 
-					timeSincePollMs = PollTimeMs - 1f; //fast forward next request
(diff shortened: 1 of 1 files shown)


libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/NodeDssSignaler.cs
```diff
@@ -248,7 +248,9 @@ private IEnumerator CO_GetAndProcessFromServer()
                             Debug.Log("Unknown message: " + msg.MessageType + ": " + msg.Data);
                             break;
                     }
-                }
+
+					timeSincePollMs = PollTimeMs - 1f; //fast forward next request
```
djee-ms: This should be `timeSincePollMs = PollTimeMs + 1f` to ensure the early-out check in `Update()` will always fail, no? Otherwise you will still waste 1 frame in `Update()` to add the frame time, and only after that will the early-out check fail on next frame and the polling execute.

Also can you please fix the tab => space?

simonchiarel: Yes, ideed.

PR #196, opened 2020-02-20
2-step external video track creation for callbacks

djee-ms:
Split creation of external video tracks into 2 steps, the first creating the track object, and the second finalizing the creation and starting the track source capture.

In the current state of the API, the frame callbacks are owned by the peer connection, so there is little difference. However with the upcoming change adding standalone track objects, this change allows the user to register a frame callback before capture starts, and therefore be able to honor any frame request from the first one, while also keeping the WebRTC restriction that track sources must start in the capturing state.

This is only a breaking change if directly using the low-level C interop API. The C# library and Unity integration are not affected, as they will handle the change internally and transparently for the user.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp
@@ -118,12 +118,10 @@ constexpr const size_t kMaxPendingRequestCount = 64;
 RefPtr<ExternalVideoTrackSource> ExternalVideoTrackSourceImpl::create(
     std::unique_ptr<BufferAdapter> adapter) {
   auto source = new ExternalVideoTrackSourceImpl(std::move(adapter));
-
-  // Video track sources start already capturing; there is no start/stop
-  // mechanism at the track level in WebRTC. A source is either being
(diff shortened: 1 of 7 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/media/external_video_track_source.cpp
```diff
@@ -141,6 +139,10 @@ ExternalVideoTrackSourceImpl::~ExternalVideoTrackSourceImpl() {
   StopCapture();
   GlobalFactory::Instance()->RemoveObject(ObjectType::kExternalVideoTrackSource,
                                           this);
+}
+
+void ExternalVideoTrackSourceImpl::FinishCreation() {
```
stephenatwork: Do we need the extra concept of finishing creation? Will it do anything other than call the existing StartCapture?

djee-ms: We do not want to expose the `StartCapture()` concept because all video track sources start in the capturing state in WebRTC, so I think we shouldn't return to the user a paused/stopped track source to avoid them creating a track from it. So we hide that under the "FinishCreation" concept.

PR #197, opened 2020-02-20
Make GlobalFactory thread-safe to initialize

djee-ms:
Make `GlobalFactory` thread-safe to initialize and shutdown by using reference counting.

Previously `RemoveObject()` was attempting to destroy the `GlobalFactory` singleton instance as soon as the number of tracked objects alive reached zero. This was problematic for two reasons:
1. If thread A retrieves a pointer to the singleton before thread B, then thread B creates a new wrapper (which calls `AddObject()`) but later fails to complete the initializing of that object and has to destroy it before thread A itself added an object, then `RemoveObject()` (called from the wrapper destructor) will find no object alive and will destroy the `GlobalFactory` singleton instance, which leaves thread A with a dangling pointer.
2. Destroying the singleton instance was done outside the lock, and access to that pointer was therefore prone to concurrent writes.

This change introduces a different approach where the singleton instance is created on first use via member function static initializing, which is thread-safe in C++11, and never destroyed until static de-initializing at the end of the process or when the module (DLL) is unloaded. Instead, reference counting and number of tracked objects alive are used to initialize and shutdown the singleton instance, which now can be reused after shutdown.

The addition of reference counting, and returning a proper `RefPtr<GlobalFactory>` instead of a unique pointer, make it easier to understand for users that the `RefPtr` instance will prevent the library from shutting down while it is in scope, and therefore that keeping that `RefPtr` around while creating a wrapper object prevents the first issue described above since having no tracked object alive is necessary but not sufficient anymore to trigger a library shutdown.

NB: To improve on this change, another subsequent PR will actually add a `RefPtr<GlobalFactory>` member to all tracked objects, making the reference count the sole mechanism to decide whether the factory needs to be shut down or not, leaving `alive_objects_` as a simple debug-only container. This will also cut down the number of mutex locks by allowing a tracked object to pass a copy of its own `RefPtr` to a new tracked object it created without the need to check if the library needs to be initialized (since it obviously is initialized already). For clarity that additional change is not added to this PR and will be reviewed separately.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -21,18 +21,37 @@ enum class mrsBool : int32_t { kTrue = -1, kFalse = 0 };
 enum class mrsShutdownOptions : uint32_t {
   kNone = 0,
 
-  /// Fail to shutdown if some objects are still alive. This is set by default
-  /// and provides safety against deadlocking, since WebRTC calls are proxied on
-  /// background threads which would be destroyed by the shutdown, causing
-  /// objects still alive to deadlock when making such calls. Note however that
-  /// keeping the library alive, and in particular the WebRTC background
-  /// threads, means that the module (DLL) cannot be unloaded, which might be
-  /// problematic in some use cases (e.g. Unity Editor hot-reload).
-  kFailOnLiveObjects = 0x1,
-
   /// Log some report about live objects when trying to shutdown, to help
-  /// debugging.
-  kLogLiveObjects = 0x2
+  /// debugging. This flag is set by default.
+  kLogLiveObjects = 0x1,
+
+  /// Advanced use only. Not set by default.
+  ///
+  /// Determine the behavior of the library on abnormal shutdown, when some
+  /// objects are still alive yet the internal factory is being destroyed, which
+  /// generally indicates some reference counting leaks. This flag has no effect
+  /// under normal use when the library is properly shut down after all objects
+  /// have been destroyed.
+  ///
+  /// - When cleared (default), this flag leaves the library in a dangling state
+  /// where the background WebRTC threads continue running forever. This means
+  /// WebRTC objects can continue to be proxied to those threads, and will not
+  /// deadlock waiting for them. However this means that the module (DLL) cannot
+  /// be unloaded as threads are still running, which might be problematic in
+  /// some use cases (e.g. Unity Editor hot-reload).
+  ///
+  /// - When set, this flag allows shutting down the library even if some
+  /// objects are still alive. This should be used with care as this makes the
+  /// library prone to deadlocking, since WebRTC calls are proxied on background
+  /// threads which would be destroyed by the shutdown, causing objects still
+  /// alive to deadlock when making such calls. However it might prove useful in
+  /// a controlled context where such objects are ensured not to be used any
+  /// further, like a unit test failing, to allow shutting down all threads and
+  /// forcefully unloading the module (DLL).
+  kIgnoreLiveObjectsAndForceShutdown = 0x2,
+
+  /// Default flags value.
+  kDefault = kLogLiveObjects
 };
 
 /// Opaque enumerator type.
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -25,31 +25,6 @@ enum class mrsShutdownOptions : uint32_t {
   /// debugging. This flag is set by default.
   kLogLiveObjects = 0x1,
 
-  /// Advanced use only. Not set by default.
-  ///
-  /// Determine the behavior of the library on abnormal shutdown, when some
-  /// objects are still alive yet the internal factory is being destroyed, which
-  /// generally indicates some reference counting leaks. This flag has no effect
-  /// under normal use when the library is properly shut down after all objects
-  /// have been destroyed.
-  ///
-  /// - When cleared (default), this flag leaves the library in a dangling state
-  /// where the background WebRTC threads continue running forever. This means
-  /// WebRTC objects can continue to be proxied to those threads, and will not
-  /// deadlock waiting for them. However this means that the module (DLL) cannot
-  /// be unloaded as threads are still running, which might be problematic in
-  /// some use cases (e.g. Unity Editor hot-reload).
-  ///
-  /// - When set, this flag allows shutting down the library even if some
-  /// objects are still alive. This should be used with care as this makes the
-  /// library prone to deadlocking, since WebRTC calls are proxied on background
-  /// threads which would be destroyed by the shutdown, causing objects still
-  /// alive to deadlock when making such calls. However it might prove useful in
-  /// a controlled context where such objects are ensured not to be used any
-  /// further, like a unit test failing, to allow shutting down all threads and
-  /// forcefully unloading the module (DLL).
-  kIgnoreLiveObjectsAndForceShutdown = 0x2,
-
   /// Default flags value.
   kDefault = kLogLiveObjects
 };
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -17,6 +17,24 @@ enum class mrsBool : int32_t { kTrue = -1, kFalse = 0 };
 // Generic utilities
 //
 
+/// Global MixedReality-WebRTC library shutdown options.
+enum class mrsShutdownOptions : uint32_t {
+  kNone = 0,
+
+  /// Fail to shutdown if some objects are still alive. This is set by default
+  /// and provides safety against deadlocking, since WebRTC calls are proxied on
+  /// background threads which would be destroyed by the shutdown, causing
+  /// objects still alive to deadlock when making such calls. Note however that
+  /// keeping the library alive, and in particular the WebRTC background
+  /// threads, means that the module (DLL) cannot be unloaded, which might be
+  /// problematic in some use cases (e.g. Unity Editor hot-reload).
+  kFailOnLiveObjects = 0x1,
+
+  /// Log some report about live objects when trying to shutdown, to help
+  /// debugging.
+  kLogLiveObjects = 0x2
+};
+
 /// Opaque enumerator type.
 struct mrsEnumerator;
 
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
@@ -13,12 +13,6 @@ namespace {
 
 using namespace Microsoft::MixedReality::WebRTC;
 
-/// Global factory of all global objects, including the peer connection factory
-/// itself, with added thread safety. This keeps a track of all objects alive,
-/// to determine when it is safe to release the WebRTC threads, thereby allowing
-/// a DLL linking this code to be unloaded.
-std::unique_ptr<GlobalFactory> g_factory = std::make_unique<GlobalFactory>();
-
 /// Utility to convert an ObjectType to a string, for debugging purpose. This
 /// returns a view over a global constant buffer (static storage), which is
 /// always valid, never deallocated.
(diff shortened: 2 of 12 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
```diff
@@ -131,14 +135,33 @@ void GlobalFactory::RemoveObject(ObjectType type, TrackedObject* obj) noexcept {
...
+  std::unique_ptr<GlobalFactory>& factory = GlobalFactory::MutableInstance();
+  if (!factory) {
+    return false;  // not shutdown by this call
+  }
+  try {
+    if (factory->TryShutdownImpl()) {
+      factory = nullptr;  // fixme : outside lock...
+      return true;
+    }
+  } catch (...) {
```
stephenatwork: Why not let the exception go to the caller?

djee-ms: The internal C++ classes are only helpers to implement the C interop API, which cannot throw. So better catch it here than add try/catch block in 20 places in the C functions. Ideally we'd disable exceptions completely in this project.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -17,6 +17,24 @@ enum class mrsBool : int32_t { kTrue = -1, kFalse = 0 };
 // Generic utilities
 //
 
+/// Global MixedReality-WebRTC library shutdown options.
+enum class mrsShutdownOptions : uint32_t {
+  kNone = 0,
```
fibann: nit: Probably needs a comment mentioning possible deadlocks. Also I find a bit surprising that 0 is not the default - should be mentioned too (or the flags be changed so that 0 is the default).

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
stephenatwork: The one place in this review which calls TryShutdown is ignoring the return value. That looks suspicious.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -16,24 +16,59 @@ enum class ObjectType : int {
...
+///
+/// The delayed call to |TryShutdown()| from releasing the temporary reference
+/// when RefPtr<GlobalFactory> goes out of scope instead of when no object is
+/// alive anymore allows the caller to create a wrapper object and fail
+/// initializing it, and remove it from the global factory but without the
+/// factory destroying itself yet. This works around a problem where the object
+/// removing itself triggers the destruction of the global factory singleton
+/// instance while another thread already acquired a reference to the global
+/// factory but did not yet add any wrapper object so cannot prevent its
+/// destruction, and is left with a dangling reference to a destroyed singleton.
```
fibann: nit: I'd avoid mentioning an issue that cannot happen anymore 😄 Maybe move this to the commit/PR message?

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -47,16 +82,38 @@ class GlobalFactory {
...
+  GlobalFactory& operator=(const GlobalFactory&) = delete;
+  static std::unique_ptr<GlobalFactory>& MutableInstance(
+      bool createIfNotExist = true);
+  mrsResult InitializeImplNoLock();
+  void ForceShutdownImpl();
+  bool TryShutdownImpl();
+  void ReportLiveObjectsNoLock();
 
  private:
   rtc::scoped_refptr<webrtc::PeerConnectionFactoryInterface> factory_
```
stephenatwork: peerfactory_? Confusing given that the container class is also a factory

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -16,24 +16,59 @@ enum class ObjectType : int {
...
+/// removing itself triggers the destruction of the global factory singleton
+/// instance while another thread already acquired a reference to the global
+/// factory but did not yet add any wrapper object so cannot prevent its
+/// destruction, and is left with a dangling reference to a destroyed singleton.
+/// The temporary reference count allows delaying that destruction for as long
+/// as any thread has a reference to the singleton instance. This is not a real
+/// reference count because the global factory is a singleton, and to avoid
+/// circular references. Users must never store a RefPtr<> to the global
+/// factory, but must use a local variable instead to temporarily block
+/// destruction during a single API call.
```
fibann: s/temporary reference count/temporary reference
(the RefPtr holds a "reference", while the "reference count" is the _number_ of references)


> This is not a real reference count because the global factory is a singleton, and to avoid circular references

Not incredibly clear, I'd rephrase as
> Note that, after any wrapper is added, the factory will be kept alive by the wrapper itself, so it is not ever necessary to store a RefPtr to the factory (in fact, it might cause circular references). As a rule, users should never store...

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -16,24 +16,59 @@ enum class ObjectType : int {
...
   /// factory itself, with added thread safety.
-  static const std::unique_ptr<GlobalFactory>& Instance();
+  static RefPtr<GlobalFactory> InstancePtr();
 
-  /// Get or create the peer connection factory.
-  rtc::scoped_refptr<webrtc::PeerConnectionFactoryInterface> GetOrCreate();
+  /// Attempt to shutdown the global factory if no live object is present
+  /// anymore. This is always conservative and safe, and will do nothing if any
+  /// object is still live.
+  static bool TryShutdown() noexcept;
```
fibann: nit: document the return value.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
```diff
@@ -240,7 +263,52 @@ mrsResult GlobalFactory::Initialize() {
...
+
+  // Shutdown
+  factory_ = nullptr;
+#if defined(WINUWP)
+  impl_ = nullptr;
+#else   // defined(WINUWP)
+  network_thread_.reset();
+  worker_thread_.reset();
+  signaling_thread_.reset();
+#endif  // defined(WINUWP)
```
fibann: nit: deduplicate with TryShutdownImpl

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -73,6 +130,17 @@ class GlobalFactory {
   /// Collection of all objects alive.
   std::unordered_map<TrackedObject*, ObjectType> alive_objects_
       RTC_GUARDED_BY(mutex_);
+
+  /// Shutdown options.
+  mrsShutdownOptions shutdown_options_;
```
fibann: This seems to be uninitialized?

Also, mrsShutdownOptions is defined in interop_api.h - does it mean there's going to be an interface to set it?

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
djee-ms: Ah you meant that one method in particular, not in general why we have try/catch blocks? Actually `RemoveRef()` is a member method so should call `TryShutdownImpl()` directly. I will remove `TryShutdown()`.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: Yes, it was not in this change, but I still needed it to be defined.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: deleted

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Added a `kDefault` to make things explicit, and inverted the meaning to have the flag set activating the dangerous behavior.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: Rephrased into user instructions, is that better? There is a strict usage pattern to follow here otherwise things get unsafe, so I'd rather have a mention of the deadlock somewhere.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: It's indirectly by the size of `alive_objects_`; there is no direct `RefPtr<>` inside the wrappers.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
djee-ms: Done

djee-ms:
As it turns out the unsafe shutdown is more problematic than previously observed, and after this change was sometimes crashing some tests. So doing a new review with a new approach that should fix this.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
```diff
@@ -19,82 +20,112 @@ enum class ObjectType : int {
...
+  /// multithread-safe, and is generally called automatically by |RefPtr<>|.
+  void RemoveRef() const noexcept {
+    // Update the reference count under the init lock to ensure it cannot change
+    // from another thread, since invoking |AddRef()| requires having already a
+    // pointer to the GlobalFactory (so this reference would never be the last
+    // one) or calling |GetLockImpl()| to get a new pointer, which will only
+    // call |AddRef()| under the lock too so will block.
+    std::scoped_lock lock(init_mutex_);
+    RTC_DCHECK(peer_factory_);
+    if (ref_count_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
```
fibann: If this is under the lock does it need to be acq_rel?

djee-ms: We still want I think to prevent code from being reordered, especially with the mutex lock above. From a dumb look at it those two lines look independent, so I am not sure the compiler or out-of-order CPU execution would be smart enough to not move the `fetch_sub` outside the mutex lock above.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
```diff
@@ -54,61 +49,95 @@ std::string ObjectToString(ObjectType type, TrackedObject* obj) {
...
+bool GlobalFactory::TryShutdown() noexcept {
+  GlobalFactory* const factory = GetInstance();
+  std::scoped_lock lock(factory->init_mutex_);
+  if (!factory->peer_factory_) {
+    return true;
+  }
+  try {
+    return factory->ShutdownImplNoLock(ShutdownAction::kTryShutdownIfSafe);
+  } catch (...) {
+    return false;  // failed to shutdown
```
fibann: nit: should we try to print out what failed here?

djee-ms: You mean in the catch() block? Not sure what we can print...

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.h
djee-ms: Discussed offline, changed to relaxed, this seems enough.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
djee-ms: Discussed offline, printing error message for `std::exception`, as WebRTC code shouldn't throw.

PR #201, opened 2020-02-26
Expose some library-wide functionalities

djee-ms:
Expose:
- Set/get shutdown options
- Report live objects
- Force shutdown


libs/Microsoft.MixedReality.WebRTC/Library.cs
```diff
@@ -0,0 +1,76 @@
...
+            /// <summary>
+            /// Default options.
+            /// </summary>
+            Default = LogLiveObjects
+        }
+
+        /// <summary>
+        /// Get the options used when shutting down the MixedReality-WebRTC library.
+        /// </summary>
+        public static ShutdownOptions GetShutdownOptions()
```
fibann: nit: property?

Issue #152, opened 2019-12-18
Add standalone media track objects


djee-ms:
Currently in 1.0 the `PeerConnection` object supports a single local/remote audio/video track (so 4 in total). This corresponds to the simple video chat use case of a single A/V feed sent and received. In `master` a standalone `LocalVideoTrack` object has been added, and with it the ability to open multiple local video tracks per peer connection (see #35). However because a single remote video track is supported, there is currently no way to receive multiple video tracks on the remote peer. And for audio multi-tracks is simply not supported by a single peer connection (multiple connections can be used as a workaround).

This feature introduces some standalone media tracks for all 4 categories of local/remote audio/video tracks, to effectively enable multi-track support per peer connection.

Issue #75, opened 2019-09-17
Remove deprecated offer_to_receive_xxx options


djee-ms:
From #51:
> Not sure about if it has any effect on performance, it could also be interesting to be able to modify lines 275-276 and 291-291 in peer_connection.cpp
>
> options.offer_to_receive_audio = true;
options.offer_to_receive_video = true;
>
> Now this is a special case for my implementation, but I only either send OR receive video/audio on each peer connection, so previously to this plugin, I modified offer_to_receive_audio/video to false depending on the connection. I also changed sendrecv in the sdp to sendonly or recvonly depending on the use-case.

Those options are deprecated anyway, so should be removed entirely. This mainly needs testing.

djee-ms:
Note to self : this changes the behavior of TestAppUWP in that creating an offer without any track will prevent the remote peer from sending its audio/video track, so TestAppUWP should be updated to add some transceivers in advance to accept those audio/video tracks, and there's no API for that at the moment.

PR #202, opened 2020-02-26
Add transceiver API to interop

djee-ms:
This is part 1/N of the multi-track change : interop library.

## Add transceiver API to interop library

This change adds a new transceiver API to the interop library allowing the user to directly manipulate transceivers, which unlocks adding multiple audio and video tracks to a single peer connection. This API extends and replaces the existing track-based API previously available on the peer connection.

From a user perspective, the API divides the actions of creating the tracks from the one of attaching them to the peer connection via the newly-added _transceiver_ concept.

### Transceiver

A _transceiver_ represents a "media pipe" between the local and remote peers, through which some audio or video data is sent. This concept is nearly identical to the same-named concept present in the Unified Plan semantic of the SDP protocol.

This transceiver has a _direction_: it can send and/or receive media data at the same time (bidirectional), or can be inactive.
- When sending, the transceiver obtain its data from an attached local track object, or sends default empty data (black frame for video, silence for audio) if not track is attached.
- When receiving, the transceiver provide its data to a remote track object it creates and owns, and which the user can use to interact with the data (register a frame callback).

The transceiver transparently encode the data on the sender side and decode it on the receiver side, and this codec is associated with the transceiver itself and negotiated during a new SDP session negotiation.

Transceivers can be added to a peer connection with `mrsPeerConnectionAddAudioTransceiver()` and `mrsPeerConnectionAddVideoTransceiver()`. Once added, transceivers remain on that peer connection forever, and cannot be removed. This generates a renegotiation needed event, and requires a new SDP negotiation to become usable and start transferring media data between the two peers. Changing the transceiver direction also requires a renegotiation.

_NB : In the WebRTC 1.0 standard, transceivers can be stopped. This feature is not available in the current API. Transceivers can only be set to inactive._

To attach a local track to a transceiver, use `mrsTransceiverSetLocalTrack()`. To remove it, pass `nullptr` (invalid handle) as the track handle. Unlike changing its direction, swapping a track on an existing transceiver is a transparent operation which do not require a renegotiation.

Remote tracks are automatically created/destroyed when the transceiver starts/stops receiving data, respectively, therefore cannot be explicitly attached to nor detached from it.

### Standalone track objects

Track objects are standalone objects that the user can obtain a reference to and manipulate. The creation/destruction process and lifetime of tracks depends on the directionality of the media data:

- **Local tracks**, which produce some data locally to be sent to the remote peer (senders), are owned by the user and created via functions such as `mrsLocalVideoTrackCreateFromDevice()` or `mrsLocalVideoTrackCreateFromExternalSource()`.

- **Remote tracks**, which consume some data received from the remote peer to make it available locally, are owned by the transceiver receiving the data. The user is notified of their creation/destruction via the `TrackAdded`/`TrackRemoved` events of the peer connection (see _e.g._ `mrsPeerConnectionRegisterVideoTrackAddedCallback()` to install handlers for those events).

### Transceiver pairing

Transceivers on the local and remote peer are paired with each other either based on their index in the transceiver list of the peer connection, that is the order in which they are added to it, or via their unique name. Their index is referred to as the _media line index_ (or _mline index_), and maps 1:1 with the order in which the `m=` lines are present in the SDP offer/answer messages. These properties are derived directly from the properties of a transceiver as defined in the WebRTC 1.0 standard.

One important internal change required for transceiver pairing to work is to remove the legacy options `offer_to_receive_video = true` and `offer_to_receive_audio = true` of `webrtc::PeerConnectionInterface::RTCOfferAnswerOptions` from the implementation of `mrsPeerConnectionCreateOffer()` and `mrsPeerConnectionCreateAnswer()`. This change (see #75) has larger implications than it looks when using the Unified Plan SDP semantic (default), as previously those flags were implicitly making an offer to receive some media, whereas this step is now explicit and the user needs to add some transceiver and set its direction to receive-only or send-receive before making an offer/answer to allow the remote peer to send. Otherwise the remote peer will not be allowed to send any media.

Bug: #152


libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -70,23 +70,112 @@ MRS_API void MRS_CALL mrsCloseEnum(mrsEnumHandle* handleRef) noexcept;
...
+using RemoteAudioTrackHandle = void*;
+
+/// Opaque handle to a native RemoteVideoTrack C++ object.
+using RemoteVideoTrackHandle = void*;
+
+/// Opaque handle to a native DataChannel C++ object.
+using DataChannelHandle = void*;
+
+/// Opaque handle to a native ExternalVideoTrackSource C++ object.
+using ExternalVideoTrackSourceHandle = void*;
```
fibann: Should either have a prefix or be inside a namespace.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -70,23 +70,116 @@ MRS_API void MRS_CALL mrsCloseEnum(mrsEnumHandle* handleRef) noexcept;
...
+/// Opaque handle to a native DataChannel interop object.
+using mrsDataChannelHandle = void*;
+
+/// Opaque handle to a native ExternalVideoTrackSource interop object.
+using mrsExternalVideoTrackSourceHandle = void*;
+
+// Handles to wrapper objects
+
 /// Opaque handle to the interop wrapper of a peer connection.
 using mrsPeerConnectionInteropHandle = void*;
```
fibann: nit: I find this thing of two "handles" for each object going through the API really confusing. The fact that the handles for wrappers are called "interop handles" while not being created/managed by the interop library compounds the problem.

IMHO we should either pass just one kind of handles through the interop API, e.g.:
- the interop lib returns a handle that identifies the native object, and the layer above keeps a mapping between this handle and its wrapper
- or, the interop lib takes a wrapper id that identifies the object in the wrapper layer, and keeps a mapping between it and the corresponding native object.

If none of the above is possible, at the very least we should change the naming so that the native "thing" is called differently from the wrapper "thing" and they are not just "handles".

Does not need to be fixed now, happy to move this to an issue to be discussed separately.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -329,93 +450,135 @@ struct PeerConnectionConfiguration {
...
+
+  /// Construct an interop object for a RemoteAudioTrack instance.
+  mrsRemoteAudioTrackCreateObjectCallback remote_audio_track_create_object{};
+
+  /// Construct an interop object for a RemoteVideooTrack instance.
+  mrsRemoteVideoTrackCreateObjectCallback remote_video_track_create_object{};
+
   /// Construct an interop object for a DataChannel instance.
   mrsDataChannelCreateObjectCallback data_channel_create_object{};
 };
```
fibann: Why do transceivers have a "FinishCreate" callback and tracks don't? Seems strangely asymmetric.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -209,26 +290,66 @@ enum class TrackKind : uint32_t {
...
-
-/// Callback fired when a remote track is removed from a connection.
-using PeerConnectionTrackRemovedCallback =
-    void(MRS_CALL*)(void* user_data, TrackKind track_kind);
+/// Callback fired when a remote audio track is added to a connection.
+/// The |audio_track| and |audio_transceiver| handle hold a reference to the
+/// underlying native object they are associated with, and therefore must be
+/// released with |mrsLocalAudioTrackRemoveRef()| and
+/// |mrsAudioTransceiverRemoveRef()|, respectively, to avoid memory leaks.
+using PeerConnectionAudioTrackAddedCallback =
```
fibann: These should have a prefix too I suppose?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -329,93 +450,135 @@ struct PeerConnectionConfiguration {
...
 
-/// Register a callback fired when a remote media track is removed from the
+/// Register a callback fired when a remote audio track is removed from the
 /// current peer connection.
-MRS_API void MRS_CALL mrsPeerConnectionRegisterTrackRemovedCallback(
-    PeerConnectionHandle peerHandle,
-    PeerConnectionTrackRemovedCallback callback,
+/// Note that the arguments include some object handles, which each hold a
+/// reference to the corresponding object and therefore must be released, even
+/// if the user does not make use of them in the callback.
```
fibann: Why does the callback implementation have to be made responsible of releasing the reference?

nit: I'd rather mention the arguments that need to be released specifically.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -431,24 +594,77 @@ enum class VideoProfileKind : int32_t {
...
+  /// Mandatory name of the audio transceiver. This must be a valid SDP token;
+  /// see |mrsSdpIsValidToken()|.
+  const char* name = nullptr;
+
+  /// Initial desired direction of the transceiver media when created.
+  mrsTransceiverDirection desired_direction =
+      mrsTransceiverDirection::kSendRecv;
+
+  /// Semi-colon separated list of stream IDs associated with the transceiver.
+  /// Use |Transceiver::DecodeStreamIDs()| and |Transceiver::EncodeStreamIDs()|.
```
fibann: nit: shouldn't mention private functions in the public API.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -502,71 +718,76 @@ struct LocalVideoTrackInitConfig {
...
-MRS_API mrsResult MRS_CALL mrsPeerConnectionRemoveLocalVideoTracksFromSource(
-    PeerConnectionHandle peer_handle,
-    ExternalVideoTrackSourceHandle source_handle) noexcept;
-
-/// Add a local audio track from a local audio capture device (microphone) to
-/// the collection of tracks to send to the remote peer.
-MRS_API mrsResult MRS_CALL
-mrsPeerConnectionAddLocalAudioTrack(PeerConnectionHandle peerHandle) noexcept;
+/// Configuration for creating a new video transceiver interop wrapper.
+struct mrsVideoTransceiverConfig {
```
fibann: nit: merge with above, just use mrsTransceiverConfig?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -431,24 +594,77 @@ enum class VideoProfileKind : int32_t {
...
+  const char* stream_ids = nullptr;
+
+  /// Handle of the video transceiver interop wrapper, if any, which will be
+  /// associated with the native video transceiver object.
+  mrsVideoTransceiverInteropHandle transceiver_interop_handle{};
+};
+
+/// Configuration for opening a local audio capture device and creating a local
+/// audio track.
+struct LocalAudioTrackInitConfig {
```
fibann: nit: move track name in here?
Move this to the local_audio_track header?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -70,23 +70,116 @@ MRS_API void MRS_CALL mrsCloseEnum(mrsEnumHandle* handleRef) noexcept;
...
+
+/// Opaque handle to the interop wrapper of a remote video track.
+using mrsRemoteVideoTrackInteropHandle = void*;
+
 /// Opaque handle to the interop wrapper of a data channel.
 using mrsDataChannelInteropHandle = void*;
 
+/// Callback to create an interop wrapper for an audio transceiver.
+/// The callback must return the handle of the created interop wrapper.
+using mrsAudioTransceiverCreateObjectCallback =
```
fibann: It might be worth explaining _when_ each of these "CreateObject" callbacks is called (which I assume is when an object is created following an action by the remote peer, but I am not so sure because there are many!)

libs/Microsoft.MixedReality.WebRTC.Native/include/remote_audio_track_interop.h
```diff
@@ -0,0 +1,43 @@
...
+
+/// Register a custom callback to be called when the local audio track received
+/// a frame.
+MRS_API void MRS_CALL
+mrsRemoteAudioTrackRegisterFrameCallback(mrsRemoteAudioTrackHandle trackHandle,
+                                         mrsAudioFrameCallback callback,
+                                         void* user_data) noexcept;
+
+/// Enable or disable a local audio track. Enabled tracks output their media
+/// content as usual. Disabled track output some void media content (silent
```
fibann:
```suggestion
/// content as usual. Disabled tracks output some void media content (silent

```

libs/Microsoft.MixedReality.WebRTC.Native/include/remote_audio_track_interop.h
```diff
@@ -0,0 +1,43 @@
...
+/// audio frames). Enabling/disabling a track is a lightweight concept similar
+/// to "mute", which does not require an SDP renegotiation.
```
fibann:
```suggestion
/// Enable or disable a remote audio track. Enabled tracks output their media
/// content as usual. Disabled tracks output some void media content (silent
/// audio frames). Enabling/disabling a track is a lightweight concept similar
/// to "mute", which does not require an SDP renegotiation.

```

libs/Microsoft.MixedReality.WebRTC.Native/include/remote_video_track_interop.h
```diff
@@ -0,0 +1,51 @@
...
+/// Register a custom callback to be called when the remote video track received
+/// a frame. The received frames is passed to the registered callback in ARGB32
+/// encoding.
+MRS_API void MRS_CALL mrsRemoteVideoTrackRegisterArgb32FrameCallback(
+    mrsRemoteVideoTrackHandle trackHandle,
+    mrsArgb32VideoFrameCallback callback,
+    void* user_data) noexcept;
+
+/// Enable or disable a remote video track. Enabled tracks output their media
+/// content as usual. Disabled track output some void media content (black video
```
fibann:
```suggestion
/// content as usual. Disabled tracks output some void media content (black video

```

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
```diff
@@ -21,10 +21,20 @@ using namespace Microsoft::MixedReality::WebRTC;
 /// always valid, never deallocated.
 std::string_view ObjectTypeToString(ObjectType type) {
```
fibann: nit: a switch might be simpler 😄

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -209,26 +290,66 @@ enum class TrackKind : uint32_t {
...
 };
 
-/// Callback fired when a remote track is added to a connection.
-using PeerConnectionTrackAddedCallback = void(MRS_CALL*)(void* user_data,
-                                                         TrackKind track_kind);
-
-/// Callback fired when a remote track is removed from a connection.
-using PeerConnectionTrackRemovedCallback =
-    void(MRS_CALL*)(void* user_data, TrackKind track_kind);
+/// Callback fired when a remote audio track is added to a connection.
```
stephenatwork: Are these for back compat only? Or a shortcut? I'm surprised to see trackadded callbacks on the peerconnection because if I understand correctly they can only be added via a tranciever.
I expected to see PeerConnection.TranceiverAdded callback and Tranceiver.TrackAdded.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
stephenatwork: +1. I expect there's a good reason, but it it pollutes e.g the trackadded signatures.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_audio_track.h
```diff
@@ -0,0 +1,116 @@
...
+
+class PeerConnection;
+class AudioTransceiver;
+
+/// A local audio track is a media track for a peer connection backed by a local
+/// source, and transmitted to a remote peer.
+///
+/// The local nature of the track implies that the local peer has control on it,
+/// including enabling or disabling the track, and removing it from the peer
+/// connection. This is in contrast with a remote track	reflecting a track sent
```
fibann:
```suggestion
/// connection. This is in contrast with a remote track reflecting a track sent

```

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_audio_track.h
```diff
@@ -0,0 +1,116 @@
...
+/// by the remote peer, for which the local peer has limited control.
+///
+/// The local audio track is backed by a local audio track source. This is
+/// typically a audio capture device (e.g. webcam), but can	also be a source
```
fibann:
```suggestion
/// typically a audio capture device (e.g. webcam), but can also be a source

```

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/audio_transceiver_interop.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+void MRS_CALL mrsAudioTransceiverRegisterStateUpdatedCallback(
+    mrsAudioTransceiverHandle handle,
+    mrsAudioTransceiverStateUpdatedCallback callback,
+    void* user_data) noexcept {
+  if (auto transceiver = static_cast<AudioTransceiver*>(handle)) {
+    transceiver->RegisterStateUpdatedCallback(
+        Transceiver::StateUpdatedCallback{callback, user_data});
+  }
```
stephenatwork: What's the criteria for warning on null handles or not?

libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
```diff
@@ -0,0 +1,142 @@
...
+  return (RtpDir)direction;
+}
+
+Transceiver::Direction Transceiver::FromRtp(
+    webrtc::RtpTransceiverDirection rtp_direction) {
+  using RtpDir = webrtc::RtpTransceiverDirection;
+  static_assert((int)Direction::kSendRecv == (int)RtpDir::kSendRecv, "");
+  static_assert((int)Direction::kSendOnly == (int)RtpDir::kSendOnly, "");
+  static_assert((int)Direction::kRecvOnly == (int)RtpDir::kRecvOnly, "");
+  static_assert((int)Direction::kInactive == (int)RtpDir::kInactive, "");
```
fibann: One copy of these is enough!

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.h
```diff
@@ -34,17 +37,25 @@ class PeerConnection;
 /// typically a video capture device (e.g. webcam), but can	also be a source
```
stephenatwork:
```suggestion
/// typically a video capture device (e.g. webcam), but can also be a source
```

libs/Microsoft.MixedReality.WebRTC.Native/src/media/audio_transceiver.h
```diff
@@ -0,0 +1,73 @@
...
+#include "media/remote_audio_track.h"
+#include "media/transceiver.h"
+#include "refptr.h"
+
+namespace Microsoft::MixedReality::WebRTC {
+
+class PeerConnection;
+
+/// Transceiver containing audio tracks.
+class AudioTransceiver : public Transceiver {
```
fibann: This and VideoTransceiver have the same exact interface, any chance that we can have just one Transceiver class that operated with media tracks, and do the checking at runtime? I don't feel like type-safety is a huge concern here - the chances of mixing audio and video by mistake seem low. If it is, we could make a template and instantiate it once for audio and once for video, so at least we save the copy&paste.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Is it though? You are supposed to provide a callback to create a wrapper, that should be generic enough, without really needing to understand why/when it's needed, no? The implementation will call it when it needs to create a wrapper to associate with a new implementation object, there is not much more to know I think.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Yes I was hesitating if we want to keep it for later extension, but since the implementation mostly uses a single transceiver class I guess it's unlikely they diverge.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: I see only two lifecycle options:
- An object is explicitly created/destroyed via its handle with a specific pair of API calls, and stays alive between those two points, whether or not it is in use. This is usually how handles are used in a standard C API.
- An object is reference-counted, and that reference count is used to drive the destruction, in which case each user must keep a reference to it. This is generally what a C++ API uses, along with a helper like a shared pointer.

Because we started originally by the latter, now each handle holds a reference, and passing through the interop layer a handle without reference makes it very confusing, as now you have handles that do have a reference (_e.g._ returned from a "create" method) and handles that do not (_e.g._ passed as argument to callback), which makes things even more confusing and error-prone in my opinion for the developers.

I do appreciate the current approach is also a bit unusual and so possibly confusing. But what is a better alternative? Using non-reference-counted plain C handles?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: Fair enough. The specific thing I find confusing is that some add/create functions expect the upper layer to pass an interop handle there, and it's not specified if these callbacks will be called in that case or not (if they are, it sounds redundant; if they aren't, we probably should make the distinction in the comments.)

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: The usual pattern that I try to follow for creating objects owned by the user is:
1. Create a wrapper (_e.g._ C#) object via a reverse P/Invoke callback
2. Create the actual low-level implementation object and assign the handle to the wrapper
3. Assign the handle of the newly created implementation object to the wrapper via a second reverse P/Invoke callback if everything went right

For remote tracks here (and only remote tracks; note there's not `FinishCreate` for local ones), those are objects created internally by the Google code and owned by the implementation, not the user (in C# they are not disposable, unlike local tracks). So they are always associated with an event, like `TrackAdded`, whose reverse P/Invoke callback will take care of assigning the handle (step 3. above), so there is no need for an explicit `FinishCreate` for those.

Local tracks seem to still use the old pattern where essentially the steps 1 => 3 above are moved to the wrapping layer (_e.g._ C#), so there is a single (forward) P/Invoke callback for step 2. instead of two reverse P/Invoke callbacks for steps 1 and 3. I don't know what's best between the two approaches.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: No, that's not really how the API works. I know, it's confusing 😉 

Basically the `TrackAdded` and `TrackRemoved` events, which concern only **remote** tracks (receivers), mean that a transceiver started/stopped receiving some media due to an SDP offer/answer being applied. When that happens, a remote track is automatically created/destroyed on the transceiver as needed. Quite often when the transceiver starts receiving, this is a new transceiver that also has just been created by that same offer/answer SDP message. This all happens at the same time during `SetRemoteDescription()`. And the `TrackAdded` event does contain the newly added track **and** the newly added transceiver. `TransceiverAdded` would make less sense, because it would always contain a newly-created receiver track anyway, but that information would not be carried by the name. And if we split the transceiver/track events, then anyway those 2 events would always fire exactly one after the other, which is not very useful.

Note that for local tracks they are not "added" or "removed" on transceivers in the same sense as other objects; instead, a transceiver has a sender "track" slot which can be NULL or any valid local track, and is solely managed by the user. Because of that there is no event associated, since the action is always initiated by the user.

About events naming, this is the official (but outdated) terminology of the Google code, so we can discuss renaming eventually. They were indeed relating to tracks in Plan B, and names remained.

More generally as a mental model things are highly asymmetrical. A naive mental model that I had when reading the standard was that both peers create transceivers, and then the SDP session pairs them 1:1 based on their index (order they have been added). This is wrong and does not work; the actual pairing process is quite a lot more complex, and generally you want the answering peer to wait before adding its transceivers:
1. Peer A adds some transceivers, sets their desired direction (_e.g._ `kSendRecv` or `kSendOnly`).
2. Peer A creates an offer. The offer is based on the existing transceivers and their desired direction at the time of the `CreateOffer()` call.
3. Signaling A => B
4. Peer B applies the offer with `SetRemoteDescription()`. During that time, some transceivers/tracks are created/destroyed based on the offer only. New transceivers are created in `kRecvOnly` mode, always.
5. Peer B optionally adds more transceivers, and/or change the desired direction of the existing ones (_e.g._ from `kRecvOnly` to `kSendRecv`).
6. Peer B creates an answer. Same rules as the offer, except that actual directions advertised are the intersection of the offer and the desired direction, so offer=`kRecvOnly` and desired=`kSendOnly` are incompatible and result in `kInactive`: peer A offered to send, but peer B was only interested in sending too, and refused to receive.
7. Signaling B => A
8. Peer A applies the answer with `SetRemoteDescription()`, same rules as above.

The critical point here is that in step 4. if the answering peer has already some transceivers then there are only 2 possible ways to match those with the `m=` lines from the SDP offer:
- Either both have the same unique name, called `mid`, generated by the implementation;
- Or there is a weird and non-intuitive fallback rule I don't remember and which works only in some cases (in other cases, this rule won't apply, and a new transceiver is created instead).

Because the former requires an external synchronization mechanism (_e.g._ using a data channel, or using the signaling server to exchange some custom messages) to match the `mid` values, and the later is plain confusing and doesn't work in all cases, the best is to start from a clean slate in step 4. (or iterate from a clean previous negotiation). The scheme above is what we do in the Unity integration by the way.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/audio_transceiver_interop.cpp
djee-ms: It is very rare that an API call accepts NULL handles. Generally those are invalid. Only one that comes to mind is changing the transceiver's local sender track to NULL to detach the previous one. So most of the time we emit warnings or return an error `kInvalidNativeHandle`.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_video_track.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_audio_track.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/local_audio_track.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
djee-ms: That's a compile-time switch 😉

libs/Microsoft.MixedReality.WebRTC.Native/include/remote_video_track_interop.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/include/remote_audio_track_interop.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Fixed as `mrsTransceiverInitConfig` to emphasize on creation of a new transceiver.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h

libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.h
```diff
@@ -0,0 +1,184 @@
...
+/// Base class for audio and video transceivers.
+class Transceiver : public TrackedObject {
+ public:
+  using Direction = mrsTransceiverDirection;
+  using OptDirection = mrsTransceiverOptDirection;
+
+  /// Construct a Plan B transceiver abstraction which tries to mimic a
+  /// transceiver for Plan B despite the fact that this semantic doesn't have
+  /// any concept of transceiver.
+  Transceiver(RefPtr<GlobalFactory> global_factory,
```
stephenatwork: named constructor idiom here?

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
fibann: No it's not! This and a switch are likely to generate the same code.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: This kind of falls within the discussion about https://github.com/microsoft/MixedReality-WebRTC/pull/202#pullrequestreview-366325725. IRL it was proposed to add API calls for the wrapper code to add/get user data o native objects explicitly. This would bypass the need for this mechanism.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: > But what is a better alternative? Using non-reference-counted plain C handles?

Discussed IRL. Sounds like this is a better alternative.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Actually, changed to `mrsTransceiverWrapperInitConfig` because this one is for creating a wrapper. The one for actually creating a transceiver is now merged into `mrsTransceiverInitConfig`.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h

libs/Microsoft.MixedReality.WebRTC.Native/src/media/audio_transceiver.h
djee-ms: Fixed transceivers.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Moved to local track header.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -880,12 +1175,25 @@ void PeerConnectionImpl::OnDataChannelAdded(
...
 
+void PeerConnectionImpl::OnStreamChanged(
+    rtc::scoped_refptr<webrtc::MediaStreamInterface> stream) noexcept {
+  webrtc::AudioTrackVector audio_tracks = stream->GetAudioTracks();
+  webrtc::VideoTrackVector video_tracks = stream->GetVideoTracks();
+  RTC_LOG(LS_INFO) << "Media stream #" << stream->id()
+                   << " changed: " << audio_tracks.size()
+                   << " audio tracks and " << video_tracks.size()
+                   << " video tracks.";
+  // for (auto&& audio_track : remote_audio_tracks_) {
```
fibann: Delete, or add TODO if something has to be added here.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: Also found mention of Encode/DecodeIceServers.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1020,6 +1245,47 @@ bool PeerConnectionImpl::SetRemoteDescriptionAsync(
...
+          //                                                 mline_index, name))
+          //                                                 {
+          //    continue;
+          //  }
+          //  ErrorOr<RefPtr<Transceiver>> err = GetOrCreateTransceiverPlanB(
+          //      mline_index, receiver, std::move(name));
+          //  RTC_DCHECK(err.ok());
+          //  err.value()->OnSessionDescUpdated(/*remote=*/true);
+          //  ++mline_index;
+          //}
```
fibann: Add TODO or delete

djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1188,47 +1463,147 @@ void PeerConnectionImpl::OnAddTrack(
...
+    {
+      auto lock = std::scoped_lock{media_track_callback_mutex_};
+      auto cb = video_track_added_callback_;
+      if (cb) {
+        mrsTransceiverHandle tranceiver_handle = transceiver.release();
+        mrsRemoteVideoTrackHandle video_handle = remote_video_track.release();
+        cb(interop_handle, video_handle, transceiver_interop_handle,
+           tranceiver_handle);
+      }
     }
```
fibann: nit: 
image
This would be nice as two calls to a template function.

fibann: Same with block in OnRemoveTrack

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1274,63 +1670,620 @@ void PeerConnectionImpl::OnLocalDescCreated(
...
+                            return (
+                                (tr.get() == &transceiver) &&
+                                (tr->GetMediaKind() == mrsMediaKind::kAudio));
+                          }) != transceivers_.end());
+  RTC_DCHECK(std::find_if(local_audio_tracks_.begin(),
+                          local_audio_tracks_.end(),
+                          [&track](const RefPtr<LocalAudioTrack>& tr) {
+                            return (tr.get() == &track);
+                          }) == local_audio_tracks_.end());
+  local_audio_tracks_.push_back(&track);
```
fibann: nit: it seems that we need a lot of work to keep these tracks array up to date. Can we do without? Any reason why we can't get to all the tracks by looking into the transceivers? We only seem to be enumerating them in the Close() method anyway.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1274,63 +1670,620 @@ void PeerConnectionImpl::OnLocalDescCreated(
...
+    if (it_tr == transceivers.end()) {
+      return Error(Result::kInvalidOperation,
+                   "Cannot match RTP sender with RTP transceiver.");
+    }
+    rtc::scoped_refptr<webrtc::RtpTransceiverInterface> impl = *it_tr;
+    mline_index = (int)std::distance(transceivers.begin(), it_tr);
+
+    // Create the transceiver wrapper
+    mrsTransceiverInitConfig config{};
+    config.desired_direction = Transceiver::Direction::kSendRecv;  //?
```
fibann: //?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1274,63 +1670,620 @@ void PeerConnectionImpl::OnLocalDescCreated(
...
+      media_kind = mrsMediaKind::kVideo;
+      break;
+    default:
+      return Error(Result::kInvalidMediaKind, "Unknown SDP semantic");
+  }
+
+  // Find an existing transceiver wrapper which would have been created just
+  // a moment ago by the remote track added callback.
+  rtc::CritScope lock(&tracks_mutex_);
+  for (auto&& tr : transceivers_) {
```
fibann: Do you need to iterate if you have the mline_index?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1274,63 +1670,620 @@ void PeerConnectionImpl::OnLocalDescCreated(
...
+    if (tr && tr->HasReceiver(rtp_receiver)) {
+      RTC_DCHECK(media_kind == tr->GetMediaKind());
+      return static_cast<RefPtr<Transceiver>>(tr);
+    }
+  }
+  // Not found - create a new one
+  return CreateTransceiverPlanB(media_kind, mline_index, std::move(name),
+                                rtp_receiver);
+}
```
fibann: nit: these might be better as one single FindTransceiver method that returns null if not found + a call to CreateTransceiver(UnifiedPlan|PlanB) at the point of call.

libs/Microsoft.MixedReality.WebRTC.Native/test/audio_transceiver_tests.cpp
```diff
@@ -0,0 +1,241 @@
+// Copyright (c) Microsoft Corporation.
```
fibann: All the tests in this file are replicated word by word in video_transceiver_test.cpp. Should we write them just once, maybe parameterize them on the media kind and run twice? (though I suspect that running them once would be good enough).

libs/Microsoft.MixedReality.WebRTC.Native/test/external_video_track_source_tests.cpp
```diff
@@ -112,30 +116,134 @@ void ValidateQuadTestFrame(const void* data,
...
+    ASSERT_EQ(Result::kSuccess, mrsTransceiverGetLocalVideoTrack(
+                                    transceiver_handle1, &track_handle_local));
+    ASSERT_EQ(nullptr, track_handle_local);
+
+    // Remote track is NULL
+    mrsRemoteVideoTrackHandle track_handle_remote{};
+    ASSERT_EQ(Result::kSuccess, mrsTransceiverGetRemoteVideoTrack(
+                                    transceiver_handle1, &track_handle_remote));
+    ASSERT_EQ(nullptr, track_handle_remote);
+  }
```
fibann: nit: IMHO we can just have one test that checks if a new transceiver is consistent, and then we can trust that all new transceivers are consistent - we don't need to repeat this block in every test that deals with transceivers.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+void* SimpleInterop::CreateObject(ObjectType type) noexcept {
+  const uint32_t id = free_id_.fetch_add(1);
+  auto handle = std::make_unique<Handle>(Handle{this, type, id});
+  void* interop_handle = handle.get();
+  try {
+    auto lock = std::scoped_lock{objects_map_mutex_};
+    objects_map_.emplace(interop_handle, std::move(handle));
+  } catch (...) {
+  }
```
fibann: nit: why the try-catch?

djee-ms: Generally when working with the interop layer all functions are `noexcept` to prevent passing an exception through the interop layer. Here `std::unordered_map::emplace()` is not declared `noexcept` and might throw, so I'm catching that. Now that being said we should probably return `nullptr` as a failure in the `catch`, and `std::make_unique` can also throw so should be inside the `try { }` block

libs/Microsoft.MixedReality.WebRTC.Native/test/video_transceiver_tests.cpp
```diff
@@ -0,0 +1,634 @@
...
+  mrsTransceiverRemoveRef(transceiver_handle1);
+}
+
+TEST_F(VideoTransceiverTests, SetDirection_InvalidHandle) {
+  ASSERT_EQ(
+      Result::kInvalidNativeHandle,
+      mrsTransceiverSetDirection(nullptr, mrsTransceiverDirection::kRecvOnly));
+}
+
+TEST_P(VideoTransceiverTests, SetLocalTrackSendRecv) {
```
fibann: This and the following test are almost exactly the same, it would make sense to make the direction a parameter, or at least extract the common code into utility functions.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: Logged as #207, resolving for now.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
fibann: This should change as part of #207, resolving for now.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
fibann: Also - as said before - we shouldn't discard exceptions silently, at the very least we should print a generic error message.

djee-ms: Note that this is a test utility, only available inside the test executable, and we cannot log with `RTC_LOG` here (not exported). I will ensure the test fails with `ASSERT_TRUE(false)` though.

libs/Microsoft.MixedReality.WebRTC.Native/test/audio_transceiver_tests.cpp
djee-ms: Fixed. This makes something like this once parameterized by SDP semantic and by media type:

image

libs/Microsoft.MixedReality.WebRTC.Native/test/external_video_track_source_tests.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/test/video_transceiver_tests.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
djee-ms: I'm not sure what you mean. We need a type to expose to the API and we can't expose the WebRTC one.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
```diff
@@ -22,14 +22,17 @@ void SimpleInterop::Unregister(mrsPeerConnectionHandle pc) noexcept {
...
-  void* interop_handle = handle.get();
   try {
+    auto handle = std::make_unique<Handle>(Handle{this, type, id});
+    void* interop_handle = handle.get();
     auto lock = std::scoped_lock{objects_map_mutex_};
     objects_map_.emplace(interop_handle, std::move(handle));
+    return interop_handle;
   } catch (...) {
+    // Make sure the test fails
+    []() { ASSERT_TRUE(false); }();
```
fibann: nit: why the lambda? and why not at line 51?

djee-ms: The Google Test macros can only be used inside a function that returns `void` unfortunately. And I don't know of any other way to force-fail a test. So wrapping inside a lambda, and calling the lambda immediately. I can also write a `void ForceTestFailure()` helper, is that better for clarity?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: No you don't, will fix.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Fixed. Also moved some code inline when the method is only called once.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
fibann: Might be worth it for readability but it's not super-important.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/test/simple_interop.cpp
djee-ms: Fixed. That's much clearer.

libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
djee-ms: Discussed offline: removing the duplicate `static_assert` block in the other function.

djee-ms: Fixed. Also added a block for `OptDirection`.

libs/Microsoft.MixedReality.WebRTC.Native/src/interop/global_factory.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Fixed too.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Actually we don't need to keep a reference to local tracks, and this is even useless and partly wrong since those objects are owned by the user, so should only be strongly referenced by the transceiver they are attached to, if any. And remote tracks are owned by transceivers, so it's a similar story, and can be removed too.

Fixed. That deleted a bunch of code :wink:

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Actually ended up being unused, so deleted. Current code reads that from the user if manually added, or defaults to `kRecvOnly` when added for a newly receiving track, which is what `libwebrtc` does internally in Unified Plan.

Fixed.

drejx:
This is great stuff!  So if I understand correctly according to your initial statement:

> This change adds a new transceiver API to the interop library allowing the user to directly manipulate transceivers, which unlocks adding multiple audio and video tracks to a single peer connection. 

These changes add transceivers, which will allow work on multiple remote tracks to move forward, is that about right?  I think issue #152 is the work to actually have multiple remote tracks working.  

Would there be an estimate of when the multiple remote tracks feature would be in?

Thank you,
Andrej

djee-ms:
Yes it's for #152, and we need to review merge the C# library changes and Unity integration, so if everything goes right I'd say about a week to get the `merge/multi_track` branch in a ready state. Then we merge to `master`, which is more a matter of figuring out the right timing given the other major changes coming, especially Android support, to avoid being too much of a pain for those as this change the API quite a lot.

PR #205, opened 2020-02-28
DataChannel: SendMessage(IntPtr,ulong) / MessageReceivedUnsafe callback

simonchiarel:
avoids per-message byte[] array allocation, can be used in unsafe contexts.


simonchiarel:
Sorry, this didn't compile at first. It is fixed now.

PR #208, opened 2020-03-05
Link against core PDBs when building interop DLL

djee-ms:
This change enables linking the interop DLL `Microsoft.MixedReality.WebRTC.Native.dll` with the PDBs from the core `webrtc.lib` and, on UWP platform, the UWP wrappers from `Org.WebRtc.WrapperGlue.lib`, to enable a better debugging experience for users.

Given the large size of PDBs for each build variant (~400 MB and more), this changes uses some Universal Packages to store the PDBs in some private Artifacts feed on Azure DevOps when compiling `webrtc.lib`, and restore them when compiling `Microsoft.MixedReality.WebRTC.Native.dll`, as the frequency of the former is much lower than the frequency of the latter, and the former build is generally shared by multiple branches of the latter (_e.g._ `master` and `release/1.0` at the moment).

Additionally, because Azure build agents have a limit of 10GB per build, this change generates on-the-fly a custom `packages.config` for the current build triple (config-platform-arch) which allows restoring only the necessary core NuGet packages (containing `webrtc.lib`). This has the extra advantage of reducing the time needed to restore all platform packages: previously 4 on Desktop and 13 on UWP, now only 1 on Desktop and 3 on UWP. The `.vcxproj` is also modified on-the-fly and in-place to import only the restored packages. Note that those steps are executed via PowerShell scripts during the build and not stored in the committed `.vcxproj`/`packages.config` files because it is uncertain whether the Visual Studio IDE would understand platform- and config-dependent NuGet restore, which are rather unusual and only required because we are hitting the NuGet file size limit on nuget.org, therefore the files shipped on GitHub are left untouched with the full set of NuGet packages for all config/arch per platform, to avoid missing any NuGet restore which would prevent the user from compiling the solution.


djee-ms:
Note that this is merging to `release/1.0` with the intention of forward-porting to `master`. This is to save time on releasing next packages `1.0.3` with the QPmax fix and this change ASAP.

tools/ci/templates/jobs-cpp.yaml
```diff
@@ -26,14 +26,75 @@ jobs:
...
+    displayName: 'NuGet restore C++ lib'
+    inputs:
+      command: restore
+      restoreSolution: 'bin/${{parameters.buildPlatform}}/${{parameters.buildArch}}/${{parameters.buildConfig}}/packages.config'
+      restoreDirectory: '$(Build.SourcesDirectory)/packages'
+      includeNuGetOrg: true
+      feedsToUse: config
+      nugetConfigPath: '$(NuGetConfigPath)'
+    timeoutInMinutes: 30
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2  # NuGetCommand@2
```
fibann: nit: Add newline and comment, one has to re-read this a few times to see that it's a different task that executes alternatively to the previous one. 

Also the `condition` line is sometimes just below `task` and sometimes at the very end, this makes the sequence difficult to follow.

tools/ci/generateCppPackagesConfig.ps1
```diff
@@ -0,0 +1,57 @@
...
+
+$reader = [System.IO.File]::OpenText($InputFile)
+$content = ""
+try {
+    for() {
+        $line = $reader.ReadLine()
+        if ($line -eq $null) { break }
+        if ($line -match "Microsoft\.MixedReality\.WebRTC\.Native\.Core\.(Desktop|UWP|WinRT)\.")
+        {
+            if ($line -match "Microsoft\.MixedReality\.WebRTC\.Native\.Core\.(Desktop|UWP|WinRT)\.$BuildArch\.$BuildConfig")
```
fibann: Aren't these redundant?

fibann: Oh I see what it does. I'd rather do something like
```
if ($line -match "Microsoft\.MixedReality\.WebRTC\.Native\.Core\.(Desktop|UWP|WinRT)\.(.*)\.(.*)")
  if ($Matches.1 != ... OR $Matches.2 != $BuildArch OR $Matches.3 != $BuildConfig)
    // reject
// accept

File: tools/ci/templates/jobs-libwebrtc-win32.yaml [hunk 3]
```diff
@@ -44,19 +54,34 @@ jobs:
...
-  - task: PublishPipelineArtifact@0
+  # Publish PDBs for webrtc.lib as pipeline artifacts (limited retention)
+  # - task: PublishPipelineArtifact@0
+  #   displayName: 'Publish PDBs for webrtc.lib ($(buildTriple))'
+  #   inputs:
+  #     artifactName: 'libwebrtc_pdbs_$(buildTriple)'
+  #     targetPath: 'external/webrtc-uwp-sdk/webrtc/xplatform/webrtc/OUTPUT/webrtc/$(scriptPlatform)/${{parameters.buildArch}}/${{parameters.buildConfig}}/pdbs'
+  #   timeoutInMinutes: 15
+
+  # Publish PDBs for webrtc.lib as Universal Package (unlimited retention)
```
fibann: Why do we need to keep the pdbs indefinitely (especially given that the lib is only kept as a pipeline artifact)?

djee-ms: There are 2 distinct data streams here for `webrtc.lib` and its PDBs:
- If we are considering only CI, then we need the `webrtc.lib` PDBs for the C++ CI. The former runs infrequently, and pipeline artifacts have a limited retention period, so it's not uncommon for all artifacts on the core build to be deleted, which leaves the C++ CI without anything. Therefore we use Universal Packages instead which have unlimited retention (like a NuGet package, but without size limit).
- For `webrtc.lib` itself we are only publishing it here as a pipeline artifact because we need to move it from that pipeline to the signing one, which will generate the NuGet package we publish on nuget.org. This generally happens within a few hours or less. After that, the C++ CI will use the published NuGet packages, and the pipeline artifacts can be deleted.

fibann: Hmm I see. In general anyone who wants to build the MR.WebRTC.Native dlls will need/want the pdbs to be around, I suppose.

tools/ci/generateCppPackagesConfig.ps1
djee-ms: This is slightly more subtle than that, because for all UWP variants we need to restore the package `Microsoft.MixedReality.WebRTC.Native.Core.UWP` which contains the UWP generated headers (and no lib). With your proposed change, the line for this package will match, but `$Matches.2` will be `1` and `$Matches.3` will be `0`, because the full packages path is `Microsoft.MixedReality.WebRTC.Native.Core.UWP.1.0.3`. So the first `-match` here is to exclude that package, which is treated as any other ignored package and copied as is. Then for packages which contain `webrtc.lib` or `Org.WebRtc.dll` and are per-config/per-arch, we do the second match.

fibann: All of this would look good in a comment 😄

djee-ms: Fixed.

PR #211, opened 2020-03-06
Merge A/V transceivers creation + fix refcount bug

djee-ms:
This simplifies both the implementation and any wrapper layer above it.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -620,7 +620,7 @@ struct mrsTransceiverInitConfig {
   const char* name = nullptr;
 
   /// Kind of media the transceiver transports.
-  mrsMediaKind media_kind{};
+  mrsMediaKind media_kind{-1};  // defaults to invalid value to catch errors
 
   /// Initial desired direction of the transceiver media when created.
   mrsTransceiverDirection desired_direction =
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -614,15 +614,20 @@ enum class mrsMediaKind : uint32_t { kAudio = 0, kVideo = 1 };
 
 /// Configuration for creating a new transceiver.
 struct mrsTransceiverInitConfig {
-  /// Name of the transceiver. This must be a valid SDP token; see
-  /// |mrsSdpIsValidToken()|.
+  /// Optional name of the transceiver. This must be a valid SDP token; see
+  /// |mrsSdpIsValidToken()|. If no name is provided (empty or null string),
+  /// then the implementation will generate a random one.
   const char* name = nullptr;
+
+  /// Kind of media the transceiver transports.
+  mrsMediaKind media_kind{};
 
   /// Initial desired direction of the transceiver media when created.
   mrsTransceiverDirection desired_direction =
       mrsTransceiverDirection::kSendRecv;
 
-  /// Semi-colon separated list of stream IDs associated with the transceiver.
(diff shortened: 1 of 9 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -614,15 +614,20 @@ enum class mrsMediaKind : uint32_t { kAudio = 0, kVideo = 1 };
...
 struct mrsTransceiverInitConfig {
-  /// Name of the transceiver. This must be a valid SDP token; see
-  /// |mrsSdpIsValidToken()|.
+  /// Optional name of the transceiver. This must be a valid SDP token; see
+  /// |mrsSdpIsValidToken()|. If no name is provided (empty or null string),
+  /// then the implementation will generate a random one.
   const char* name = nullptr;
 
+  /// Kind of media the transceiver transports.
+  mrsMediaKind media_kind{};
```
stephenatwork: This defaults to audio (0)? We probably want this to be set always, so default to 'unset'?

djee-ms: There's no `unset`. I explicitly default to `{}` and not `{kAudio}`, even if they resolve to the same thing, to show we shouldn't expect a default. I prefer that than leaving it uninitialized. And adding a `kUnset` or `kUnknown` means going all over all other places where `mrsMediaKind` is used to add checks that the media kind is valid. Only alternative I can think of is initializing with an invalid value like `{(mrsMediaKind)-1}` to catch errors.

stephenatwork: My main point is that it seems like something the user should explicitly set. And as it stands, there's no way to check that. With an invalid default value, we could give better error messages at the point of use of the config.

djee-ms: This is a struct the wrapper layer passes to the interop DLL to create a transceiver. If it doesn't initialize it correctly, that's the wrapper layer's implementation fault. We cannot put checks on every single field. Here we ask the wrapper layer if it wants an audio or a video transceiver; this is to replace having 2 different functions. If we go down that route, `desired_direction` also has no more reason to be `kSendRecv` by default than `kInactive`. The user, as opposed to a wrapper implementer, would not typically fill that structure manually. The C# library for example allocates that struct on the stack and fills it with values from the user passed as arguments to `AddTransceiver()` right before the P/Invoke call. If we really want a check to prevent errors from wrapper implementers (and not users) then the invalid enum value `(mrsMediaKind)-1` sounds good enough to me to catch some uninitialized value, no? And that avoids having to check in all other places where we use that enum if the value is not `kUnset`, which the constructor of the transceiver already checked. At worse I'd be OK to add an `mrsOptMediaKind` if you really want to name that default argument, but that enum would only be used here.

djee-ms: More specifically, there are enums where "unset" is a valid value carrying some meaning, and which can be used at some point in the lifetime of the program. For example `mrsTransceiverOptDirection::kNotSet` is an actual value with a meaning, not just the user forgetting to assign a value. Same for `mrsVideoProfileKind::kUnspecified`, which means the user purposefully did not force a video profile kind. Here however we are talking about assigning a name to something that is an erroneous value that should in theory never happen. I looked at all other enums in the API, and either they do not have an "unset/unknown/unspecified" value, or they have one and it has a particular meaning like the ones above.

stephenatwork: OK, point taken that this is in the internal layer so devs generally don't have to concern themselves with it.
...unless they're C++ devs who need to use it until we have c++ wrappers :). 
I agree, it's a good idea to not name the invalid value so -1 seems like a good choice.

Issue #207, opened 2020-03-03
Refactor wrapper handles in interop API


fibann:
Each object in the interop API has two "handles" - a "handle" (simply), that identifies the native (C++) object, and an "interop handle", which, despite the name, usually identifies a wrapper for the native object in the upper layer (e.g. C# binding). The API functions generally use "simple handles", but the upper layer is also supposed to pass the wrapper handle to the API on object creation, and the interop library is supposed to pass it back as a callback argument when needed. The result is that there are a lot of handle types and handle parameters in the API, and both interop lib and wrapper layer have to deal with, and keep around, both. This is very confusing (because of the naming, among other things).

There are several ways this can be addressed, see e.g. https://github.com/microsoft/MixedReality-WebRTC/pull/202#discussion_r385640676. From IRL discussions, it might be worth trying a model where a) interop/wrapper handles are renamed to "user data", and b) user data for an object is not requested on every object creation, but explicitly for the minority (?) of objects that need it.

stephenatwork:
We discussed 3 options:

1. exposing the c++/c# handles as is currently in #202 
2. exposing only the c++ handle and having an associator of c++ & c# handle (e.g. map or weak ptr, etc)
3. exposing only the c++ handles and allowing duplicates of c# objects ("lightweight wrappers")

1&2 keep the 1:1 relationship between c# wrappers and implementation objects at the cost of extra complexity/management.
2&3 can project to more than one language within the same process. 1 appears to have a problem if there are more than one kind of interop handle?
3 allocates additional wrapper objects (increased gc pressure). And also can cause confusion about object identity; wrappers over the same underlying object may compare as different.

djee-ms:
The version proposed by @fibann is the one we discussed further yesterday, and is a variant of 1. where the "handle" terminology refers exclusively to an implementation object, and instead of manipulating explicitly some "wrapper handle" (_e.g._ reference to C# object) the implementation layer associates to the implementation objects some opaque "user data" with some API to set/retrieve those. This has several advantages:
- No more confusion about handles
- "User data" is a well-known concept, carries the meaning of opaque data the implementation DLL will not touch but only pass around where needed, and that the user can use for anything
- In the hypothetical case where a binding does not need wrappers (_e.g._ pure C app?) then it can simply skip setting any user data, so the current "wrapper handle" concept is not mandatory anymore.

PR #215, opened 2020-03-10
Replace interop handles with opaque user data

djee-ms:
Refactor the "interop handle" concept into a opaque "user data" storage on each implementation object, to disambiguate from native handles to those implementation objects.

This changes the creation steps of objects owned by the peer connection (transceivers, remote tracks, data channels), which are now created internally then notify the user of their existence via "added" callbacks. At this point any interop wrapper layer (like the C# library) can create wrapper objects associated with those native implementation ones, and optionally store any data in them via the "user data" storage, for example storing a raw pointer to the wrapper object.

This change also simplify as a result the lifecycle of those objects by making their handles pure C-style handles with a definite start/end lifetime, removing any reference count manipulation via handles and the risk to leak references.

Bug: #207

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
@@ -274,10 +274,34 @@ using mrsPeerConnectionVideoTrackRemovedCallback =
                     mrsRemoteVideoTrackHandle video_track,
                     mrsTransceiverHandle transceiver);
 
+enum class mrsDataChannelConfigFlags : uint32_t {
+  kNone = 0,
+  kOrdered = 0x1,
+  kReliable = 0x2,
+};
+
+inline mrsDataChannelConfigFlags operator|(
+    mrsDataChannelConfigFlags a,
+    mrsDataChannelConfigFlags b) noexcept {
+  return (mrsDataChannelConfigFlags)((uint32_t)a | (uint32_t)b);
+}
+
+inline uint32_t operator&(mrsDataChannelConfigFlags a,
+                          mrsDataChannelConfigFlags b) noexcept {
+  return ((uint32_t)a | (uint32_t)b);
+}
+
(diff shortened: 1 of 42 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
```diff
@@ -683,64 +609,39 @@ struct mrsRemoteVideoTrackConfig {
...
-/// In that case the channel is announced to the remote peer for it to create a
-/// channel with the same ID.
-/// - If id >= 0, then it adds a new out-of-band negotiated channel with the
-/// given ID, and it is the responsibility of the app to create a channel with
-/// the same ID on the remote peer to be able to use the channel.
+/// Add a new data channel to a peer connection.
+///
+/// The initial configuration of the data channel is provided by |config|, and
+/// is mandatory. The caller can optionally provide a set of callbacks to
+/// register with the data channel via the |callbacks| optional argument. The
```
fibann: There is no `callbacks`! Maybe this is meant to point to mrsDataChannelRegisterCallbacks?

djee-ms:
But that's the thing I think, this API makes it possible to have multiple layers in parallel, if instead of using the `user_data` storage they use some global hash map for example. Whereas the previous variant with an explicit wrapper handle was stuck with a single layer. No?

libs/Microsoft.MixedReality.WebRTC.Native/include/interop_api.h
djee-ms: Ah yes that comment is stale, callbacks have been removed from the creation method, and are not registered separately.

stephenatwork:
> this API makes it possible to have multiple layers in parallel,  ... some global hash map 

Yes, ok if we treat the userdata as an optimization for c#, we're good.

djee-ms:
More generally, as an optimization for any single-language wrapper, yes.

PR #221, opened 2020-03-12
Add multi-track support to C# library via transceiver API

djee-ms:
This is part 2/N of the multi-track support : C# library. See part 1 (interop; #202) for an overall description of the API changes.

Bug: #152


examples/TestAppUwp/MainPage.xaml.cs
```diff
@@ -551,6 +555,10 @@ private async void OnLoaded(object sender, RoutedEventArgs e)
         {
             LogMessage("Initializing the WebRTC native plugin...");
 
+            // Cannot run in UI thread on UWP because this will initialize the global factory
+            // (first library call) which needs to be done on a background thread.
```
fibann: nit: not for this review, but this looks like a very easy pitfall, it might be better to have an explicit initialization call instead.

djee-ms: Yes, although explicit call is quite tricky in Unity

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
```diff
@@ -0,0 +1,91 @@
...
+        //[UnmanagedFunctionPointer(CallingConvention.StdCall, CharSet = CharSet.Ansi)]
+        //public delegate IntPtr CreateObjectDelegate(IntPtr peer, in CreateConfig config);
+
+        //[MonoPInvokeCallback(typeof(CreateObjectDelegate))]
+        //public static IntPtr RemoteAudioTrackCreateObjectCallback(IntPtr peer, in CreateConfig config)
+        //{
+        //    var peerWrapper = Utils.ToWrapper<PeerConnection>(peer);
+        //    var remoteAudioTrackWrapper = CreateWrapper(peerWrapper, in config);
+        //    return Utils.MakeWrapperRef(remoteAudioTrackWrapper);
+        //}
```
fibann: Delete

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Microsoft Corporation.
```
fibann: Might seem a dejavu, but... can we have a single class for Transceiver? Methods could take a MediaTrack and then cast it whenever necessary.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
```diff
@@ -0,0 +1,146 @@
...
+            LocalTrack?.OnMute(muted);
+        }
+
+        /// <inheritdoc/>
+        protected override void OnRemoteTrackMuteChanged(bool muted)
+        {
+            RemoteTrack?.OnMute(muted);
+        }
+
+        internal void OnLocalTrackAdded(LocalAudioTrack track)
```
fibann: This should be called "SetLocalTrackInternal" or something similar, the `On` is conventionally used for callbacks and this is called normally by other methods.

In general all these methods called "On[Local|Remote]TrackAdded" in various classes sound mis-named, the name doesn't say what they actually do and it's impossible to figure out which calls which.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
```diff
@@ -0,0 +1,146 @@
...
+                Utils.ThrowOnErrorCode(res);
+            }
+            else
+            {
+                // Note: Cannot pass null for SafeHandle parameter value (ArgumentNullException)
+                var res = TransceiverInterop.Transceiver_SetLocalAudioTrack(_nativeHandle, new LocalAudioTrackHandle());
+                Utils.ThrowOnErrorCode(res);
+            }
+
+            // Capture peer connection; it gets reset during track manipulation below
```
fibann: It's not clear how this is possible, and anyway it sounds like an anti-pattern that A.x() calls B.y() that in turn changes some member in A - the dependencies are difficult to understand. This should be simplified - I'd try to suggest a way but I am not sure how this works yet 😄

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -280,11 +280,201 @@ public struct VideoCaptureFormat
...
+        public void EndSuspend()
+        {
+            Action cb = null;
+            lock (_lock)
+            {
+                Debug.Assert(_suspendCount > 0);
+                --_suspendCount;
+                if (_eventPending && (_suspendCount == 0))
+                {
+                    cb = Event;
```
fibann: `Event` is not set under lock. Unless we want to serialize changes to it together with calls `Begin`/`EndSuspend` - in which case we should set it under lock - there is no reason to read it under lock.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -280,11 +280,201 @@ public struct VideoCaptureFormat
...
+        }
+
+        /// <summary>
+        /// Try to invoke <see cref="Event"/> asynchronously.
+        /// This is equivalent to <code>Invoke(async: true)</code>.
+        /// </summary>
+        public void InvokeAsync() => Invoke(async: true);
+
+        /// <summary>
+        /// Lock for renegotiation event suspending variables:
```
fibann: nit: this mentions renegotiation but the class is generic.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -280,11 +280,201 @@ public struct VideoCaptureFormat
...
+        /// <summary>
+        /// List of stream IDs to associate the transceiver with.
+        /// </summary>
+        public List<string> StreamIDs;
+    }
+
+    /// <summary>
+    /// Wrapper for an event possibly delayed.
+    /// </summary>
+    public class DelayedEvent
```
fibann: nit: possibly move the class to a separate file. Also make internal unless you want to add it to the public API/have people take a dependency on it.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: There must be a global state change from good known state A to good known state B, which involves touching multiple objects. So to avoid A=>B=>A=>B=>... recursive changes (classical SetParent/AddChild issue), there needs to be a "hierarchy":
- Any call is redirected to tracks
- Tracks self-update, then update their transceivers
- Transceivers self-update, then update their peer connection
- The chain of update stops there.

Here the transceiver receives a user call so redirect to the track (_e.g._ `parent.AddChild(child)` => `child.SetParent(parent)`). Then the track self-update, update its transceiver (the caller), and finally the peer connection. So `Transceiver.PeerConnection` gets updated by `OnTrackRemoved()` and `OnTrackAdded()`.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: It's not about reading it under the lock, it's about reading vs. not reading it, which tells whether it should be called now or not. And this decision depends on variables under the lock. I could instead read it unconditionally and have a boolean `shouldCallNow`, is that more clear?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: Legacy comment from the time it was not in its own class, will fix.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: Moving to `internal` and leaving there, it's only used by `PeerConnection`.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Yes I hesitated but I am more inclined to keep the type safety here. Would a generic class be better?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
fibann: Yes, I don't think this gives us much safety - the only thing that it prevents is passing an audio track to a video transceiver, which seems quite easy to avoid anyway. I am much more worried about bugs caused by duplicate code getting out of sync.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
fibann: Why can't it be:
- transceiver receives call
- transceiver updates track
- transceiver updates peer connection
?

A chain of side effects is much more complicated to follow/debug than a flat list.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
fibann: Yes, or just return early if no call is necessary.

Same of `Invoke()` below.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Because the starting point is not always a user call to the transceiver. The transceiver might be created by the implementation. Also other changes, _e.g._ to tracks, might do the same operation than the one done here as a result of changing the transceiver, so those `OnXxx` calls might be called from multiple places.

I will try to have a pass to flatten/simplify, see what we can improve.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Fair enough, will try merging and see how it looks.

libs/Microsoft.MixedReality.WebRTC/Transceiver.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        /// <summary>
+        /// Change the media flowing direction of the transceiver.
+        /// This triggers a renegotiation needed event to synchronize with the remote peer.
+        /// </summary>
+        /// <param name="newDirection">The new flowing direction.</param>
+        /// <seealso cref="DesiredDirection"/>
+        /// <seealso cref="NegotiatedDirection"/>
+        public void SetDirection(Direction newDirection)
```
fibann: nit: IMHO a `SetDirection` and a settable `Direction` are redundant

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
fibann: > Because the starting point is not always a user call to the transceiver

I don't think it matters too much that it is a _user_ call - the point is, it's a transceiver operation so some code is calling a method in the transceiver.

> those OnXxx calls might be called from multiple places

These seem to be called only here though 😄 In general, if we do want to trigger a set of side effects across multiple objects from multiple places then a chain of calls might be an option, but it makes the code more complicated and less transparent, so I'd avoid it if possible.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Looks good enough. Fixed.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -951,7 +951,8 @@ void PeerConnectionImpl::RemoveAllDataChannels() noexcept {
 
     // Invoke the DataChannelRemoved callback on the wrapper if any
     if (removed_cb) {
-      mrsDataChannelHandle data_native_handle = (void*)&data_channel;
+      DataChannel* const dc = data_channel.operator->();
```
fibann: nit:
```suggestion
      DataChannel* const dc = data_channel.get();
```

libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
```diff
@@ -160,7 +160,12 @@ internal DataChannel(IntPtr nativeHandle, PeerConnection peer, IntPtr argsRef, i
         public void Dispose()
         {
             State = ChannelState.Closing;
+            // This will invoke some callback from native, which calls Destroy()
```
fibann: There seem to be a few dodgy things going on here:
- the same back and forth of control between classes that's going on in tracks/transceivers. It would be simpler if `PeerConnection.RemoveDataChannel` simply _removed_ the DataChannel from the PC and the call to `Destroy` were explicit (here and in the PC).
- the fact that a disposable class can be destructed under the hood without an explicit call to Dispose is a bit weird (though acceptable as long as it's documented)
- `Dispose()` should be safe to call twice - especially if it can be called under the hood without the user's request.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: meh, IntelliSense refused it, but apparently `std::shared_ptr` does have a `get()` 😢

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/Transceiver.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/AudioTransceiver.cs
fibann: Much nicer now IMHO 😉

libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
djee-ms: Fixed. `DataChannel` is not `IDisposable` anymore. `PeerConnection` owns a collection of them to keep the C# objects alive while a native implementation is alive too, between the add and remove calls.

libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
```diff
@@ -146,25 +160,9 @@ internal DataChannel(IntPtr nativeHandle, PeerConnection peer, IntPtr argsRef, i
...
         /// </summary>
-        public void Dispose()
-        {
-            State = ChannelState.Closing;
-            // This will invoke some callback from native, which calls Destroy()
-            PeerConnection.RemoveDataChannel(_nativeHandle);
-        }
-
-        internal void Destroy()
+        internal void DestroyNative()
```
fibann: The call to GC.SuppressFinalize is unnecessary now, right?

djee-ms: Fixed.

PR #228, opened 2020-03-19
Add multi-track support to Unity via transceiver API

djee-ms:
This is part 3/N of the multi-track support : Unity integration. See part 1 (interop; #202) for an overall description of the API changes.

**TODO** : Preferred codecs and custom codec parameters won't work with this change. This needs to be fixed first in the C# library. This will not be addressed in this PR.

Bug: #152

Unity-specific changes are highlighted below.

## Declarative transceiver pairing

The Unity integration adds an extra feature on top of the transceiver API : the ability to automatically pair transceivers in a declarative way.

- The user declares an array of transceivers inside the `PeerConnection` component, specifying which sender/receiver component to attach with each transceiver.
   image
   These "transceivers declarations" are named _media lines_ (`MediaLine` class) and do not add any transceiver to the peer connection yet.

- The implementation overloads `WebRTC.PeerConnection.CreateOffer()` (C#) with its own `Unity.PeerConnection.CreateOffer()` (Unity), and same for `SetRemoteDescriptionAsync()`. These two methods take care of adding the transceivers to the peer connection at the right time, and pairing them by their index into the transceiver array (_media line index_).

- Once the connection is established, the sender and receiver components automatically receive a callback notification with the transceiver they are attached to, allowing them to create a media track and set it on the transceiver.

## Local device capture components

The component for media capture from local devices have been renamed to use the explicit "webcam" and "microphone" terms. They are now both built on top of a generic `AudioSender` and `VideoSender`, which allow extensibility by the user to define new types, and allow future extension to custom audio tracks.

image

## New multi-track sample scene

A new multi-track sample scene `StandaloneDemo` has been added, which showcases the multi-track feature as a standalone demo using hard-coded signaling (direct function call) without the need to setup a `node-dss` signaler and without the use of any hardware device (webcam/microphone). Some helper component `FakeVideoSource` generates the low-resolution (16x16) uniform color video stream.

image


libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/FakeVideoSource.cs
```diff
@@ -0,0 +1,104 @@
+﻿using Microsoft.MixedReality.WebRTC;
+using Microsoft.MixedReality.WebRTC.Unity;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+/// <summary>
+/// A video source producing some colored frames generated programmatically.
+/// </summary>
+public class FakeVideoSource : CustomVideoSender<Argb32VideoFrameStorage>
```
stephenatwork: nit: something more descriptive than fake? e.g. UniformColor or ColorCycle etc

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
```diff
@@ -0,0 +1,115 @@
...
+
+/// <summary>
+/// Simple signaler using two peer connections in the same process,
+/// and hard-coding their SDP message delivery to avoid the need for
+/// any kind of networking to deliver SDP messages.
+/// 
+/// This component is designed to be used in demos where both peers
+/// are present in the same scene.
+/// </summary>
+public class HardcodedSignaler : MonoBehaviour
```
stephenatwork: nit: InProcessSignaler? LocalOnlySignaler?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
```diff
@@ -0,0 +1,525 @@
...
+                        }
+                        else
+                        {
+                            _preferredVideoCodec.stringValue = Enum.GetName(typeof(SdpVideoCodecs), newCodecValue);
+                        }
+                    }
+                }
+                catch (Exception)
+                {
+                    EditorGUILayout.PropertyField(_preferredVideoCodec);
```
fibann: Document what's supposed to trigger this and what's the logic here.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
```diff
@@ -0,0 +1,525 @@
...
+            {
+                throw new ArgumentOutOfRangeException("Default value cannot be invalid.");
+            }
+
+            using (new EditorGUILayout.HorizontalScope())
+            {
+                var rect = EditorGUILayout.GetControlRect();
+                using (new EditorGUI.PropertyScope(rect, label, intProperty))
+                {
+                    bool hasConstraint = (intProperty.intValue > 0);
```
fibann: All this constraint stuff is a bit obscure. The name `OptionalIntField` makes me think to something like `std::optional<int>` so I can't make much sense of why constraints and explicit comparisons to 0 are involved. Should be documented.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
```diff
@@ -0,0 +1,525 @@
...
+                    {
+                        // Force a valid value, otherwise the edit field won't show up
+                        if (string.IsNullOrWhiteSpace(newValue))
+                        {
+                            newValue = defaultValue;
+                        }
+
+                        // Make this delayed to allow override for negative numbers before it takes effect.
+                        // This allows the user to type "-5" and only on Return/Enter or focus loss see the
+                        // value change to 0.
```
fibann: This should be reworded.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    private void Start()
+    {
+        Peer1.OnInitialized.AddListener(OnInitialized1);
+        Peer2.OnInitialized.AddListener(OnInitialized2);
+
+        CreateOfferButton?.onClick.AddListener(() =>
+        {
+            Peer1.CreateOffer();
+        });
```
fibann: nit: this pattern - the component takes a reference to an existing button and adds an onClick listener on start - seems a bit unusual, I'd expect the button to be configured in the inspector to trigger a call to Connect in this component.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
djee-ms: The problem is you have to work with Unity serializing only the `int` value. So you cannot add a `bool` to indicate the optional. The way this works in the interop API is that zero/negative values are considered to be invalid, and mean that the field does not have a value. Here we reuse that same pattern. The word "constraint" here is a leftover; I can reword. But the general mechanism stays, I don't have any alternative for it.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
djee-ms: Something like this?
> // Make updating the value of the serialized property delayed to allow overriding the
> // value the user will input before it's assigned to the property, for validation.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
djee-ms: Removed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/FakeVideoSource.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
djee-ms: I tried but it doesn't seem like you can access a non-Unity object like `Peer1` to call a method on it from a Unity event assigned in the editor. This needs to be done from code.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime/VideoSourceTests.cs
```diff
@@ -0,0 +1,1110 @@
...
+            // Connect
+            Assert.IsTrue(sig.Connect(millisecondsTimeout: 60000));
+
+            // Wait a frame so that the Unity events for streams started can propagate
+            yield return null;
+
+            // Check pairing
+            Assert.NotNull(receiver1.Track);
+            Assert.NotNull(receiver2.Track);
+        }
```
fibann: Making these a parameterized function wouldn't be bad - there are only a few lines that are different between each.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
djee-ms: Actually, `Peer1` is a component here. Still, the `CreateOffer()` method doesn't show up in the list of thing you can call automatically on button click.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime/VideoSourceTests.cs
fibann: Also moving some of the blocks in a function/method would help since they seem to be shared by the other tests below.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
fibann: Still a few issues:
- `defaultValue` is not a great name since it's not really a default. `lastValidValue` maybe?
- `hasValidValue` is used twice with the meaning of "property is set to a valid value" first and "checkbox is set" aka "must get a new value with a text field" later. This is quite confusing, should be two variables. 
- there are a few instances of "constraint" left over

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
fibann: I just meant that this talks about integers but it's about a text property 😄 I guess it's a bit more precise now.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
fibann: I would expect any void method to be available for wiring? Can you simply wire `LocalOnlySignaler.Connect()`?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Editor/WebcamSourceEditor.cs
djee-ms: Good suggestions, fixed all.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Examples/StandaloneDemo/HardcodedSignaler.cs
djee-ms: Yes that one works, will fix.

djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioReceiver.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// once the <see cref="AudioStreamStarted"/> event was triggered.
+        /// </summary>
+        /// <param name="callback">The new frame callback to register.</param>
+        /// <remarks>
+        /// A typical application might use this callback to display some feedback of a local webcam recording.
+        /// 
+        /// Note that registering a callback does not influence the video capture and sending to the
+        /// remote peer, which occurs whether or not a callback is registered.
+        /// </remarks>
+        public void RegisterCallback(AudioFrameDelegate callback) { }
```
fibann: Add a TODO saying that this is not doing anything?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioSender.cs
```diff
@@ -0,0 +1,177 @@
...
+                Track.Enabled = mute;
+            }
+        }
+
+        /// <summary>
+        /// Implement this callback to create the <see cref="Track"/> instance.
+        /// On failure, this method must throw an exception. Otherwise it must set the <see cref="Track"/>
+        /// property to a non-<c>null</c> instance.
+        /// </summary>
+        protected abstract Task CreateLocalAudioTrackAsync();
```
fibann: Rename to `DoCreateLocalTrackAsync` or `CreateLocalTrackAsyncImpl`? So it's clear that it's supposed to be an implementation, the extra "Audio" doesn't say much.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -309,7 +309,23 @@ internal class DelayedEvent
...
+                    return _event;
+                }
+            }
+            set
+            {
+                lock (_lock)
+                {
+                    _event = value;
+                }
+            }
```
fibann: Why do we need the lock? `?.` shouldn't need it.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioReceiver.cs
djee-ms: Actually this was missing. The one on `AudioSender` does nothing though, so is now throwing a `NotImplementedException`. Also added a bunch of comments and warnings showing up prominently in the docs about all the audio limitations until we fix them.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioSender.cs
djee-ms: There's no `DoCreateLocalTrackAsync`, but do you think it needs the `Impl` anyway?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: The locks allows the behavior to be deterministic. I realized that if you change `Event` around the same time that `EndSuspend()` is running, then you might end up caching the action before it's changed, then invoke the old handler. Admittedly in our use case it's not happening, but that makes the class safe to reuse in another context, and shouldn't impact performance.

If I am not mistaken, the `?.` in `EndSuspend()` as well as the `(cb != null)` check in `Invoke()` are actually critical, and fixing an existing bug in some C# test. Because `Event` is a delegate it evaluates to `null` if there is no handler registered with it. And that happens if the user doesn't register any handler with `_renegotiationNeededEvent`, the event we use that `DelayedEvent` helper with.
https://codeblog.jonskeet.uk/2015/01/30/clean-event-handlers-invocation-with-c-6/

fibann: Yeah, just saying that using `?.` doesn't require a lock. Determinism is good reason enough I suppose.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
```diff
@@ -1,608 +1,1208 @@
...
+            if (Kind == MediaKind.Audio)
+            {
+                var audioSender = (AudioSender)Sender;
+                await audioSender.AttachTrackAsync();
+            }
+            else if (Kind == MediaKind.Video)
+            {
+                var videoSender = (VideoSender)Sender;
+                await videoSender.AttachTrackAsync();
+            }
```
fibann: AttachTrackAsync and DetachTrack could be virtual methods in MediaSender and we'd avoid dispatching explicitly here.

fibann: Same with OnPaired/OnUnpaired (could take a MediaTrack and cast to the specific track type internally).

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
```diff
@@ -1,608 +1,1208 @@
...
+        /// it will signal the remote peer that it does not wish to send (receive-only or inactive).
+        /// </summary>
+        public MediaSender Sender;
+
+        /// <summary>
+        /// Media receiver consuming the media received through the transceiver attached to this media line.
+        /// This must be an instance of a class derived from <see cref="AudioReceiver"/> or <see cref="VideoReceiver"/>
+        /// depending on <see cref="Kind"/>.
+        /// If this is non-<c>null</c> then the peer connection will negotiate receiving some media, otherwise
+        /// it will signal the remote peer that it does not wish to receive (send-only or inactive).
```
fibann: nit: it is a bit unclear what happens if this is changed after the negotiation has happened.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
```diff
@@ -1,608 +1,1208 @@
...
+        /// underlying peer connection is initialized, whereas any <see cref="OnInitialized"/>
+        /// event handler is invoked when control returns to the main Unity app thread. The former
+        /// is faster, but does not allow accessing the underlying peer connection because it
+        /// returns before <see cref="OnPostInitialize"/> executed. Therefore it is generally
+        /// recommended to listen to the <see cref="OnInitialized"/> event, and ignore the returned
+        /// <see xref="System.Threading.Tasks.Task"/> object.
+        /// 
+        /// If the peer connection is already initialized, this method returns immediately with
+        /// a <see xref="System.Threading.Tasks.Task.CompletedTask"/> object. The caller can check
+        /// that the <see cref="Peer"/> property is non-<c>null</c> to confirmat that the connection
```
fibann:
```suggestion
        /// that the <see cref="Peer"/> property is non-<c>null</c> to confirm that the connection
```

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: Good catch. Currently this requires a renegotiation, and the `RenegotiationNeeded` event is **not** raised. This is a bit unexpected I think. Maybe changing those values should immediately change the transceiver's desired direction for consistency, which would raise the event. However this only works after the first negotiation, because before it the transceiver is not yet available, so this introduces some inconsistency. Thoughts?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: But `MediaSender` is for local (sender) and remote (receiver) tracks, while `AttachTrackAsync()` is for senders only. I wanted to avoid mixing the 2 and have some empty implementation on the receivers, which sounds weird. Same for `OnPaired()`, this is only for receivers, not senders.

djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
```diff
@@ -1,608 +1,1208 @@
...
+                    // the remote description is applied (so for the offering side, at the end of the exchange when the
+                    // answer is received).
+                    bool wantsSend = (mediaLine.Sender != null);
+                    bool wantsRecv = (mediaLine.Receiver != null);
+                    var wantsDir = (wantsSend ? (wantsRecv ? Transceiver.Direction.SendReceive : Transceiver.Direction.SendOnly)
+                        : (wantsRecv ? Transceiver.Direction.ReceiveOnly : Transceiver.Direction.Inactive));
+
+                    // Ensure the media line has a transceiver, and update its desired direction
+                    Transceiver tr = null;
+                    if (mlineIndex < transceivers.Count)
```
fibann: nit: IMHO it would be nicer to split into two loops:
```
for (int mlineIndex = 0; mlineIndex < transceivers.Count; ++mlineIndex) { ... }
for (int mlineIndex = transceivers.Count; mlineIndex < _mediaLines.Count; ++mlineIndex) { ... }
```
(transceivers are always <= media lines, right?) 
This way the part that goes
```
tr = _nativePeer.AddTransceiver(mediaLine.Kind, settings);
...
Debug.Assert(tr.MlineIndex == mlineIndex);
```
becomes a bit clearer because it's evident that the new transceivers are all sequential.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioReceiver.cs
```diff
@@ -144,15 +144,18 @@ internal void AttachToTransceiver(Transceiver audioTransceiver)
         /// with this receiver, which enqueues the <see cref="AudioSource.AudioStreamStarted"/>
         /// event to be fired from the main Unity app thread.
         /// </summary>
-        internal void OnPaired(RemoteAudioTrack track)
+        internal override void OnPaired(MediaTrack track)
         {
+            Debug.Assert(track is RemoteAudioTrack);
+            var remoteAudioTrack = (RemoteAudioTrack)track;
```
fibann: Do we need the assert? Won't it throw an exception just like the cast?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
fibann: Discussed in chat, changing these members should change the direction on the corresponding transceiver. It's fine if it does nothing before the first negotiation, this (and the fact that the native objects should be used with care) will be documented.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioReceiver.cs
djee-ms: Alright my bad I still had the C/C++ cast model in mind, C# cast actually throws `InvalidCastException`. I will remove the assert.

djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioSender.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: Fixed.

Issue #188, opened 2020-02-13
Use enum for type in SDP methods/events


fibann:
It would be safer than arbitrary strings.

Issue #172, opened 2020-01-21
Unity version of abstract Signaler doesn't work without modifications to base class


pnolen:
For use of the `Signaler` class in the Microsoft.MixedReality.WebRTC.Unity namespace, it is expected that the class is abstract and should be extended. There are four events that are used by the `PeerConnection` class and others to handle rtc messaging from whatever signaling server is used, indicated with the comment "Those events must be invoked by the derived class". As follows:

`       public event Action OnConnect;
        public event Action OnDisconnect;
        public event Action<Message> OnMessage;
        public event Action<Exception> OnFailure;`

The problem is that events can only be invoked from the class they are declared in and can not be invoked in deriving classes, so in order to make this functionality work, modifications to the webrtc code have to be made for basic implementation (unless you manually handle the RTC events and basically step around the `PeerConnection` implementation). In fact, the example `NodeDssSignaler` does not use the events and instead has it's own IEnumerator method that processes messages and calls methods on the `WebRTC.PeerConnection` itself, which is a duplication of the code in the Unity `Signaler` class.

If the intent is that this Unity code is supposed to be usable "out of the box", then some method of invoking these events should be added (some protected methods probably) and the `NodeDssSignaler` should probably be updated to use the complete pattern in the `Signaler`.  If the intent is that this Unity code is an example, I feel that it should be marked as such.

djee-ms:
Hi @pnolen- I agree this is a bit of a mess at the moment, first because the signaler was changed mid-way to not use an interface anymore, and second because we have a single demo implementation and so are lacking a real-world implementation to validate it properly.

To answer your questions, `NodeDssSignaler` is demo-level code and should not be used in production, but the original intent was that `Signaler` be extended by the user, so working out of the box. We have not settled yet on what approach is the most suited, either via events like the API suggests, or via direct calls like `NodeDssSignaler` does in practice, and I am open to suggestion on which one is more suited.

PR #233, opened 2020-03-24
Remove Signaler use from PeerConnection

djee-ms:
- Remove any mention of Signaler in PeerConnection.
- Remove the node-dss specific message type from the Signaler abstract
  base class, and move it to the specific NodeDssSignaler
  implementation.

Note that currently Signaler encapsulates messages into a Message type,
which is then re-encapsulated into a different NodeDssMessage type by
the NodeDssSignaler. This looks redundant, but there are 2 reasons for
this:
- NodeDssSignaler needs a specific message type that serializes to JSON
  with specific field names, to be compatible with the node-dss
  protocol.
- The Signaler.Message type will eventually move into the C# library API
  and be transparent to the user (#188).

Bug: #172

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/Signaler.cs
+++ b/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/Signaler.cs
@@ -152,7 +152,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             _nativePeer = PeerConnection.Peer;
 
             // Register handlers for the SDP events
-            _nativePeer.IceCandidateReadytoSend += OnIceCandiateReadyToSend_Listener;
+            _nativePeer.IceCandidateReadytoSend += OnIceCandidateReadyToSend_Listener;
             _nativePeer.LocalSdpReadytoSend += OnLocalSdpReadyToSend_Listener;
         }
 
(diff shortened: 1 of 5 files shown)


libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Signaling/Signaler.cs
```diff
@@ -1,199 +1,255 @@
...
+        /// Callback fired from the <see cref="PeerConnection"/> when it finished
+        /// initializing, to subscribe to signaling-related events.
+        /// </summary>
+        /// <param name="peer">The peer connection to attach to</param>
+        public void OnPeerInitialized()
+        {
+            _nativePeer = PeerConnection.Peer;
+
+            // Register handlers for the SDP events
+            _nativePeer.IceCandidateReadytoSend += OnIceCandiateReadyToSend_Listener;
```
fibann: There are a few "IceCandiate" that should be "Ice Candi**d**ate".

djee-ms: meh 😞

djee-ms: Fixed... in so many places 🤣

PR #236, opened 2020-03-25
Update Unity tutorial for multi-track changes

djee-ms:
Update the Unity tutorial to reflect the changes introduced by multi-track support.
This change does not touch the C# documentation.


docs/manual/helloworld-unity-connection.md
```diff
@@ -1,10 +1,10 @@
 # Establishing a connection
 
-Now that we have a both local and remote video players, we can attempt to establish a connection with a remote peer.
+Now that we have a both media senders and receivers, we can attempt to establish a connection with a remote peer.
```
fibann:
```suggestion
Now that we have both media senders and receivers, we can attempt to establish a connection with a remote peer.

```

docs/manual/helloworld-unity-localvideo.md
```diff
@@ -1,38 +1,124 @@
...
+  - the component is enabled, that is [`MonoBehaviour.OnEnable()`](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnEnable.html) executed; and
+  - the peer connection is initialized, that is the [`PeerConnection.Initialized`](xref:Microsoft.MixedReality.WebRTC.PeerConnection.Initialized) event was raised.
+
+  This property is checked by default for developer convenience. If unchecked, then the application has to explicitly call [`MediaSender.StartCaptureAsync()`](xref:Microsoft.MixedReality.WebRTC.Unity.MediaSender.StartCaptureAsync) once the two conditions above are true. This can be useful if the application wants to enforce an explicit user action to start video capture, which is generally best practice anyway privacy-wise. Therefore developers are encouraged to uncheck it, and instead bind an end-user action (_e.g._ button click) to a call to [`MediaSender.StartCaptureAsync()`](xref:Microsoft.MixedReality.WebRTC.Unity.MediaSender.StartCaptureAsync).
```
fibann: So maybe it should be unchecked by default?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
```diff
@@ -645,8 +645,27 @@ public MediaLine AddTransceiver(MediaKind kind)
...
+        /// event to allow the user to send the message via the chosen signaling solution to the remote
+        /// peer.
+        /// 
+        /// <div class="IMPORTANT alert alert-important">
+        /// <h5>IMPORTANT</h5>
+        /// <p>
+        /// This method is very similar to the method available in the underlying C# library, and
+        /// actually calls it. However it also performs additional work in order to pair the transceivers
+        /// of the local and remote peer. Therefore Unity applications must call this overload instead
+        /// of the base C# library one to ensure transceiver pairing works as intended.
```
fibann: nit: all these disclaimers make me think that it might be worth considering calling these methods differently.

docs/manual/helloworld-unity-localvideo.md
djee-ms: This is the usual dilemma of the thing that works out-of-the-box but is not the best for production, vs. the best production setting that forces the user to do some setup out-of-the-box before they can make things work. I prefer the latter, and document how to disable it.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: That will not change things much IMHO. Users looking at the Unity scripts will do the right things, while users digging into the C# library API are at risk of breaking the Unity integration. Ideally we'd hide the C# peer, as already discussed, but that prevents access to features not explicitly exposed in Unity.

docs/manual/helloworld-unity-connection.md
djee-ms: Fixed.

docs/manual/helloworld-unity-localvideo.md
fibann: If this is meant to be used in production then IMHO the default setting should be what's best for production, especially if privacy is concerned. This way we are removing setup cost from prototypes and demos and adding it to real apps (one hast to virtually always uncheck the box) which doesn't seem that clever to me (also, in this case the "setup" is just ticking a box). 

I don't have strong views about this anyway, if you want to keep it like this is fine, just rephrase the docs to something like "this makes the component work out of the box for developer convenience. In cases where privacy is a concern, you might want to keep this unchecked and use an explicit call to .... instead". At the moment the docs read as "we decided to make this checked it by default, but it's better if you uncheck it" which sounds a bit silly 😄

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
fibann: Yes, I agree there is no way to make everything 100% safe, just saying that by using the exact same names we might be making things worse than necessary without a strong reason.

djee-ms: What name are you suggesting? How about:
- `StartConnectionAsync()` == `CreateOffer[Async]()`
- `ApplyRemoteOfferAsync()`
- `ApplyRemoteAnswerAsync()`

...with the last 2 ones calling `SetRemoteDescriptionAsync()` from the C# library?

djee-ms: Or even `ConnectAsync()` to make it even more simple?

fibann: "Connect" sounds good. Splitting offer and answer handling might bring to duplicating some code paths so I'd rather leave them one method. Maybe `HandleConnectionMessage`?

docs/manual/helloworld-unity-localvideo.md
djee-ms: Alright, I will rephrase.

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs
djee-ms: Fixed. As discussed offline, kept `StartConnection()` until we refactor `CreateOffer()` to be a proper C# async method.

PR #245, opened 2020-03-30
Add warnings to docs where due

djee-ms:
Add temporary warnings where due until some serious issues are fixed, to prevent users from getting stuck on them.


djee-ms:
> It would be nice if we could tag pages as 'last verified against vX.Y.Z' or some other way to automatically flag out of date docs.

Normally the `master` docs is for the `master` branch, etc. and are supposed to be up-to-date with the latest changes on that branch. We are exceptionally "relaxing" the rule here to move forward with multi-track due to timeline constraints, but normally this shouldn't happen.

PR #251, opened 2020-04-01
More fixes for M.MR.WebRTC.Native.dll rename

djee-ms:
Rename occurrences of `Microsoft.MixedReality.WebRTC.Native.dll` into `mrwebrtc.dll` in the documentation and the Unity integration sample project.


docs/manual/download.md
```diff
@@ -23,7 +23,7 @@ The C++ library is distributed as two separate packages for Windows Desktop and
 - [Windows Desktop package `Microsoft.MixedReality.WebRTC.Native.Desktop`](https://www.nuget.org/packages/Microsoft.MixedReality.WebRTC.Native.Desktop)
 - [Windows UWP package `Microsoft.MixedReality.WebRTC.Native.UWP`](https://www.nuget.org/packages/Microsoft.MixedReality.WebRTC.Native.UWP)
 
-The C++ packages contain the shared library `Microsoft.MixedReality.WebRTC.Native.dll` as well as its import library (`.lib`).
+The C++ NuGet packages contain the shared library `Microsoft.MixedReality.WebRTC.Native.dll` as well as its import library (`.lib`).
```
fibann, line 26: Missing rename here?

fibann: Ooh, is the thing that all NuGet pkgs still use the old name?

I'd rather write this kind of comments from the point of view of 2.0, using `mrwebrtc.lib`, and add a note block like "In versions before 2.0, mrwebrtc.dll is named Microsoft.MixedReality.WebRTC.Native.dll. The following steps are necessary...". So a) we don't need to come back here and change this when we generate new NuGet packages, b) it's cleared what's the norm and what's outdated, and c) we can remove those blocks more easily if they become obsolete.

docs/manual/helloworld-unity-importwebrtc.md
```diff
@@ -31,27 +31,36 @@ cd /D D:\testproj
...
 
-![Configure the import settings for a native C++ DLL](helloworld-unity-2.png)
+> [!IMPORTANT]
+> The `mrwebrtc.dll` module was previously named `Microsoft.MixedReality.WebRTC.Native.dll` when this module was also the C++ library. Those two have now been split apart (see issue [#123](https://github.com/microsoft/MixedReality-WebRTC/issues/123), and commit [169502a](https://github.com/microsoft/MixedReality-WebRTC/commit/169502aab851b61c260831892e3eff9deeb876ae) for rename). The `Microsoft.MixedReality.WebRTC.Native` name is still in use in some NuGet packages, so depending on the installation method one or the other name are used. However the Unity project must use the `mrwebrtc.dll` name only for the C# library to find it.
```
fibann, line 46: nit: I'd leave out commit and issue, not really relevant within the tutorial docs.

docs/manual/download.md
djee-ms: No, there's currently no NuGet package containing an `mrwebrtc.dll` file.

docs/manual/helloworld-unity-importwebrtc.md
djee-ms: I think it is in the sense that NuGet packages and local builds will have different file names.

docs/manual/download.md
djee-ms: Ok, makes sense. But indeed currently NuGet packages all use the old name.

docs/manual/helloworld-unity-importwebrtc.md
fibann: Yeah it's fine to point that out, but the exact commit isn't that useful, most people won't care about rolling back to a precise commit.

docs/manual/download.md
djee-ms: I actually expanded and clarified the differences between `v1.0` and `v2.0` for the C++ library. How does that look?

docs/manual/helloworld-unity-importwebrtc.md
djee-ms: Removed.

docs/manual/download.md
```diff
@@ -14,16 +16,24 @@ The C# library is distributed as two separate packages for Windows Desktop and W
...
+The C# library packages contain the C# assembly `Microsoft.MixedReality.WebRTC` as well as the per-architecture native DLLs. Therefore those packages are standalone, and there is no need to reference any other package in your project.
 
 ## C++ library
 
-The C++ library is distributed as two separate packages for Windows Desktop and Windows UWP:
+The C++ library architecture and distribution changed between the 1.0 and 2.0 releases.
+
+### New standalone library (v2.0)
+
+Starting from v2.0, the C++ library is a standalone source-only library building on top of the native implementation module `mrwebrtc.dll`. The legacy `Microsoft.MixedReality.WebRTC.Native.dll` previously containing the native implementation is gone. There is currently no precompiled binaries of that `mrwebrtc.dll` implementation module. The C++ library itself is only distributed as sources for C++ projects to integrate it and statically link with it.
```
fibann, line 27: Consumers of 2.0 shouldn't need to care about a `Microsoft.MixedReality.WebRTC.Native.dll` that previously contained a native implementation and now is gone.

Also 
```
There is currently no precompiled binaries
```
True, but see other comment - this phrasing means that as soon as we have new packages we need to come back here (and to all the "transitional" doc blocks) and change this. I'd say we don't need to specify, for now if the docs say that there is a mrwebrtc.dll in 2.0 packages they are not lying! And by the time we make a release with this docs there are going to be packages by definition 😄

docs/manual/helloworld-unity-importwebrtc.md
fibann: Good, I'd change "previously" with "in 1.0" which seems a good tradeoff between the exact commit and not saying at all.

Also see reference at line 61.

docs/manual/download.md
fibann: Much easier to follow now.

docs/manual/helloworld-unity-importwebrtc.md
djee-ms: Fixed.

docs/manual/download.md
djee-ms: Fixed.

fibann: I was only thinking of removing those two sentences, but yeah after those are gone there's not a lot of meat left in these paragraphs. Ship it!

PR #253, opened 2020-04-02
Change projects to link library produced by WinRTC patches

fibann:
Projects are changed to link the webrtc library produced by the patches at
https://github.com/microsoft/winrtc/tree/user/LoadLibrary/Arm64/patches_for_WebRTC_org/m80

Duplicate/obsolete sections in the vcxprojs have been tidied up.

Note that while the Win32 version of Microsoft.MixedReality.WebRTC.Native builds
using C++14, the UWP version uses C++17. This is because webrtc.lib is currently
build with C++17 for UWP but not for other configurations.


libs/Microsoft.MixedReality.WebRTC.Native/src/uwp/Microsoft.MixedReality.WebRTC.Native.UWP.vcxproj
```diff
@@ -103,79 +100,61 @@
...
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>_DEBUG;_LIBCPP_DEBUG=0;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_WINDLL;MR_SHARING_WIN;MRS_USE_STR_WRAPPER;WINUWP;_SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING;_ITERATOR_DEBUG_LEVEL=0;USE_AURA=1;_HAS_EXCEPTIONS=0;__STD_C;_CRT_RAND_S;_CRT_SECURE_NO_DEPRECATE;_SCL_SECURE_NO_DEPRECATE;_ATL_NO_OPENGL;_WINDOWS;CERT_CHAIN_PARA_HAS_EXTRA_FIELDS;PSAPI_VERSION=2;WIN32;_SECURE_ATL;_USING_V110_SDK71_;WINUWP;__WRL_NO_DEFAULT_LIB__;WINAPI_FAMILY=WINAPI_FAMILY_PC_APP;WIN10=_WIN32_WINNT_WIN10;WIN32_LEAN_AND_MEAN;NOMINMAX;_UNICODE;UNICODE;NTDDI_VERSION=NTDDI_WIN10_RS2;_WIN32_WINNT=0x0A00;WINVER=0x0A00;NVALGRIND;DYNAMIC_ANNOTATIONS_ENABLED=0;WEBRTC_ENABLE_PROTOBUF=1;WEBRTC_INCLUDE_INTERNAL_AUDIO_DEVICE;RTC_ENABLE_VP9;HAVE_SCTP;WEBRTC_LIBRARY_IMPL;WEBRTC_NON_STATIC_TRACE_EVENT_HANDLERS=0;WEBRTC_WIN;ABSL_ALLOCATOR_NOTHROW=1;GOOGLE_PROTOBUF_NO_RTTI;GOOGLE_PROTOBUF_NO_STATIC_INITIALIZER;_DEBUG%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
djee-ms, line 123: Why do we need all of that? What are the implications of all those changes? Are there any changes to included products that require re-evaluating licenses? This looks like a very large change...
- What's about Valgrind? We're using `cl.exe`, there's no Valgrind here.
- VP9 was already enabled, why do we need a new flag? Same for SCTP, this is the default and can't be changed, why is there a flag at all?
- What's with ProtoBuf? We don't use it AFAIK.
- `WINAPI_FAMILY=WINAPI_FAMILY_PC_APP` forces non-store apps. AFAIK this means people who want to use this project in an app and deploy it to the Windows store can't anymore.
- _etc._

fibann: I just copied the list from https://github.com/microsoft/winrtc/blob/user/LoadLibrary/Arm64/patches_for_WebRTC_org/m80/readme.md. I don't know about all of those, my guess is that it's simply the full list of flags used to compile webrtc.lib and so we reuse it as is to prevent inconsistencies when including WebRTC headers. Possibly some of them never appear in headers and aren't necessary.

Note that it's `NVALGRIND` which I assume means "no Valgrind".

WINAPI_FAMILY does look suspicious - @LoadLibrary do you know where it comes from? Is it required?

fibann: It seems fine actually, from config/win/BUILD.gn:
```
# possible values:
  #   "app" - Windows Store Applications
  #   "phone" - Windows Phone Applications
  #   "system" - Windows Drivers and Tools
  #   "server" - Windows Server Applications
  #   "desktop" - Windows Desktop Applications
  target_winuwp_family = "app"
...
if (target_winuwp_family == "app") {
      defines += [ "WINAPI_FAMILY=WINAPI_FAMILY_PC_APP" ]
    } else if (target_winuwp_family == "phone") {
      defines += [ "WINAPI_FAMILY=WINAPI_FAMILY_PHONE_APP" ]
    } else if (target_winuwp_family == "system") {
      defines += [ "WINAPI_FAMILY=WINAPI_FAMILY_SYSTEM" ]
    } else if (target_winuwp_family == "server") {
      defines += [ "WINAPI_FAMILY=WINAPI_FAMILY_SERVER" ]
    } else {
      defines += [ "WINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP" ]
    }
```
So it's the default that should be Store app.

djee-ms: Actually the naming is confusing:
```
 * WINAPI_PARTITION_DESKTOP            // usable for Desktop Win32 apps (but not store apps)
 * WINAPI_PARTITION_APP                // usable for Windows Universal store apps
 * WINAPI_PARTITION_PC_APP             // specific to Desktop-only store apps
```

Whatever "Desktop-only store apps" means...

fibann: Yes, the `_PARTITION_` macros are a combination of the `_FAMILY_` ones and don't match 1:1
image

fibann: WebRTC has a convenient `rtc_enable_protobuf` and `use_aura` flags. Building without them seems to work fine.

djee-ms: Let's review later for legals etc. and merge for now.

PR #255, opened 2020-04-06
Fix duplicate media line index by exposing real mid

djee-ms:
Fix duplicate media line index occurring when both peers create a transceiver before an offer by exposing the real `mid` attribute of the JSEP session as it appears in the SDP message. This allows better insight into the actual state of the transceivers.

Practically this changes a few things:
- The `Transceiver.MlineIndex` property may be invalid (negative) when the transceiver is not yet associated with a media line. Once an offer associates it, a new `Transceiver.Associated` event is raised and the property takes the actual value of the `mid` SDP attribute.
- Corollary of the above, the media line index of the transceivers of the `PeerConnection.Transceivers` array do not form anymore a contiguous sequence of indices [0..N], since data channels also consume a media line, and non-associated transceivers have a negative media line index.
- Corollary of the above, the `Transceiver.MlineIndex` property is not anymore the index into the `PeerConnection.Transceivers` array.

This change also fix an issue where "holes" appeared in the transceivers numbering due to data channels also consuming media lines yet not having any transceiver.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
@@ -459,6 +459,18 @@ void Transceiver::SetTrackPlanB(webrtc::MediaStreamTrackInterface* new_track) {
   }
 }
 
+void Transceiver::OnAssociated(int mline_index) {
+  assert(mline_index >= 0);
+  // Transceiver recycling is not used, to transceivers can only be associated
+  // once, from their initial non-associated state.
+  assert(mline_index_ < 0);
+  mline_index_ = mline_index;
+  auto lock = std::scoped_lock{cb_mutex_};
+  if (auto cb = associated_callback_) {
+    cb(mline_index);
+  }
+}
+
 void Transceiver::OnSessionDescUpdated(bool remote, bool forced) {
   // Parse state to check for changes
   bool changed = false;
--- a/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
+++ b/libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
@@ -17,7 +17,6 @@
 #include "sdp_utils.h"
 #include "utils.h"
 #include "video_frame_observer.h"
-#include "sdp_utils.h"
 
 #include <functional>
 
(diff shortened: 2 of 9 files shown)


libs/Microsoft.MixedReality.WebRTC.Native/src/media/transceiver.cpp
```diff
@@ -1,515 +1,527 @@
...
-    const std::string& encoded_string,
-    int& mline_index_out,
-    std::string& name,
-    std::vector<std::string>& stream_ids_out) {
-  if (encoded_string.empty()) {
-    mline_index_out = -1;
-    stream_ids_out.clear();
-    return false;
-  }
+// Copyright (c) Microsoft Corporation.
```
djee-ms, line 1: This is the actual diff (thanks GitHub...)

```diff
+void Transceiver::OnAssociated(int mline_index) {
+  assert(mline_index >= 0);
+  // Transceiver recycling is not used, to transceivers can only be associated
+  // once, from their initial non-associated state.
+  assert(mline_index_ < 0);
+  mline_index_ = mline_index;
+  auto lock = std::scoped_lock{cb_mutex_};
+  if (auto cb = associated_callback_) {
+    cb(mline_index);
+  }
+}
```

djee-ms: For future reference, I get that with this command:
```shell
git diff --ignore-space-at-eol HEAD~2 HEAD -- transceiver.cpp
```

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1678,39 +1675,31 @@ PeerConnectionImpl::GetOrCreateTransceiverForNewRemoteTrack(
...
-    {
-      rtc::CritScope lock(&transceivers_mutex_);
-      if (mline_index < transceivers_.size()) {
-        transceiver = transceivers_[mline_index];
-      }
-      RTC_DCHECK(!transceiver || (transceiver->impl() == tr));
-    }
+  for (auto&& rtp_tr : peer_->GetTransceivers()) {
+    const int mline_index = ExtractMlineIndexFromRtpTransceiver(rtp_tr);
+    RefPtr<Transceiver> transceiver = FindWrapperFromRtpTransceiver(rtp_tr);
```
fibann, line 1680: This is a quadratic loop. The element of `peer_->GetTransceivers()` and `transceivers_` should match by index, right? If not it should be easy enough to match them linearly anyway. I don't expect this to be a problem in practice but better safe than sorry 😄

djee-ms: Fixed with sorting by memory address of RTP transceiver. The index probably match, but this was an assumption about an implementation detail that I am glad to depart from. This should have no impact on performance in any case, as we expect no more than a few transceivers.

PR #259, opened 2020-04-06
Fix license of TestReceiveAV/Program.cs

djee-ms:
Restore the original author's license header and remove the Microsoft
license header to reflect the fact that this file was contributed by a
third-party author and not written by a Microsoft employee, in
accordance with the official Microsoft copyright headers policy for
open-source projects.

### DIFF
--- a/examples/TestReceiveAV/Program.cs
+++ b/examples/TestReceiveAV/Program.cs
@@ -1,5 +1,19 @@
-// Copyright (c) Microsoft Corporation.
+//-----------------------------------------------------------------------------
+// Filename: Program.cs
+//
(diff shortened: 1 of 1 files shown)


djee-ms:
Ping @sipsorcery for visibility.

I made a mistake on the original PR, the header should not have been changed to the Microsoft header since you were the sole author of it. Apology for that. Fixing as per Microsoft's official policy which states that:
> "Microsoft should not add a Microsoft copyright header to third party contributions to Microsoft OSS projects."

I also fixed the LF line endings, hence the larger diff.

This doesn't change anything about the `CONTRIBUTORS` mention.

PR #268, opened 2020-04-14
Merge master 843c7db (Android build)

fibann:


fibann:
Note: WebRTC seem to become deadlocked when stopping play in Unity editor, which prevents mrwebrtc.dll from being unloaded. Will investigate later.

PR #269, opened 2020-04-14
Add support for multi-track to TestAppUWP

djee-ms:
Add support for multiple audio and video tracks to the TestAppUWP
sample. This change refactors the app to use MVVM in an attempt to
clarify usage of the C# API, and to separate the tracks from the
transceivers in the UI.

### DIFF
--- a/examples/TestAppUwp/App.xaml.cs
+++ b/examples/TestAppUwp/App.xaml.cs
@@ -26,7 +26,7 @@ namespace TestAppUwp
         }
 
         /// <summary>
-        /// Invoked when the application is launched normally by the end user.  Other entry points
+        /// Invoked when the application is launched normally by the end user. Other entry points
         /// will be used such as when the application is launched to open a specific file.
         /// </summary>
         /// <param name="e">Details about the launch request and process.</param>
@@ -85,9 +85,9 @@ namespace TestAppUwp
         /// <param name="e">Details about the suspend request.</param>
         private void OnSuspending(object sender, SuspendingEventArgs e)
         {
-            var deferral = e.SuspendingOperation.GetDeferral();
-            //TODO: Save application state and stop any background activity
-            deferral.Complete();
+            //var deferral = e.SuspendingOperation.GetDeferral();
+            ////TODO: Save application state and stop any background activity
+            //deferral.Complete();
         }
     }
 }
--- a/examples/TestAppUwp/DebugConsolePage.xaml.cs
+++ b/examples/TestAppUwp/DebugConsolePage.xaml.cs
@@ -33,7 +33,9 @@ namespace TestAppUwp
     }
 
     /// <summary>
-    /// An empty page that can be used on its own or navigated to within a Frame.
+    /// Page displaying the so-called "debug console", a raw text dump of all messages
+    /// output through the <see cref="Logger"/> singleton instance by the various part
+    /// of the application, and mainly used for debugging.
     /// </summary>
     public sealed partial class DebugConsolePage : Page
     {
--- a/examples/TestAppUwp/DebugConsolePage.xaml.cs
+++ b/examples/TestAppUwp/DebugConsolePage.xaml.cs
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics;
+using Windows.UI.Xaml.Controls;
+
+namespace TestAppUwp
+{
+    public class Logger : NotifierBase
+    {
+        public static void Log(string message)
+        {
+            Instance.LogMessage(message);
+        }
+
+        private string _fullText;
+
+        public static Logger Instance { get; } = new Logger();
+
+        public string FullText
+        {
+            get { return _fullText; }
+            set { SetProperty(ref _fullText, value); }
+        }
+
+        public void LogMessage(string message)
+        {
+            Debugger.Log(4, "TestAppUWP", message);
+            message += "\n";
+            _fullText += message;
+            RaisePropertyChanged("FullText");
+        }
+    }
+
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
+    /// </summary>
+    public sealed partial class DebugConsolePage : Page
+    {
+        public DebugConsolePage()
+        {
+            this.InitializeComponent();
+        }
+    }
+}
--- a/examples/TestAppUwp/MainPage.xaml.cs
+++ b/examples/TestAppUwp/MainPage.xaml.cs
@@ -2,258 +2,71 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Diagnostics;
-using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.MixedReality.WebRTC;
-using TestAppUwp.Video;
 using Windows.ApplicationModel;
-using Windows.Media.Capture;
-using Windows.Media.Core;
-using Windows.Media.MediaProperties;
-using Windows.Media.Playback;
 using Windows.Storage;
-using Windows.System.Profile;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Markup;
 
 namespace TestAppUwp
 {
-    public class NavLink
+    public class CategoryBase { }
+
+    public class Category : CategoryBase
     {
-        public string Id { get; set; }
-        public string Label { get; set; }
-        public Symbol Symbol { get; set; }
+        public string Name { get; set; }
+        public string Tooltip { get; set; }
+        public Symbol Glyph { get; set; }
+        public Type PageType { get; set; }
     }
 
-    public class VideoCaptureDeviceInfo
-    {
-        public string Id;
-        public string DisplayName;
-        public Symbol Symbol;
+    public class Separator : CategoryBase { }
 
-        public bool SupportsVideoProfiles
+    public class Header : CategoryBase
+    {
+        public string Name { get; set; }
+    }
+
+    [ContentProperty(Name = "ItemTemplate")]
+    class MenuItemTemplateSelector : DataTemplateSelector
+    {
+        public DataTemplate ItemTemplate { get; set; }
+
+        internal DataTemplate HeaderTemplate = (DataTemplate)XamlReader.Load(
+            @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'>
+                   <NavigationViewItemHeader Content='{Binding Name}' />
+                  </DataTemplate>");
+
+        internal DataTemplate SeparatorTemplate = (DataTemplate)XamlReader.Load(
+            @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'>
+                    <NavigationViewItemSeparator />
+                  </DataTemplate>");
+
+        protected override DataTemplate SelectTemplateCore(object item)
         {
-            get
-            {
-                if (Id != null)
-                {
-                    return MediaCapture.IsVideoProfileSupported(Id);
-                }
-                return false;
-            }
+            return item is Separator ? SeparatorTemplate : item is Header ? HeaderTemplate : ItemTemplate;
         }
     }
 
-    public class ChatChannel
-    {
-        public DataChannel DataChannel;
-        public string Text = "";
-        public string Label { get { return DataChannel?.Label; } }
-    }
-
     /// <summary>
-    /// The main application page.
+    /// Main application page with the navigation menu.
     /// </summary>
     public sealed partial class MainPage : Page
     {
         /// <summary>
-        /// Indicates whether the native plugin has been successfully initialized,
-        /// and the <see cref="_peerConnection"/> object can be used.
+        /// Collection of categories for the main navigation menu.
         /// </summary>
-        public bool PluginInitialized { get; private set; } = false;
-
-        private Transceiver _audioTransceiver = null;
-        private Transceiver _videoTransceiver = null;
-
-        private MediaStreamSource localVideoSource = null;
-        private MediaSource localMediaSource = null;
-        private MediaPlayer localVideoPlayer = new MediaPlayer();
-        private bool _isLocalVideoPlaying = false;
-        private object _isLocalMediaPlayingLock = new object();
-        private LocalAudioTrack _localAudioTrack = null;
-        private LocalVideoTrack _localVideoTrack = null;
-
-        private MediaStreamSource remoteVideoSource = null;
-        private MediaSource remoteMediaSource = null;
-        private MediaPlayer remoteVideoPlayer = new MediaPlayer();
-        private bool _isRemoteVideoPlaying = false;
-        private bool _isRemoteAudioPlaying = false;
-        private uint _remoteVideoWidth = 0;
-        private uint _remoteVideoHeight = 0;
-        private uint _remoteAudioChannelCount = 0;
-        private uint _remoteAudioSampleRate = 0;
-        private object _isRemoteMediaPlayingLock = new object();
-        private RemoteAudioTrack _remoteAudioTrack = null;
-        private RemoteVideoTrack _remoteVideoTrack = null;
-
-        /// <summary>
-        /// The underlying <see cref="PeerConnection"/> object.
-        /// </summary>
-        private PeerConnection _peerConnection;
-
-        /// <summary>
-        /// Enable automatically creating a new SDP offer when the renegotiation event is fired.
-        /// This can be disabled when adding multiple tracks, to bundle all changes together and
-        /// avoid multiple round trips to the signaler and remote peer.
-        /// </summary>
-        private bool _renegotiationOfferEnabled = true;
-
-        /// <summary>
-        /// Predetermined chat data channel ID, negotiated out of band.
-        /// </summary>
-        private const int ChatChannelID = 42;
-
-        private bool isDssPolling = false;
-        private NodeDssSignaler dssSignaler = new NodeDssSignaler();
-
-        private DispatcherTimer localVideoStatsTimer = new DispatcherTimer();
-        private DispatcherTimer remoteVideoStatsTimer = new DispatcherTimer();
-
-        /// <summary>
-        /// Get the string representing the preferred audio codec the user selected.
-        /// </summary>
-        public string PreferredAudioCodec
+        public ObservableCollection<CategoryBase> Categories { get; } = new ObservableCollection<CategoryBase>()
         {
-            get
-            {
-                if (PreferredAudioCodec_Custom.IsChecked.GetValueOrDefault(false))
-                {
-                    return CustomPreferredAudioCodec.Text;
-                }
-                else if (PreferredAudioCodec_OPUS.IsChecked.GetValueOrDefault(false))
-                {
-                    return "opus";
-                }
-                return string.Empty;
-            }
-        }
-
-        /// <summary>
-        /// Get the string representing the preferred video codec the user selected.
-        /// </summary>
-        public string PreferredVideoCodec
-        {
-            get
-            {
-                if (PreferredVideoCodec_Custom.IsChecked.GetValueOrDefault(false))
-                {
-                    return CustomPreferredVideoCodec.Text;
-                }
-                else if (PreferredVideoCodec_H264.IsChecked.GetValueOrDefault(false))
-                {
-                    return "H264";
-                }
-                else if (PreferredVideoCodec_VP8.IsChecked.GetValueOrDefault(false))
-                {
-                    return "VP8";
-                }
-                return string.Empty;
-            }
-        }
-
-        public ObservableCollection<VideoCaptureDeviceInfo> VideoCaptureDevices { get; private set; }
-            = new ObservableCollection<VideoCaptureDeviceInfo>();
-
-        public VideoCaptureDeviceInfo SelectedVideoCaptureDevice
-        {
-            get
-            {
-                var deviceIndex = VideoCaptureDeviceList.SelectedIndex;
-                if ((deviceIndex < 0) || (deviceIndex >= VideoCaptureDevices.Count))
-                {
-                    return null;
-                }
-                return VideoCaptureDevices[deviceIndex];
-            }
-        }
-
-        public VideoProfileKind SelectedVideoProfileKind
-        {
-            get
-            {
-                var videoProfileKindIndex = KnownVideoProfileKindComboBox.SelectedIndex;
-                if (videoProfileKindIndex < 0)
-                {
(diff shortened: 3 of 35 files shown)


djee-ms:
Brief overview of the UI organization:
- Signaling has the usual `node-dss` stuffs
- Local Tracks allows adding new audio/video tracks from webcam and microphone
- Session is the core connection with the list of transceivers, and allows attaching tracks to transceivers and hot-swapping tracks during connection
- Media Player shows the content of a video track (audio doesn't really work yet)
- Debug Logs is the usual logs viewer
- Settings has SDP semantic, STUN server, and (for now) preferred codecs, at least until we support per-transceiver codecs.

mt_testappuwp

examples/TestAppUwp/App.xaml.cs
```diff
@@ -85,9 +85,9 @@ void OnNavigationFailed(object sender, NavigationFailedEventArgs e)
         /// <param name="e">Details about the suspend request.</param>
         private void OnSuspending(object sender, SuspendingEventArgs e)
         {
-            var deferral = e.SuspendingOperation.GetDeferral();
-            //TODO: Save application state and stop any background activity
-            deferral.Complete();
+            //var deferral = e.SuspendingOperation.GetDeferral();
+            ////TODO: Save application state and stop any background activity
+            //deferral.Complete();
```
fibann, line 90: nit: what is this for?

djee-ms: When you use automated negotiation but want to batch 2+ changes together, _e.g._ "Add some audio + video transceiver" and avoid automated negotiation starting in-between the 2 calls. I ended up not using it here, but I think that would be a nice addition to the base C# library, so left for now and will move it there later.

examples/TestAppUwp/MainPage.xaml.cs
```diff
@@ -262,47 +75,54 @@ public static string GetDeviceName()
...
+            //            DataContext = catBase
+            //        };
+            //        AutomationProperties.SetName(item, headerCat.Name);
+            //        navigationView.MenuItems.Add(item);
+            //    }
+            //    else if (catBase is Separator sepCat)
+            //    {
+            //        navigationView.MenuItems.Add(new NavigationViewItemSeparator());
+            //    }
+            //}
```
fibann, line 112: Delete?

examples/TestAppUwp/MainPage.xaml.cs
```diff
@@ -262,47 +75,54 @@ public static string GetDeviceName()
...
 
-            // This will be enabled once the signaling is initialized and ready to send data
-            createOfferButton.IsEnabled = false;
+            // Open the "Signaling" page by default
+            //navigationView.MenuItemsSource = Categories;
+            rootFrame.Navigate((Categories[0] as Category).PageType);
```
fibann, line 116: nit: change to `rootFrame.Navigate(typeof(SignalingPage))`?

examples/TestAppUwp/MainPage.xaml.cs
```diff
@@ -262,47 +75,54 @@ public static string GetDeviceName()
...
-            _peerConnection.IceCandidateReadytoSend += OnIceCandidateReadyToSend;
-            _peerConnection.IceStateChanged += OnIceStateChanged;
-            _peerConnection.IceGatheringStateChanged += OnIceGatheringStateChanged;
-            _peerConnection.RenegotiationNeeded += OnPeerRenegotiationNeeded;
-            _peerConnection.AudioTrackAdded += Peer_RemoteAudioTrackAdded;
-            _peerConnection.AudioTrackRemoved += Peer_RemoteAudioTrackRemoved;
-            _peerConnection.VideoTrackAdded += Peer_RemoteVideoTrackAdded;
-            _peerConnection.VideoTrackRemoved += Peer_RemoteVideoTrackRemoved;
+            //PreferredAudioCodecChecked(null, null);
+            //PreferredVideoCodecChecked(null, null);
```
fibann, line 122: Delete?

examples/TestAppUwp/MainPage.xaml.cs
djee-ms: No the comment is wrong but the code is right, that's the core navigation code which actually switch to the correct page (ignoring the bug when you click twice on the same page, first goes correctly, second move to last page for some reason, but let's ignore for now).

examples/TestAppUwp/MainPage.xaml.cs
djee-ms: I need to fix a bug in navigation, need to keep that code around until I understand how to fix the bug, as this may be needed.

examples/TestAppUwp/MainPage.xaml.cs
djee-ms: Let's skip all stale/unused code comments, there are 2 other changes coming after this one, and some others needed to polish the app. I just want to merge multitrack ASAP in master.

examples/TestAppUwp/MainPage.xaml.cs
fibann: Fix the comment then! 😄

examples/TestAppUwp/MainPage.xaml.cs
fibann: It would be nice to add TODOs where there's remaining work, but I'll resolve if these are going to be fixed shortly.

examples/TestAppUwp/MediaPlayerPage.xaml.cs
```diff
@@ -0,0 +1,301 @@
...
+                    Debug.Assert(DeviceName == null);
+                    label = "<none>";
+                }
+                return label;
+            }
+        }
+    }
+
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
```
fibann, line 84: Fix comment

examples/TestAppUwp/MainPage.xaml.cs
djee-ms: I'm sorry you're right it does navigate to the signaling page. The advantage of that code is we don't need to maintain 2 different places where we have the type of the page, instead `Categories` is the only source of truth. This makes also the code (but not the comment) robust to changes if we prepend a new page at index #0, it will still open the first one.

examples/TestAppUwp/MediaPlayerPage.xaml.cs
djee-ms: Fixed.

examples/TestAppUwp/MainPage.xaml.cs
djee-ms: Fixed.

examples/TestAppUwp/Model/SessionModel.cs
```diff
@@ -0,0 +1,998 @@
...
+        /// <summary>
+        /// Ready to negotiate a new session. This happens once the peer connection is
+        /// initialized, and once again after an answer is either created and applied locally
+        /// (see <see cref="NotifyLocalAnswerApplied"/>) or received from the remote peer
+        /// (see <see cref="ApplyRemoteAnswerAsync(string)"/>.
+        /// </summary>
+        Stable,
+
+        /// <summary>
+        /// New negotitation has been started locally, waiting for offer message to be created.
```
fibann, line 42:
```suggestion
        /// New negotiation has been started locally, waiting for offer message to be created.
```

examples/TestAppUwp/Model/SessionModel.cs
```diff
@@ -0,0 +1,998 @@
...
+
+        public string PreferredVideoCodecExtraParamsRemote
+        {
+            get { return _preferredVideoCodecExtraParamsRemote; }
+            set { SetProperty(ref _preferredVideoCodecExtraParamsRemote, value); }
+        }
+
+        /// <summary>
+        /// Helper class to temporarily defer an automated negotiation until this object
+        /// is diposed.
```
fibann, line 339:
```suggestion
        /// is disposed.
```

examples/TestAppUwp/Model/SessionModel.cs
```diff
@@ -0,0 +1,998 @@
...
+        /// </summary>
+        public void NotifyLocalAnswerApplied()
+        {
+            lock (_stateLock)
+            {
+                Debug.Assert(_negotiationState == NegotiationState.HaveRemoteOffer);
+                _negotiationState = NegotiationState.Stable;
+            }
+            RaisePropertyChanged("NegotiationState");
+            RaisePropertyChanged("CanNegotiate");
```
fibann, line 689: nit: add a `private set` to `NegotiationState` that raises these events instead of duplicating these lines everywhere

fibann: Hmm, it might not be easy to always deal with the lock correctly though, so feel free to ignore - maybe just add a comment to `_negotiationState` reminding maintainers to raise the events on change.

examples/TestAppUwp/NodeDssSignaler.cs
```diff
@@ -289,110 +291,119 @@ public Task SendMessageAsync(Message message)
...
-                lastPollTimeTicks = DateTime.UtcNow.Ticks;
-                _httpClient.GetAsync(requestUri, HttpCompletionOption.ResponseHeadersRead).ContinueWith((getTask) =>
+                try
                 {
-                    if (getTask.Exception != null)
+                    masterToken.ThrowIfCancellationRequested();
+
+                    // Send GET request to DSS server.
+                    lastPollTimeTicks = DateTime.UtcNow.Ticks;
+                    HttpResponseMessage response = await _httpClient.GetAsync(requestUri, HttpCompletionOption.ResponseHeadersRead);
```
fibann, line 347:
```suggestion
                    HttpResponseMessage response = await _httpClient.GetAsync(requestUri, HttpCompletionOption.ResponseHeadersRead, masterToken);
```

examples/TestAppUwp/NodeDssSignaler.cs
```diff
@@ -289,110 +291,119 @@ public Task SendMessageAsync(Message message)
...
-                    token.ThrowIfCancellationRequested();
+                    masterToken.ThrowIfCancellationRequested();
 
-                    HttpResponseMessage response = getTask.Result;
+                    // In order to avoid exceptions in GetStreamAsync() when the server returns a non-success status code (e.g. 404),
+                    // first get the HTTP headers, check the status code, then if successful wait for content.
                     if (response.IsSuccessStatusCode)
                     {
-                        response.Content.ReadAsStringAsync().ContinueWith((readTask) =>
+                        string jsonMsg = await response.Content.ReadAsStringAsync();
```
fibann, line 361: nit: should abort/discard this task if `masterToken` is canceled - though I would expect getting a string here to be reasonably fast (dunno what happens under the hood), so maybe it's fine as it is.

examples/TestAppUwp/NodeDssSignaler.cs
```diff
@@ -289,110 +291,119 @@ public Task SendMessageAsync(Message message)
...
+                        Task.Run(PollServerOnce, masterToken);
                     }
                     else
                     {
                         // Previous GET task took less time than polling delay, schedule next polling
                         long remainTicks = pollTimeTicks - deltaTicks;
                         int nextScheduleTimeMs = (int)(new TimeSpan(remainTicks).TotalMilliseconds);
-                        Task.Delay(nextScheduleTimeMs, token).ContinueWith(_ => nextAction(), token);
+                        Task.Delay(nextScheduleTimeMs, masterToken).ContinueWith(_ => PollServerOnce(), masterToken);
                     }
```
fibann, line 396: nit: this shape is clearer IMHO
```
async void PollServerLoop(...) {
  while(true) {
    // poll server once
    long remainTicks = Math.max(0, ...);
    await Task.Delay(remainTicks);
  }
}
```

examples/TestAppUwp/NodeDssSignaler.cs
```diff
@@ -289,110 +291,119 @@ public Task SendMessageAsync(Message message)
...
-                });
-            };
+                }
+                catch (Exception ex)
+                {
+                    OnFailure?.Invoke(ex);
+                    masterTokenSource.Cancel();
```
fibann, line 401: AFAICS we shouldn't need to cancel `masterTokenSource` since it's only used in this loop which is terminating anyway.

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Not sure I understand. `ReadAsStringAsync()` doesn't take a cancellation token, and the `masterToken` is tested right before and right after the call. How do you expect to do better than that?

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Right, it is more obvious now that it's rewritten as an actual loop. I thought we could have a pending task scheduled at that point, but thinking about it more I think you're right that's impossible.

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Fixed.

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Fixed.

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Fixed.

examples/TestAppUwp/Model/SessionModel.cs
djee-ms: Fixed. I added a single private method to avoid duplicating the code, and explicitly coded the reason why `CanNegotiate` is only raised sometimes, which was not obvious in the previous code.

examples/TestAppUwp/Model/SessionModel.cs
djee-ms: Fixed.

examples/TestAppUwp/Model/SessionModel.cs
djee-ms: Fixed.

examples/TestAppUwp/NodeDssSignaler.cs
fibann: There are a number of ways to work around lack of a CancellationToken param, see https://www.planetgeek.ch/2016/12/02/async-method-without-cancellation-support-oh-my/ for example. But might not be strictly needed here, there should be no dire consequences if the task stays alive a few ms longer.

examples/TestAppUwp/SignalingPage.xaml.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace TestAppUwp
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public sealed partial class SignalingPage : Page
```
fibann, line 12: nit: fill comment, or make internal so we don't need boilerplate docs.

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Right, good resource, I didn't know about that pattern. However I'd close as won't fix here though because the `node-dss` messages are tiny (few KB at most for offer/answer), and `ReadAsStringAsync()` most likely immediately returns (note that we already read the HTTP header above, so we're reading only the content here, and by the time we reached that code there's an even higher chance the body of the HTTP message was received). In the linked example, `Task.WhenAny()` will indeed return and present a completed task to the outer code waiting for it, but the actual task doing the work internally (not the one with delay) doesn't seem to be cancelled, so is just running in the background until it's completed and discarded. So the only thing that changes is how fast is the outermost `Task`; turns out `StartPollingAsync()` does not return that `Task` anyway so nobody can wait for it, so this makes no difference here.

examples/TestAppUwp/SignalingPage.xaml.cs
djee-ms: Fixed.

examples/TestAppUwp/NodeDssSignaler.cs
fibann: > but the actual task doing the work internally (not the one with delay) doesn't seem to be cancelled

FWIW I guess that you could kill that task too by disposing the HttpContent once you get control back. Anyway, let's move on!

examples/TestAppUwp/ViewModel/AudioCaptureViewModel.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                Track = track,
+                TrackImpl = track,
+                IsRemote = false,
+                DeviceName = DefaultAudioDeviceName
+            });
+            SessionModel.Current.LocalTracks.Add(new TrackViewModel(Symbol.Volume) { DisplayName = DefaultAudioDeviceName });
+        }
+
+        private async Task RequestMediaAccesssAsync(StreamingCaptureMode mode)
```
fibann, line 52: s/sss/ss

djee-ms: pssss.... 😆

djee-ms: Fixed.

examples/TestAppUwp/SignalingPage.xaml.cs
djee-ms: Note: DocFX doesn't run on TestAppUWP at the moment, so comments are technically not required, though they do provide IMHO some huge benefit for us in remembering how things work and for other users in learning those things. So I like to even comment on protected/private/internal items.

examples/TestAppUwp/DebugConsolePage.xaml.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Debugger.Log(4, "TestAppUWP", message);
+            message += "\n";
+            _fullText += message;
+            RaisePropertyChanged("FullText");
+        }
+    }
+
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
```
fibann, line 36: Comment is incorrect.

djee-ms: Fixed.

examples/TestAppUwp/ViewModel/Converters.cs
```diff
@@ -0,0 +1,205 @@
...
+using System;
+using Microsoft.MixedReality.WebRTC;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Data;
+
+namespace TestAppUwp
+{
+    /// <summary>
+    /// Convert <c>true</c> to <c>false</c>, and anything else (including <c>false</c>) to <c>true</c>.
```
fibann, line 13: nit: do we expect this to get values that aren't `true` or `false`? Should we throw if that's not the case?

djee-ms: For UI I think it's better to try and be as silent as possible rather than throw here and break the UI. I think throwing here also crashes the app since it's called via runtime reflection by the XAML implementation.

examples/TestAppUwp/ViewModel/Converters.cs
```diff
@@ -0,0 +1,205 @@
...
+    /// The back-conversion is also available, and simply negate the boolean value.
+    /// </summary>
+    public class BooleanInverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, string langauge)
```
fibann, line 18: s/langauge/language

examples/TestAppUwp/ViewModel/TransceiverViewModel.cs
```diff
@@ -0,0 +1,200 @@
...
+        public bool IsVideoTransceiver => Transceiver.MediaKind == MediaKind.Video;
+
+        /// <summary>
+        /// Collection of available sender tracks for the currently selected transceiver
+        /// media kind. This collection is rebuilt each time the selected transceiver changed
+        /// and its media kind is different from the previous transceiver.
+        /// </summary>
+        /// FIXME - There is no check to avoid assigning the same sender track to 2 tranceivers,
+        /// which asserts internally as this is not supported.
+        public ObservableCollection<SenderTrackViewModel> AvailableSenders { get; }
```
fibann, line 140: Unused?

examples/TestAppUwp/ViewModel/VideoCaptureViewModel.cs
```diff
@@ -0,0 +1,343 @@
...
+            {
+                Track = track,
+                TrackImpl = track,
+                IsRemote = false,
+                DeviceName = deviceInfo.DisplayName
+            });
+            SessionModel.Current.LocalTracks.Add(new TrackViewModel(Symbol.Video) { DisplayName = deviceInfo.DisplayName });
+        }
+
+        private async Task RequestMediaAccesssAsync(StreamingCaptureMode mode)
```
fibann, line 187: Another "sss"

djee-ms: Fixed.

examples/TestAppUwp/ViewModel/Converters.cs
djee-ms: Fixed.

examples/TestAppUwp/ViewModel/TransceiverViewModel.cs
djee-ms: This is bound in `SessionPage.xaml`.

examples/TestAppUwp/ViewModel/Converters.cs
fibann: Yeah, I find the asymmetric behavior a bit weird, but it doesn't matter too much if this is only going to be only used by XAML machinery.

examples/TestAppUwp/ViewModel/TransceiverViewModel.cs
djee-ms: Or not. Fixed.

djee-ms:
> The number of clicks necessary to simply capture and stream has tripled, this is going to be a pain when prototyping, especially on HoloLens.

Agreed, I had thought about the same thing. And possibly adding some "profiles" like "auto-create 2 transceivers, one audio and one video, and associate some webcam and mic tracks to them". That could be saved and restored on startup or manually. That would greatly simplify testing.

> The amount of boilerplate related to ViewModels and property notifications is mildly worrying, but I am not familiar with XAML nor MVVM so I can't say if it's a net win or not, nor am able to suggest a way to improve this.

This has simplified a lot of code related to maintaining the state of controls. The previous version was not using binding at all, so required some `if (enabled) then { control1.show(); control2.hide(); }`, whereas now it makes it more clear what happens and more importantly we're sure that we don't miss a call site. For the rest, the amount of UI somewhat increased, which makes the amount of code increase too, so although I had a first initial reaction similar to yours, I think in the end it's probably better this way than it was before. And certainly a lot more maintainable.

PR #279, opened 2020-04-15
Add build script for libwebrtc Windows

djee-ms:
Add some Powershell build script `tools/build/libwebrtc/build.ps1` to build libwebrtc on Windows Desktop and UWP, after checking out and setting up the Google repository in `external/libwebrtc`, and automatically applying the patches of [WinRTC](https://github.com/microsoft/winrtc) over the M80 branch.

This is a stub script, which is missing some functionalities (select build triplet, select branch M80 vs master), but already greatly simplifies the user experience.

This change adds 2 submodules:
- `external/depot_tools` from Google, to be able to easily find and use it
- `external/winrtc` to have a particular version of the patches knows to work and available in a known location


djee-ms:
Here's a sample output of running that script incrementally on an already set up repository :

image

libs/Microsoft.MixedReality.WebRTC.Native/src/uwp/Microsoft.MixedReality.WebRTC.Native.UWP.vcxproj
```diff
@@ -104,7 +104,7 @@
       <CompileAsWinRT>false</CompileAsWinRT>
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
-      <AdditionalIncludeDirectories>..\..\include;..\;$(WebRTCCoreRepoPath);$(WebRTCCoreRepoPath)out\$(Configuration)_$(PlatformTarget)\gen;$(WebRTCCoreRepoPath)third_party\abseil-cpp;$(WebRTCCoreRepoPath)third_party\libyuv\include;$(WebRTCCoreRepoPath)out\$(Configuration)_$(PlatformTarget)\gen\protoc_out;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\..\include;..\;$(WebRTCCoreRepoPath);$(WebRTCCoreRepoPath)out\$(Configuration)_$(PlatformTarget)_uwp\gen;$(WebRTCCoreRepoPath)third_party\abseil-cpp;$(WebRTCCoreRepoPath)third_party\libyuv\include;$(WebRTCCoreRepoPath)out\$(Configuration)_$(PlatformTarget)_uwp\gen\protoc_out;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
fibann, line 107: nit: I was thinking to switch the build path to something like `out\(win|winuwp)\$(PlatformTarget)\$(Configuration)` so it's a bit more "standard"

djee-ms: Good point, that even better. I will change to that.

tools/build/libwebrtc/README.md
```diff
@@ -1,18 +1,36 @@
...
-    * **Windows, UWP**: TBD - in development.
+## Windows Desktop and UWP
+
+1. [Powershell 5.1](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-windows-powershell?view=powershell-5.1) is required to build libwebrtc. It is installed by default with Windows 10 and Windows Server 2016.
+
+2. Open a Powershell console and change directory to the folder containing this README, then run the build script `.\build.ps1`. This command automatically does all steps necessary to build `webrtc.lib`:
+   - Fetch the Google repository of libwebrtc inside `external/libwebrtc`
+   - Checkout the M80 branch
+   - Apply the UWP patches from [WinRTC](https://github.com/microsoft/winrtc) to allows building on UWP
+   - Synchronize dependencies (`glcient sync`)
```
fibann, line 13:
```suggestion
   - Synchronize dependencies (`gclient sync`)
```

tools/build/libwebrtc/build.ps1
```diff
@@ -0,0 +1,282 @@
...
+
+    $repoRoot = Get-RootFolder
+    Write-KeyValue "Repository root: " $repoRoot
+
+    # Add depot_tools folder to PATH
+    $externalFolder = Join-Path $repoRoot "external/"
+    $depotToolsPath = Join-Path $externalFolder "depot_tools"
+    if (!($env:PATH.Contains($depotToolsPath)))
+    {
+        # Be sure to append, not prepend. depot_tools/ contains some git version,
```
fibann, line 51: Are we sure that appending is good enough? I am not sure what Google's guidelines say.

tools/build/libwebrtc/build.ps1
```diff
@@ -0,0 +1,282 @@
...
+    Write-TaskStart "Fetching Google repository, this might take some time..."
+    try {
+        Invoke-Expression "fetch --nohooks --no-history webrtc"
+    }
+    catch {
+        Write-Failure "Failed to fetch Google repository 'webrtc'."
+        throw
+    }
+    
+    # Checkout + sync
```
fibann, line 93: We should separate checkout script from build script so one can build with local changes to WebRTC. Can be done as a separate change though.

tools/build/libwebrtc/build.ps1
djee-ms: Yeah [they say "to the front"](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/depot_tools_tutorial.html#_setting_up). But the recommendations are for installing `depot_tools` when used with Chromium as I understand. The reason they do that AFAIK is to inject some custom git commands like `git freeze` and `git thaw`, but we don't need those since we don't intend to actually make any change to libwebrtc.

But if I do that then the git version inside `external/depot_tools/` is used and the credential manager _etc._ are not configured, so for example I couldn't push that change anymore to GitHub. So I think it's safe to ignore and use the user's installed git instead, since we don't rely on most tools of `depot_tools`, only `gclient` and `fetch`.

tools/build/libwebrtc/build.ps1
djee-ms: Agreed.

tools/build/libwebrtc/build.ps1
fibann: Maybe we can prepend and then restore the original path at the end of the script, just to be sure? Or just add a warning that things _shouldn't_ but _might_ break because of this 😄

tools/build/libwebrtc/build.ps1
```diff
@@ -0,0 +1,282 @@
...
+    # Add depot_tools folder to PATH
+    $externalFolder = Join-Path $repoRoot "external/"
+    $depotToolsPath = Join-Path $externalFolder "depot_tools"
+    if (!($env:PATH.Contains($depotToolsPath)))
+    {
+        # Be sure to append, not prepend. depot_tools/ contains some git version,
+        # and prepending will make that version the default one, which may confuse
+        # some users as e.g. on Windows it won't have the credential manager set up
+        # like the one the user had installed already.
+        $env:PATH = "$env:PATH;$depotToolsPath"
```
stephenatwork, line 55: This seems to change the environment of the calling process. For cleanliness, does PS have a way to keep the changes in the scope of the script only? Or to fork a subshell with the changed env?

libs/Microsoft.MixedReality.WebRTC.Native/src/uwp/Microsoft.MixedReality.WebRTC.Native.UWP.vcxproj
djee-ms: Fixed.

tools/build/libwebrtc/README.md
djee-ms: Fixed.

tools/build/libwebrtc/build.ps1
djee-ms: Doesn't look like there is a good solution out of the box like on Unix : https://stackoverflow.com/questions/1420719/powershell-setting-an-environment-variable-for-a-single-command-only

I think the way to go is to run jobs :
- https://www.reddit.com/r/PowerShell/comments/qvk0d/powershell_fork/
- https://powerscripting.wordpress.com/2010/06/03/a-look-at-powershell-jobs-part-1/

tools/build/libwebrtc/build.ps1
```diff
@@ -157,13 +159,10 @@ function Write-GnArgs([string]$Platform, [string]$Arch, [string]$Config, [string
...
-            $extra="_win32"
         }
         "UWP" {
             $target_os="winuwp"
-            $extra="_uwp"
-            $UwpContent = @"
-
+            $extra_content=@"
 # Use WinRT video capture for UWP
 rtc_win_video_capture_winrt=true
```
fibann: I just remembered this: the latest recommendation is using the WinRT capture module on win32 too.

djee-ms: Fixed.

tools/build/libwebrtc/build.ps1
djee-ms: Fixed with scope.

tools/build/libwebrtc/build.ps1
djee-ms: Fixed with scope.

tools/build/libwebrtc/mrwebrtc.ps1
```diff
@@ -0,0 +1,270 @@
...
+{
+    Write-TaskStart "Installing dependencies..."
+
+    # Check python
+    try
+    {
+        Get-Command "python" -ErrorAction Stop | Out-Null
+    }
+    catch
+    {
```
stephenatwork, line 131: nit: are we doing Egyptian braces? There are mixed styles.

djee-ms: I don't quite know what's the canonical way in PowerShell. Seems like VSCode "Format Document" command produces some inline braces instead, so happy to change to that.

stephenatwork: It needn't block this PR. Just something for the future. BTW I went looking for opinions and hadn't realized that 
```
foo { 
  bar
}
```
and
```
foo
{
  bar
}

```
are quite different things to powershell. Which is why some people recommend K&R braces.
https://powershell.org/2013/07/powershell-great-debate-formatting-constructs/#comment-4968

PR #283, opened 2020-04-16
Add multi-track support

djee-ms:
Add support for multiple audio and video tracks per peer connection.

This change introduces a new _transceiver API_ closely related to [the same-named concept in the WebRTC 1.0 standard](https://www.w3.org/TR/webrtc/#rtcrtptransceiver-interface). The transceiver API allows the user to add audio and video transceivers to a peer connection, then optionally associate some local track to send some media through those transceivers and/or receive some media from the remote peer via a remote track associated with the transceiver. The new API is more flexible and more powerful, offering more control to the user and opening up many use cases not previously available, at the expense of some moderately increased complexity. In particular it allows:
- multiple audio and video tracks per peer connection, limited only by the bandwidth
- hot-swapping a track for another track without a renegotiation, dramatically reducing the delay of replacing a track with another one (_e.g._ switching to "on hold" music instead of voice chat)
- "track warm up" (getting a track transport ready even before the track media is available) and more aggressive media transport establishing (callee peer can start sending media to caller peer earlier than before)

This change constitutes a major API breaking change, which moves tracks management away from the `PeerConnection` object to the new `Transceiver` object. **Users are encouraged to read the [migration guide](https://github.com/microsoft/MixedReality-WebRTC/blob/merge/multi_track/docs/manual/migration-guide.md) associated with this change** to migrate existing v1.0-style applications. In particular, understanding the functioning of transceivers and the **asymmetrical** role of the _caller peer_ (the peer initiating a session with a new offer) and the _callee peer_ (the peer responding to a remote offer) are critical to be able to use this new API correctly.


fibann:
Is there anything that needs being looked at, or is it all content from previous reviews?

libs/Microsoft.MixedReality.WebRTC.Native/src/peer_connection.cpp
```diff
@@ -1698,11 +1730,22 @@ void PeerConnectionImpl::SynchronizeTransceiversUnifiedPlan(bool remote) {
...
   RTC_DCHECK_GE(rtp_transceivers.size(), wrappers.size());
   auto it_wrapper = wrappers.begin();
+  RTC_LOG(LS_INFO) << "Synchronizing " << rtp_transceivers.size()
+                   << " RTP transceivers with " << wrappers.size()
+                   << " transceiver wrappers (remote = " << remote << ").";
   for (auto&& rtp_tr : rtp_transceivers) {
     const int mline_index = ExtractMlineIndexFromRtpTransceiver(rtp_tr);
     // Create a wrapper if it doesn't exist yet
-    if (it_wrapper == wrappers.end()) {
+    if ((it_wrapper == wrappers.end()) || ((*it_wrapper)->impl() != rtp_tr)) {
```
fibann, line 1739: What does it mean here if there is another transceiver already?

djee-ms: The transceivers are sorted by address in memory. You have 2 arrays being iterated in parallel. If one value doesn't match with the other, it means one array (`wrappers`) is missing an element and you need to insert it.
```
RTP | Wrapper
----|--------
A   | A
B   | B   v
C   | D   ^ Need to insert here because C != D
D   |
```

fibann: Yeah it's reasonably clear once you un-hide the surrounding lines 😄 
nit: might be worth adding a comment for good measure.

djee-ms: Yes the comment confuses the reader I think, but thinking about the actual algorithm (sort 2 arrays to avoid O(n^2) comparison) makes things clear.

djee-ms: Fixed.

PR #287, opened 2020-04-21
[TestAppUWP] Add support for chat channels

jehumb-havok:
Add support for data channels in TestAppUWP in the form of "chat channels" transferring UTF-8 text messages. This restores parity with the old pre-multitrack version of TestAppUWP.


djee-ms:
> PR mentions "channels" but there's only _one_ channel at the moment, right?

I think there's no "add channel" button, but otherwise the rest should work. Can fix later.

PR #292, opened 2020-04-23
Rename native library to mrwebrtc

djee-ms:
This change performs some clean-up for consistency and clarity:
- Rename the native project from `Microsoft.MixedReality.WebRTC.Native` to `mrwebrtc`, for consistency with the output it produces `mrwebrtc.dll`. This also clarifies the fact that this project is no longer the C++ library, but instead is an internal implementation module.
- Rename the project's folder in `libs/` in the same way.
- Move all build-related files (Visual Studio projects and filters, `packages.config`, _etc._) to the `tools/build/mrwebrtc/` folder, out of the source tree, for consistency with the Pitchfork Layout this repository intend to follow.
- Rename `mrwebrtc.targets` to `mrwebrtc.props`, as the file only defines some properties.
- Fix some misconfiguration with 32/64-bit dependency to `mrwebrtc.dll` in the C# test assembly.


tools/build/mrwebrtc/uwp/mrwebrtc-uwp.vcxproj
```diff
@@ -50,7 +50,7 @@
     <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
fibann, line 50: nit: any particular reason to move the vcxprojs to tools/build?

djee-ms: A bit of consistency. The other build tools for Android and libwebrtc (undock) are already there.

PR #301, opened 2020-04-27
Only create local tracks if component is active

djee-ms:
Avoid creating a local audio or video track (webcam/microphone) if the
component is not active. Instead leave the Track property as NULL.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime/VideoSenderTests.cs
+++ b/libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime/VideoSenderTests.cs
@@ -118,7 +118,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime
             Assert.IsTrue(sender.IsCapturing);
         }
 
-        [Test(Description = "Capture can be manyally started even if not active.")]
+        [Test(Description = "Capture can be manually started even if not active.")]
         public async void CaptureStartManualInactive()
         {
             // Create the peer connections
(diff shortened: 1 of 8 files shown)


libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/Media/AudioSender.cs
```diff
@@ -197,8 +197,9 @@ internal override async Task AttachTrackAsync()
             //< FIXME - Multi-track override!!!
             Transceiver.PeerConnection.PreferredAudioCodec = PreferredAudioCodec;
 
-            // Ensure the local sender track exists
-            if (Track == null)
+            // Ensure the local sender track exists, but do not create
+            // if the component is not active.
+            if ((Track == null) && isActiveAndEnabled)
```
fibann, line 202: nit: if the component is enabled after `AttachTrackAsync` I suppose we should create the track in `OnEnable`. Or at least we should ensure that what happens if one does that is deterministic and documented.

djee-ms:
Rebased, handled more corner cases, and added a bunch of Unity tests to validate all of that. Can you please re-review?

libs/Microsoft.MixedReality.WebRTC.Unity/Assets/Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime/VideoSenderTests.cs
```diff
@@ -0,0 +1,380 @@
...
+            Assert.IsNull(sender.Track);
+            Assert.IsFalse(sender.IsCapturing);
+
+            // Manually start capture
+            await sender.StartCaptureAsync();
+            Assert.IsNotNull(sender.Track);
+            Assert.IsTrue(sender.IsCapturing);
+        }
+
+        [Test(Description = "Capture can be manyally started even if not active.")]
```
fibann, line 121:
```suggestion
        [Test(Description = "Capture can be manually started even if not active.")]
```

djee-ms: Fixed.

PR #315, opened 2020-05-01
Fix release pipeline with component detection

djee-ms:
- Fix the release pipeline building the Google's implementation library `libwebrtc` to build on on-prem agents.
- Add a Component Governance task before the build to detect components in use by the project.
- Ensure the security report for those components is clean by deleting unused code before detection and building to make sure potentially vulnerable code present in Google's build tools _etc._ is not used in the build and not shipped.
- Update `cgmanifest.json` with the correct commits for all used git repositories.

### DIFF
--- a/tools/ci/templates/steps-install-python2x.yaml
+++ b/tools/ci/templates/steps-install-python2x.yaml
@@ -1,0 +1,69 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# [TEMPLATE] Steps to install Python 2.x on the build agent
+
+parameters:
+# Temporary folder where to download the Python 2.x NuGet package before moving
+# it to the agent's tools folder. This folder will be completely cleared.
+- name: 'tempFolder'
+  type: string
+  default: ''
+# Python version (e.g. 2.7.17)
+- name: 'pythonVersion'
+  type: string
+  default: ''
+
+steps:
+
+# Clean-up the temporary folder
+- powershell: 'Remove-Item "${{parameters.tempFolder}}\*" -Force -Recurse -ErrorAction Ignore'
+  displayName: 'Clean-up temp install folder'
+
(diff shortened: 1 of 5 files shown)


djee-ms:
For context the Google checkout/build tools are pulling a lot more code than necessary to just build `libwerbrtc` (we do not build tests/examples/tools), and some of those use old dependencies with known security vulnerabilities. This has no impact on the actual security of MixedReality-WebRTC since those tests/examples/tools are not even built, but the Component Governance task which monitors security is conservative and will report vulnerabilities for any component detected on the Azure Agent disk. So we delete those unused components with known vulnerabilities before scanning and before building, which ensures that they cannot possibly make it into MixedReality-WebRTC and makes security checks pass.

tools/ci/templates/steps-install-python2x.yaml
```diff
@@ -0,0 +1,69 @@
...
+    foreach ($f in $(Get-ChildItem -Path $(Agent.ToolsDirectory)\Python\${{parameters.pythonVersion}} -Recurse)) {
+        Write-Host $f.FullName
+    }
+  displayName: 'Move Python ${{parameters.pythonVersion}}'
+
+# Register the new tool with the agent
+- powershell: |
+    Write-Host "Creating '.complete' file for build tool registration..."
+    New-Item "$(Agent.ToolsDirectory)\Python\${{parameters.pythonVersion}}\x64.complete" -Force
+  displayName: 'Register Python ${{parameters.pythonVersion}} with build agent'
```
fibann, line 50: nit: I am not sure whether this is common knowledge, if not it's worth adding a comment/link about how this works.

djee-ms: I can't find the link to the docs, I'll leave for now. This is part of the standard procedure to install a tool on an agent; admittedly not common knowledge for devs, but should be for devops.

PR #316, opened 2020-05-01
Delete unused components before CG scan

djee-ms:
Before Component Governance scanning and `libwebrtc` building occur, delete components checked out on disk by Google's tools but unused, and which trigger some security vulnerability alerts. This makes sure those vulnerabilities cannot possibly make it into MixedReality-WebRTC, and makes the security checks pass.

### DIFF
--- a/tools/build/libwebrtc/mrwebrtc.ps1
+++ b/tools/build/libwebrtc/mrwebrtc.ps1
@@ -95,6 +95,16 @@ function Install-GoogleRepository {
         throw
     }
 
+    # Delete sources not needed; this prevent security alerts on unused components,
+    # and makes the overall checkout size smaller.
+    $toolsFolder = Join-Path $libwebrtcFolder "tools" -Resolve
+    Rename-Item -Path $toolsFolder -NewName "_tools"
+    $oldToolsFolder = Join-Path $libwebrtcFolder "_tools" -Resolve
+    New-Item -Path $toolsFolder -ItemType Directory | Out-Null
+    Move-Item -Path $(Join-Path $oldToolsFolder "clang") -Destination $toolsFolder
+    Move-Item -Path $(Join-Path $oldToolsFolder "protoc_wrapper") -Destination $toolsFolder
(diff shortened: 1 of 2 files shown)


tools/build/libwebrtc/mrwebrtc.ps1
```diff
@@ -59,6 +59,30 @@ function Initialize-BuildEnvironment {
     $env:GYP_MSVS_VERSION = "2019"
 }
 
+# Post-checkout clean-up
+function Clear-PostCheckout([string]$SourceFolder) {
+    # Remove tools/ (except tools/clang/ and tools/protoc_wrapper/)
+    $toolsFolder = Join-Path $SourceFolder "tools" -Resolve
+    Rename-Item -Path $toolsFolder -NewName "_tools"
+    $oldToolsFolder = Join-Path $SourceFolder "_tools" -Resolve
```
stephenatwork, line 67: Does "-resolve" throw  if the path doesn't exist?

djee-ms: Yes. And that's OK here because `tools/` always exists, and must be renamed successfully, so `_tools` also always exists.

tools/build/libwebrtc/mrwebrtc.ps1
```diff
@@ -95,6 +119,13 @@ function Install-GoogleRepository {
         throw
     }
 
+    # Delete sources not needed; this prevent security alerts on unused components,
+    # and makes the overall checkout size smaller.
+    Clear-PostCheckout -SourceFolder $libwebrtcFolder
+
+    # Remove depot_tools/external_bin/gsutil
+    Remove-Item -Path $(Join-Path $externalFolder "depot_tools/external_bin/gsutil" -Resolve) -Force -Recurse | Out-Null
```
fibann, line 127: nit: specify that this is done because of security alerts too (or simply group with line above).

djee-ms: Grouped with above.

Issue #92, opened 2019-10-04
Support for connecting a WebRTC audio track to a Unity AudioSource


eanders-MS:
Can you help me solve a mystery? Looking through the Unity sample project, I'm not seeing how the audio track is getting played. Scene doesn't have a Unity `AudioSource` MonoBehaviour. The code doesn't do anything with audio frames. How is audio getting to the speakers?

Context: In my project, I need to pipe the WebRTC audio track to a spatial audio emitter so that the audio is played at the media player's location in the scene. How can I pipe the audio track to a Unity `AudioSource`?

Thanks.

djee-ms:
Yes, this is not implemented, as pointed out in the documentation : https://microsoft.github.io/MixedReality-WebRTC/manual/unity-remoteaudiosource.html

Currently the audio output is managed internally in C++ by the WebRTC code which sends it out to the audio device directly for playback. Unity is never involved.

Implementing what you need, which is a perfectly valid use case that we should support, requires disabling the automatic audio output, and instead use a callback mechanism like video, then plug that somehow into Unity, likely via an `AudioSource` component. But I have not investigated the feasibility of this yet.

eanders-MS:
Thanks for the info. Is it on your roadmap to address this soon? :)

stephenatwork:
FWIW, you can disable automatic rendering of audio tracks in UWP by editing api.cpp:189
factoryConfig->audioRenderingEnabled = false; //was "true"

For desktop, I haven't confirmed this yet, but it seems to be the null AudioDeviceModule passed to webrtc::CreatePeerConnectionFactory which, if null, gets a default module created much later in WebRtcVoiceEngine::Init()

This still leaves the task of hooking up the frame event of course.

stephenatwork:
Linking #99

stephenatwork:
From #99 the RemoteAudioSource component can now play to a sibling AudioSource component.

The latency is much, much better than the previous AudioClip version thanks! And it shouldn't be too difficult to move the resampling code to a different thread if needed. (currently it happens lazily on the unity audio thread).

There are still some TODOs: Experimentally, Unity will only call OnAudioFilterRead if there is a sibling (Unity)AudioSource component. Is checking for the existence of the sibling (say on Play()) sufficient to toggle between direct & spatial audio? Or should we have an additional flag to explicitly state the users intention? We could give a sensible error message if the flag doesn't match the existence of AudioSource.

[...]

PR #319, opened 2020-05-05
Move RenderToDevice to RemoteAudioTrack

fibann:
See #92.


libs/Microsoft.MixedReality.WebRTC/RemoteAudioTrack.cs
```diff
@@ -35,6 +44,31 @@ public bool Enabled
...
+        /// <remarks>
+        /// The default behavior is for every remote audio frame to be passed to
+        /// remote audio frame callbacks, as well as rendered automatically on the
+        /// system audio device. If |false| is passed to this function, remote audio
+        /// frames will still be received and passed to callbacks, but won't be rendered
+        /// on the system device.
+        ///
+        /// NOTE: Changing the default behavior is not supported on UWP.
+        /// </remarks>
+        public void RenderToDevice(bool render)
```
djee-ms, line 59: nit: `OutputToDefaultDevice()`? I don't mind too much `Render`, but I'd like `Default` to appear somewhere for clarity, no?

libs/mrwebrtc/src/toggle_audio_mixer.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  return true;
+}
+
+void ToggleAudioMixer::TryAddToBaseImpl(KnownSource& known_source) {
+  bool added_succesfully = base_impl_->AddSource(known_source.source);
+  if (!added_succesfully) {
+    RTC_LOG_F(LS_WARNING) << "Cannot mix source "
+                          << known_source.source->Ssrc();
+    known_source.is_rendered = false;
```
djee-ms, line 44: Shall we propagate this error back to `AddSource()` above? For some reason this method is called `Try...` but doesn't return anything.

libs/mrwebrtc/src/toggle_audio_mixer.h
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+
+namespace Microsoft::MixedReality::WebRTC {
+
+  /// Can mix selected audio sources only.
```
djee-ms, line 8: nit: remove leading comment spaces

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
```diff
@@ -29,6 +29,14 @@ internal class RemoteAudioTrackInterop
             EntryPoint = "mrsRemoteAudioTrackIsEnabled")]
         public static extern int RemoteAudioTrack_IsEnabled(IntPtr trackHandle);
 
+        [DllImport(Utils.dllPath, CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Ansi,
+            EntryPoint = "mrsRemoteAudioTrackRenderToDevice")]
+        public static extern void RemoteAudioTrack_RenderToDevice(IntPtr trackHandle, int render);
+
+        [DllImport(Utils.dllPath, CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Ansi,
+            EntryPoint = "mrsRemoteAudioTrackIsRenderingToDevice")]
+        public static extern int RemoteAudioTrack_IsRenderingToDevice(IntPtr trackHandle);
```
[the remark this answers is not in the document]
fibann: Logged as #321. Will leave it like this for now (doesn't make it worse).

libs/Microsoft.MixedReality.WebRTC/RemoteAudioTrack.cs
fibann: I actually do like "output" more than "render".

As discussed in chat, "Device" here means "whatever device WebRTC has decided to play tracks on", so "default" might be misleading here if WebRTC is configured not to use the default system device. Also to make things weirder WebRTC won't use the "default device" by default on Windows (!), but the "default communications device" which might be different. So let's avoid using "default" unless we can make the context clear.

libs/mrwebrtc/src/toggle_audio_mixer.cpp
fibann: There's not a lot of semantics attached to the `AddSource` return value, from what I see the caller just prints an error message and `AudioMixerImpl` always return true. If we want to handle `false` "properly" on our end, we need to roll back addition to the local map in `AddSource`, and make `RenderSource` failable too (together with all the long chain of callers). Since AudioMixer is marked as WIP this seems over-engineering - I'd say let's print an error rather than a warning and call it done.

libs/Microsoft.MixedReality.WebRTC/RemoteAudioTrack.cs
fibann: Changed the comments to talk about the "device used by WebRTC" for clarity.

libs/mrwebrtc/src/toggle_audio_mixer.cpp
fibann: Changed warning to error.

Issue #313, opened 2020-04-30
[Android] Infinite wait in SetRemoteDescriptionAsync()


JWSon:
Hello,

I've built the android tools as instruction on `master` and created the libwebrtc.aar and mrwebrtc.aar.
Webrtc initialization was successful, but after receiving a SDP data from server the `SetRemoteDescriptionAsync()` is waiting indefinitely.

Return value of `PeerConnection_SetRemoteDescriptionAsync()` on `PeerConnectionInterop.cs` is `Utils.MRS_SUCCESS`, but `args.completedEvent.Wait()` of `RemoteDescArgs` still waits.
I guess it is related to synchronize transceiver in native c++ since `RemoteDescArgs` callback method is called on `SetRemoteSessionDescObserver`.

the build summary is as follows.

1. built libwebrtc.aar on Hyper-V Ubuntu 18.04
2. built mrwebrtc.aar on Windows using Android Studio.
    - added  `{mr-webrtc-native-dir}/src/utils.cpp` 
3. built Microsoft.MixedReality.WebRTC.dll on Visual studio 2019
4. made Unity project and configured lib and gameobject like user guide.
5. changed NodeDss to Websocket (Newtonsoft)

It would be appreciated if you would give some comments.

djee-ms:
My guess is that the remote description fails to apply. Do you have access to the Visual Studio Output window and can see the debug logs printed by the native implementation? If so, can you please try to find the line "Remote description set. err=" and see if there's an error message? I see that the callback is not invoked if there's an error, which effectively makes `SetRemoteDescriptionAsync()` deadlock, we should fix that. However this only happens if the remote description does not apply correctly, so you already have some other issue here I think.

djee-ms:
I think #320 should fix the deadlock, if you want to try it locally while the review is pending to confirm that it works for you.

JWSon:
Thank you for your kind information.

I've applied #320 sources and added try-catch statement to `SetRemoteDescriptionAsync()`.
now I can check a result and error message of `RemoteDescriptionApplied` callback. like your above comment, it seems to not apply the remote description correctly.
the result and error message are "MRS_E_INVALID_PARAMETER" and "Failed to set remote offer sdp: Failed to set remote video description send parameters".
I am suspicious about unified plan and plan b of sdp semantics.
the sdp data which is generated from my server is as follows.
```
v=0
o=- 8887904561871063630 0 IN IP4 0.0.0.0
s=-
[...]

PR #320, opened 2020-05-05
Prevent deadlock on SetRemoteDescriptionAsync fail

djee-ms:
Ensure that the task returned by SetRemoteDescriptionAsync() always
terminates, even when the underlying job fails to apply the remote
description, and that any error is propagated back to the caller.

Bug: #313


libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
```diff
@@ -814,31 +824,34 @@ internal struct RemoteVideoTrackAddedInfo
...
         public static Task SetRemoteDescriptionAsync(PeerConnectionHandle peerHandle, string type, string sdp)
         {
-            return Task.Run(() =>
+            var args = new RemoteDescArgs
+            {
+                callback = RemoteDescriptionApplied,
+                tcs = new TaskCompletionSource<bool>() // dummy bool, unused
+            };
+
+            Task.Run(() =>
```
fibann, line 842: You don't need this Task anymore, the request can be started synchronously.

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
```diff
@@ -814,31 +824,34 @@ internal struct RemoteVideoTrackAddedInfo
...
+            public PeerConnectionRemoteDescriptionAppliedDelegate callback;
+            public TaskCompletionSource<bool> tcs;
         }
 
         public static Task SetRemoteDescriptionAsync(PeerConnectionHandle peerHandle, string type, string sdp)
         {
-            return Task.Run(() =>
+            var args = new RemoteDescArgs
+            {
+                callback = RemoteDescriptionApplied,
```
fibann, line 838: The `callback` field is needed to keep the delegate alive, right? Probably needs a comment.

libs/Microsoft.MixedReality.WebRTC/Interop/PeerConnectionInterop.cs
djee-ms: You're probably right. Trying to think of a reason to keep it on a background .NET worker thread, but I don't think there's any for that one.

PR #322, opened 2020-05-05
Use enum and message type for SDP messages

djee-ms:
Add an enum type for the SDP message type, and a C# SdpMessage class, to
pass the SDP messages around with additional type safety and without the
risks associated with string comparison (locale, misspelling, ...).

Bug: #188

### DIFF
--- a/examples/TestAppUwp/NodeDssSignaler.cs
+++ b/examples/TestAppUwp/NodeDssSignaler.cs
@@ -77,6 +77,26 @@ namespace TestAppUwp
                 throw new ArgumentException($"Unkown signaler message type '{stringType}'");
             }
 
+            public static WireMessageType TypeFromSdpMessageType(SdpMessageType type)
+            {
+                switch (type)
+                {
+                case SdpMessageType.Offer: return WireMessageType.Offer;
+                case SdpMessageType.Answer: return WireMessageType.Answer;
+                }
+                throw new ArgumentException($"Invalid SDP message type '{type}'.");
+            }
+
+            public static Message FromSdpMessage(SdpMessage message)
+            {
+                return new Message
+                {
+                    MessageType = TypeFromSdpMessageType(message.Type),
+                    Data = message.Content,
+                    IceDataSeparator = "|"
+                };
+            }
+
             /// <summary>
             /// The message type
             /// </summary>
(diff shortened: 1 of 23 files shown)


examples/TestAppUwp/NodeDssSignaler.cs
```diff
@@ -77,6 +77,26 @@ public static WireMessageType WireMessageTypeFromString(string stringType)
                 throw new ArgumentException($"Unkown signaler message type '{stringType}'");
             }
 
+            public static WireMessageType TypeFromSdpMessageType(SdpMessageType type)
+            {
+                switch (type)
+                {
+                case SdpMessageType.Offer: return WireMessageType.Offer;
+                case SdpMessageType.Answer: return WireMessageType.Answer;
+                }
```
fibann, line 86: nit:
```suggestion
                switch (type)
                {
                    case SdpMessageType.Offer: return WireMessageType.Offer;
                    case SdpMessageType.Answer: return WireMessageType.Answer;
                }
```

libs/mrwebrtc/src/peer_connection.cpp
```diff
@@ -665,29 +665,27 @@ ErrorOr<Transceiver*> PeerConnection::AddTransceiver(
...
 
-bool PeerConnection::SetRemoteDescriptionAsync(
-    const char* type,
+Error PeerConnection::SetRemoteDescriptionAsync(
+    mrsSdpMessageType type,
     const char* sdp,
     RemoteDescriptionAppliedCallback callback) noexcept {
   if (!peer_) {
-    return false;
+    return Error(mrsResult::kInvalidOperation);
```
fibann, line 673: nit: should be kInvalidParameter instead?

djee-ms: `peer_` is not a parameter to this method. Instead the method was called in an invalid state, so the operation is invalid. No?

examples/TestAppUwp/NodeDssSignaler.cs
djee-ms: Ignoring for now, there's another change coming that clean-up that.

libs/mrwebrtc/src/peer_connection.cpp
fibann: Right, disregard this.

PR #326, opened 2020-05-06
Maintain tools/generate-stubs

fibann:
Dependency of the desktop_capture module, deleting it prevents building win32.

### DIFF
--- a/tools/build/libwebrtc/mrwebrtc.ps1
+++ b/tools/build/libwebrtc/mrwebrtc.ps1
@@ -61,13 +61,19 @@ function Initialize-BuildEnvironment {
 
 # Post-checkout clean-up
 function Clear-PostCheckout([string]$SourceFolder) {
-    # Remove tools/ (except tools/clang/ and tools/protoc_wrapper/)
+    # Cleanup temp _tools folder if left over.
+    if (Test-Path "_tools") {
+        Write-Host "Removing temporary _tools directory"
+        Remove-Item -Path "_tools" -Recurse -Force
+    }
+    # Remove tools/* (except tools/clang/ and tools/protoc_wrapper/)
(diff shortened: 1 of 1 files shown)


tools/build/libwebrtc/mrwebrtc.ps1
```diff
@@ -61,13 +61,19 @@ function Initialize-BuildEnvironment {
 
 # Post-checkout clean-up
 function Clear-PostCheckout([string]$SourceFolder) {
-    # Remove tools/ (except tools/clang/ and tools/protoc_wrapper/)
+    # Cleanup temp _tools folder if left over.
+    if (Test-Path "_tools") {
+        Write-Host "Removing temporary _tools directory"
+        Remove-Item -Path "_tools" -Recurse -Force
```
djee-ms, line 67: `| Out-Null`? Otherwise it will list all files deleted 1 by 1 no?

djee-ms: Note also that according to [Example 4](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/remove-item?view=powershell-7#examples) in the doc of `Remove-Item`, we should in theory not use `-Recurse` but I couldn't find a good alternative (the one in the example doesn't work well either).

> Because the Recurse parameter in Remove-Item has a known issue, the command in this example uses Get-ChildItem to get the desired files, and then uses the pipeline operator to pass them to Remove-Item.

fibann: As said in chat `Remove-Item` shouldn't print anything by default. 

Logged the `-Recurse` issue as #327, will stick with the plain command here for consistency with the rest of the file (should be a rare, easy to diagnose issue anyway).

PR #334, opened 2020-05-11
Add documentation for Android Studio debugging

djee-ms:
Add some short documentation on how to setup Android Studio to do
source-level debugging when generating a project with Unity and IL2CPP.


tools/build/android/debugging.md
```diff
@@ -0,0 +1,155 @@
...
+- Same as above, open the **Debugger** tab of the **Run/Debug Configurations** dialog.
+- Change **Debug type** to:
+  - **Native Only** for debugging only C++.
+  - **Dual (Java + Native)** to debug both C++ and Java at the same time. This allows launching two separate debuggers in parallel when the application starts, the Java debugger and the C++ debugger (LLDB).
+
+![Configure the debugger type in Android Studio](docs/android-studio-debugger-type.png)
+
+## Ensure library is built with debug symbols
+
+On Android, and more generally on *nix platforms, the debug symbols are often embedded into the shared modules (`*.so`) instead of being in a separate file like on Windows (`*.pdb`). This is currently the case for `libmrwebrtc.so`.
```
fibann, line 42:
```suggestion
On Android, and more generally on \*nix platforms, the debug symbols are often embedded into the shared modules (`*.so`) instead of being in a separate file like on Windows (`*.pdb`). This is currently the case for `libmrwebrtc.so`.
```

djee-ms: Fixed.

PR #337, opened 2020-05-12
More Android debugging docs

djee-ms:
Add some more docs about Android debugging, especially caveats of source
remapping.


djee-ms:
> All this info is _really_ nice though it's not exclusive to MR-WebRTC. I wonder if it might be the case to collect all that refers to general debugging of Android+Unity+IL2CPP into some ad hoc repo/page and just link it from here.

That seems overkill, though I get the idea. Let's keep it light for now, this is more a memo for when someone needs a quick start or refresh.

PR #339, opened 2020-05-12
Split long log messages on Android for logcat

djee-ms:
Work around the built-in logcat limit on Android (about 1000 characters)
by splitting some hand-picked debug messages known to generally be
longer than that. This prevent those messages from being truncated in
logcat.


djee-ms:
Notes:
- The logcat limit could be retrieved dynamically I guess, don't want to bother here, that sounds complex from C#, and not sure it can be changed dynamically anyway so would not provide much benefit (I think it's hard-coded in the Android kernel).
- Unity normally appends the C# callstack, which therefore here gets completely truncated since we already use up to 1000 characters per call. We could do better, but it's unlikely to be worth the effort. Also the last bit of the message likely will have enough space left for that callstack.

PR #340, opened 2020-05-12
Downgrade mrwebrtc on Android to C++14

djee-ms:
This prevent binary incompatibility between `libwebrtc`, which is always
compiled as C++14 officially (except on UWP where some experimental C++17
support was added by WebRTC UWP SDK) and `mrwebrtc` which links statically
against it. In particular this avoids binary incompatible `std::optional` and
`absl::optional`, which cause hard-to-debug issues.

Note that in theory this change should also downgrade the `mrwebrtc-win32`
project, but this breaks compatibility with the NuGet prebuilt libwebrtc
v1.0.3, and therefore change on that project is delayed to prevent binary
incompatibility until C++14-compatible NuGet packages are available.

Bug: #329

### DIFF
--- a/libs/mrwebrtc/src/pch.h
+++ b/libs/mrwebrtc/src/pch.h
@@ -31,6 +31,10 @@
 #define WEBRTC_POSIX 1
 #define WEBRTC_ANDROID 1
 
+#else
+
+#error Unknown platform
+
 #endif
 
 #if (__cplusplus >= 201703L)
--- a/libs/mrwebrtc/src/pch.h
+++ b/libs/mrwebrtc/src/pch.h
@@ -8,7 +8,6 @@
 #include <functional>
 #include <mutex>
 #include <string>
-#include <string_view>
 #include <unordered_set>
 
 #if defined(MR_SHARING_WIN)
--- a/libs/mrwebrtc/src/pch.h
+++ b/libs/mrwebrtc/src/pch.h
@@ -37,12 +37,6 @@
 
 #endif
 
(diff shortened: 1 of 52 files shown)


libs/mrwebrtc/src/pch.h
```diff
@@ -34,6 +33,12 @@
 
 #endif
 
+#if (__cplusplus >= 201703L)
```
fibann, line 36: Make sure that we are using https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus

djee-ms:
> 
> 
> It would have been nicer to cherry-pick the commits from the undock branch, but I guess that if the changes are exactly the same that shouldn't matter too much.

I thought about it, but that doesn't change anything from git's point of view (cherry-pick is a brand new commit without history information about the relation with the source commit) and I was worried that the branches being fairly different the cherry-pick would result in too many conflicts or unintended changes getting dragged in the middle.

djee-ms:
Actually we can't merge that as is, this will break CI and users compiling from Core NuGet packages which have the `libwebrtc` from NuGet 1.0.3 built with C++17 even on Win32. So I'll revert the change for `mrwebrtc-win32.vcxproj` until we are closer to release and have some strategy for `libwerbtc` (either compile from source from undock, or new packages compiled with C++14).

libs/mrwebrtc/src/pch.h
djee-ms: Fixed.

PR #347, opened 2020-05-14
Add simple README to document libs/

djee-ms:


libs/README.md
```diff
@@ -0,0 +1,25 @@
+# Libraries
+
+This folder contains the MixedReality-WebRTC libraries for C/C++, C#, and Unity.
```
fibann, line 3: Hmm wouldn't one README per lib be better?

djee-ms: Be better, no. Be a good thing in addition, maybe. This is mostly to direct users to the right folder, and is displayed on GitHub when browsing to that folder.

fibann: It doesn't seem to me that it adds a lot to the main README, but can't hurt I guess!

PR #350, opened 2020-05-14
Add release pipeline for Unity

djee-ms:
Add Azure DevOps pipeline scripts to enable setting up an NPM package release pipeline for Unity. The pipeline allows building the `mrwebrtc` native library in Release config for all 5 supported platforms/architectures on Windows, as well as the C# library. It then packages those with the Unity library package and publish it alongside the samples package to a chosen NPM register.


tools/ci/templates/jobs-cpp.yaml
```diff
@@ -150,50 +151,62 @@ jobs:
...
-    displayName: 'NuGet restore tests'
-    inputs:
-      command: restore
-      restoreSolution: '$(Build.SourcesDirectory)/tools/build/mrwebrtc/win32/tests/packages.config'
-      restoreDirectory: '$(Build.SourcesDirectory)/packages'
-      includeNuGetOrg: true
-      feedsToUse: 'config'
-      nugetConfigPath: '$(NuGetConfigPath)'
-    timeoutInMinutes: 10
+  - ${{ if eq(parameters.withTesting, 'true') }}:
```
fibann, line 177: nit: move the testing steps into a template and include it from here, so you don't need to check `withTesting` 4 times

jonywalker1998:
The most recent build lost tranceiver in peerconnection, so that I am not able to add media send and receiver channel

tools/ci/templates/jobs-cpp.yaml
djee-ms: Let's leave as TODO, there's much cleaning to be done on pipelines.

PR #357, opened 2020-05-18
Point to WinRTC master branch

fibann:


.gitmodules
```diff
@@ -4,4 +4,3 @@
 [submodule "external/winrtc"]
 	path = external/winrtc
 	url = https://github.com/microsoft/winrtc.git
-	branch = user/LoadLibrary/Arm64
```
djee-ms, line 7: This is not enough to point to the master branch, there should also be a change in the `external/webrtc` "file" that shows the commit ID changing. This branch attribute here is _not_ where the submodule point (I forget what it's used for, but is optional).

fibann: Fixed.

PR #359, opened 2020-05-19
Uncomment and tidy up audio track read buffer

fibann:
The API has been refactored to create a buffer from a track (rather than the PeerConnection) and to address comments on #99.


libs/Microsoft.MixedReality.WebRTC/AudioTrackReadBuffer.cs
```diff
@@ -1,48 +1,111 @@
...
-        private AudioTrackReadBufferInterop.Handle _nativeHandle;
+        private readonly RemoteAudioTrackInterop.ReadBufferHandle _nativeHandle;
 
-        internal AudioTrackReadBuffer(PeerConnectionHandle peerHandle, int bufferMs)
+        /// <summary>
+        /// Controls the padding behavior of
+        /// <see cref="Read(int, int, float[], out int, out bool, PadBehavior)"/>
+        /// on underrun.
+        /// </summary>
+        public enum PadBehavior
```
djee-ms, line 26: Nice! 👍

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
```diff
@@ -37,6 +52,20 @@ internal class RemoteAudioTrackInterop
             EntryPoint = "mrsRemoteAudioTrackIsOutputToDevice")]
         public static extern mrsBool RemoteAudioTrack_IsOutputToDevice(IntPtr trackHandle);
 
+        [DllImport(Utils.dllPath, CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Ansi,
+            EntryPoint = "mrsRemoteAudioTrackCreateReadBuffer")]
+        public static extern uint CreateReadBuffer(IntPtr trackHandle, out ReadBufferHandle audioTrackReadBufferOut);
```
djee-ms, line 57: nit: follow existing convention `RemoteAudioTrack_CreateReadBuffer`, and below `ReadBuffer_Read` and `ReadBuffer_Destroy`.

libs/mrwebrtc/src/media/audio_track_read_buffer.cpp
```diff
@@ -145,49 +138,78 @@ void AudioTrackReadBuffer::Buffer::addFrame(const Frame& frame,
...
+      } else {
+        // no more input! fill with sin wave
+        constexpr float freq = 2 * 222 * float(M_PI);
+        switch (pad_behavior) {
+          case mrsAudioTrackReadBufferPadBehavior::kDoNotPad:
+            break;
+          case mrsAudioTrackReadBufferPadBehavior::kPadWithZero:
+            for (int i = 0; i < dst_len; ++i) {
+              dst[i] = 0;
+            }
```
djee-ms, line 192: `memset(dst, 0, dst_len * sizeof(float))` ?
I know it most likely compile to the same, but be lazy! ;)

libs/mrwebrtc/src/media/audio_track_read_buffer.h
```diff
@@ -56,12 +56,14 @@ class AudioTrackReadBuffer {
   };
   // Incoming frames received from webrtc - see also buffer_
   std::deque<Frame> frames_;
-  // protects frames_ in Read() and audioFrameCallback()
+  // protects frames_ and has_overrun_ in Read() and audioFrameCallback()
   std::mutex frames_mutex_;
   // max ms of audio data stored in frames_
-  int buffer_ms_ = 0;
+  int buffer_ms_{};
```
djee-ms, line 62: nit: `buffering_ms_` or `buffer_size_ms_`? Sounds like this could be a pointer to a buffer, whereas it's a buffering length.

libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
```diff
@@ -10,22 +10,15 @@ namespace Microsoft.MixedReality.WebRTC.Unity
     /// existing WebRTC peer connection by a remote peer and received locally.
     /// The audio track can optionally be displayed locally with a <see cref="MediaPlayer"/>.
     /// </summary>
+    /// <remarks>
```
djee-ms, line 13: I skipped this file; I am changing all of that for split sources so will fix when I merge my branch if needed.

libs/Microsoft.MixedReality.WebRTC/AudioTrackReadBuffer.cs
```diff
@@ -1,48 +1,111 @@
...
+
+        /// <summary>
+        /// Fill <paramref name="data"/> with samples from the internal buffer.
+        /// See <see cref="Read(int, int, float[], out int, out bool, PadBehavior)"/>.
+        /// </summary>
+        public void Read(int sampleRate, int channels,
+            float[] data, PadBehavior padBehavior = PadBehavior.PadWithZero)
         {
-            AudioTrackReadBufferInterop.Read(_nativeHandle, sampleRate, data, data.Length, channels);
+            Read(sampleRate, channels, data, out int _, out bool _, padBehavior);
```
stephenatwork, line 100: Nice! I've not seen that syntax

libs/mrwebrtc/include/remote_audio_track_interop.h
```diff
@@ -52,10 +59,90 @@ mrsRemoteAudioTrackIsEnabled(mrsRemoteAudioTrackHandle track_handle) noexcept;
...
+/// |sample_rate|: Desired sample rate. Data in the buffer is resampled if this
+/// is different from the native track rate.
+///
+/// |num_channels|: Desired number of channels. Should be 1 or 2. Data in the
+/// buffer is split/averaged if this is different from the native track channels
+/// number.
+///
+/// |pad_behavior|: Controls how |data| is padded in case of underrun.
+///
+/// |data_out|: Will be filled with the samples read from the internal buffer.
```
stephenatwork, line 126: Let's be explicit that the buffer is allocated by the user (obvious from the signature) and how big the buffer needs to be (less obvious)

stephenatwork: Forcing the caller to pass the size might help users from making mistakes, even though it's redundant. It means we can give better error messages too.

libs/mrwebrtc/include/remote_audio_track_interop.h
```diff
@@ -52,10 +59,90 @@ mrsRemoteAudioTrackIsEnabled(mrsRemoteAudioTrackHandle track_handle) noexcept;
...
+///
+/// |num_read_samples_out|: Set to the effective number
```
stephenatwork, line 128: Pointer arguments 'at the end' are often optional (null allowed). Let's be explicit if not.

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
```diff
@@ -1,72 +1,123 @@
...
+                            mrsAudioTrackReadBufferPadBehavior pad_behavior,
+                            float* data_out,
+                            int* data_len_in_out,
+                            mrsBool* has_overrun_out) {
+  if (!sample_rate || !num_channels ||
+      !IsValidAudioTrackBufferPadBehavior(pad_behavior) || !data_len_in_out ||
+      *data_len_in_out < 0 || (*data_len_in_out && !data_out) ||
+      !has_overrun_out)
+  {
+    return Result::kInvalidParameter;
```
stephenatwork, line 105: LOL, best of luck to the caller if they get this and don't have the PDBs! Even artificially splitting the cases and codes might help narrow down the cause.

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
```diff
@@ -1,72 +1,123 @@
...
+  return Result::kInvalidNativeHandle;
+}
+
+mrsResult MRS_CALL
+mrsAudioTrackReadBufferRead(mrsAudioTrackReadBufferHandle buffer,
+                            int sample_rate,
+                            int num_channels,
+                            mrsAudioTrackReadBufferPadBehavior pad_behavior,
+                            float* data_out,
+                            int* data_len_in_out,
```
stephenatwork, line 98: This is out only in the header. Even though longer, perhaps separate in & out size params are more idiomatic?

djee-ms: We use in+out parameters in a few places already, so it's more consistent like this.

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
djee-ms: +1
I tend to write `RTC_LOG(LS_ERROR) << "..."` now for each parameter so users can self-diagnose.

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
stephenatwork: The header and docstring is inconsistent.

```
mrsAudioTrackReadBufferRead(mrsAudioTrackReadBufferHandle buffer,
                            ...
                            float* data_out,
                            int* num_read_samples_out,
```

stephenatwork: BTW I agree inout combining is OK here since the API is internal and the param list is long.

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
fibann: The prefix is a bit redundant IMHO (RemoteAudioTrackInterop.RemoteAudioTrack_*) but I'll keep for consistency, can be changed later.

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
fibann: Yes it was meant to be `num_read_samples_in_out` and the docs are incomplete - I copied the C# comments where the buffer size is passed in the array, and forgot to update 😄 

This is public API technically, and I agree separate in and out make sense (they mean different things) so will separate.

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
fibann: Fixed

libs/mrwebrtc/src/media/audio_track_read_buffer.cpp
fibann: Fixed

libs/mrwebrtc/src/media/audio_track_read_buffer.h
fibann: Fixed

libs/mrwebrtc/include/remote_audio_track_interop.h
fibann: Fixed

libs/mrwebrtc/include/remote_audio_track_interop.h
fibann: Fixed

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
fibann: Fixed

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
fibann: Fixed

libs/mrwebrtc/src/interop/remote_audio_track_interop.cpp
djee-ms: I hope `#__VA_ARGS__` works with `clang` on Android...

fibann: Godbolt says it's fine on clang

stephenatwork:
🍾

PR #361, opened 2020-05-20
Add mrsOptBool

djee-ms:
Add a new interop type mrsOptBool conceptually equivalent to
std::optional<bool>, with 1-byte size.


djee-ms:
This will be used for the various audio options, most of which use `std::optional<bool>` internally. I _think_ the optional part means that unset is for "use implementation default".

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
```diff
@@ -23,6 +23,72 @@ internal struct mrsBool
...
+        }
+        public static explicit operator mrsOptBool(bool b) { return (b ? True : False); }
+        public static explicit operator mrsOptBool(bool? b)
+        {
+            if (b.HasValue)
+            {
+                return (b.Value ? True : False);
+            }
+            return Unset;
+        }
```
fibann, line 73: What's the rationale for having both ctor and conversion operator? Can we have just the ctor (or just the operator, if we want to avoid `new` calls)?

djee-ms: No rationale, I copied what was for `mrsBool`. I guess it makes sense to have only the cast operators. I removed the constructors.

Fixed.

djee-ms: Actually note there's a hidden effect : the compiler will now use the implicit zeroing constructor so the default is implicitly false. With explicit constructors, well, things are more explicit 😄 

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-constructors
> structs cannot contain an explicit parameterless constructor because one is provided automatically by the compiler. This constructor initializes each field in the struct to the default value.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values
> Any built-in integral numeric type : 0 (zero)

fibann: nit: I guess we could add a default ctor that initializes to unassigned, and/or making it private so it's not used externally. But not really needed.

Issue #244, opened 2020-03-30
Update the C# tutorials to use the new multi-track API


djee-ms:
This is a TODO task to update the C# tutorials (.NET Core 3 and UWP) to use the new multi-track API. The Unity tutorial has already been updated.

AtosNicoS:
Is the example for UWP up to date? Even if there is no tutorial, can I safely use the example as orientation?

Is it to be expected that the documentation will be adapted promptly?

djee-ms:
I am not sure I understand any of those questions. Isn't that just answered above? If not can you please detail what you want to know?

>Is the example for UWP up to date? 
>> This is a TODO task to update the C# tutorials (.NET Core 3 and **UWP**) 

So, no.

>Even if there is no tutorial,
>> to update the C# **tutorials**

What examples do you want to use?

>Is it to be expected that the documentation will be adapted promptly?
>>This is a **TODO task** 

Yes, that is why we logged that task. This is a mandatory TODO before we can release.

AtosNicoS:
I am using version 1.0.3 of the library, and refer to the examples in https://github.com/microsoft/MixedReality-WebRTC/tree/release/1.0/examples/TestAppUwp

So for 1.0.3 the currently online available tutorial is correct? I am a bit confused now, as it seems the website is already using docs for 2.x but there is no release yet? I am sorry, that this is not clear to me, I am new to this library.

djee-ms:
Yes, the docs for `release/1.0` are correct for the NuGet packages 1.0.x, and are [here](https://microsoft.github.io/MixedReality-WebRTC/versions/release/1.0/). And the TestAppUWP example as you linked it is also correct.

The docs for `master` on the other hand (see [here](https://microsoft.github.io/MixedReality-WebRTC/)) use the new API present on `master` which will be released later as NuGet packages 2.0 once we are ready with that milestone (see [roadmap here](https://github.com/microsoft/MixedReality-WebRTC/issues/142)).

You have a drop-down control to select the version of the docs you want at the top right of the documentation webpage on github.io. Docs for 1.0 are titles as such, and docs for master don't have a version associated with them. The default is for `master` because the default branch is `master`.

![image](https://user-images.githubusercontent.com/49064175/80598491-57cfa000-8a21-11ea-91da-dcec7bcd3d30.png)

![image](https://user-images.githubusercontent.com/49064175/80598496-59996380-8a21-11ea-8d0a-2b5900ed5234.png)

[...]

PR #369, opened 2020-05-20
Update the .NET Core 3.0 tutorial

djee-ms:
Update the tutorial for the .NET Core 3.0 project to reflect latest
changes already present in the repository.

Bug: #244


djee-ms:
The tutorial was actually already using the 2.0 API and was fairly up-to-date; just minor changes and fixing a missing `Dispose()` in the example console app.

PR #373, opened 2020-05-22
Add unified release pipeline for NuGet packages (Windows)

djee-ms:
Add a unified release pipeline for producing NuGet packages for Windows release. In a first stage, the pipeline compiles `libmrwebrtc` from sources, then all build variants (platform/arch/config) of the native library `mrwebrtc`, as well as the C# library. Once building is successful, a second stage signs all DLLs and generate NuGet packages, then sign them. A third stage then run some tests on those final NuGet packages.


djee-ms:
Notes:
- Testing should be expanded to ensure NuGet packages are well-formed and useable, but this can be incrementally added on top of this change.
-  Currently there is no automated publishing to nuget.org because ADO doesn't support gates in multi-stage build pipelines (manual confirmation from maintainer before publishing). So publishing should be done manually instead by downloading the pipeline artifacts (4 packages) and uploading to nuget.org.

tools/ci/templates/jobs-cslib-release-build.yaml
```diff
@@ -0,0 +1,72 @@
...
+  #       workingDirectory: 'tests/Microsoft.MixedReality.WebRTC.Tests/'
+
+  # # Run tests (Win32-x64-buildConfig)
+  # - ${{ if eq(parameters.withTesting, 'true') }}:
+  #   - task: VSTest@2
+  #     displayName: 'C# Testing (mrwebrtc Win32-x64-${{parameters.buildConfig}} / C# ${{parameters.buildConfig}})'
+  #     inputs:
+  #       testAssemblyVer2: 'Microsoft.MixedReality.WebRTC.Tests.dll'
+  #       searchFolder: 'bin/netstandard2.0/${{parameters.buildConfig}}'
+  #       runOnlyImpactedTests: false
```
fibann, line 57: Delete?

djee-ms: Yes that needs to move to the testing stage. I left it for easy copy/paste.

tools/ci/templates/jobs-mrwebrtc-release-build.yaml
```diff
@@ -0,0 +1,190 @@
+# Copyright (c) Microsoft Corporation.
```
fibann, line 1: What's the difference between jobs-cpp.yaml and this file?

fibann: Looking at them side by side it seems that jobs-cpp uses the NuGets for webrtc-uwp, while this builds from source. Is there a reason to have these as separate jobs sequences, or can we merge them (or at least extract common steps in templates)

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.Native.Desktop.props
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemDefinitionGroup>
+        <ClCompile>
+            <LanguageStandard>stdcpp14</LanguageStandard>
```
fibann, line 5: What does this do?

tools/ci/templates/jobs-mrwebrtc-release-build.yaml
djee-ms: `jobs-cpp` builds for CI. We should probably try to merge some parts where possible, but I wanted to get that PR out so at least we have something working, and we can improve after.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.Native.Desktop.props
djee-ms: Taken offline: good catch, not needed since we now expose a pure C API so are not relying on any C++ standard feature.

djee-ms: Fixed.

PR #374, opened 2020-05-23
Moving the C# DLL to /Assets/Plugins/

sundarmuru:
The initial placeholder for C# Managed DLL under 'Assets/Plugins/Win32/x86_64' failed on running as an ARM app on Hololens 2. The Managed DLL was not copied over. Moving to a folder higher in the hierarchy solved this problem.


djee-ms:
I have the opposite experience, which is that putting the managed assembly anywhere else upper in the hierarchy than where Unity already found another DLL (in x64 case, `mrwebrtc.dll`) will make Unity fail to load that DLL. The mechanism for DLL discovery is undocumented and very brittle. Unfortunately I won't be able to take that change unless we do some extensive testing, and as of today I am not sure we have time for that. Note also that you mention the issue is about copying not loading, and we never experienced any copying issue on HoloLens 2. So please open an issue if there is a problem for you and we can diagnose it. If it appears there is indeed an issue then we can fix it, but we have many users using HoloLens 2 and we do most of our testing on it, and this is the first time I hear about copy issues on ARM.

PR #377, opened 2020-05-25
Add split audio track source

djee-ms:
Split the audio track source from the audio track, to allow sharing
across peer connections.

This change splits the former local audio track into 2 separate
components:
- The audio track source, which generates the raw audio frames,
  generally from a local audio capture device (microphone) but not
  necessarily. At this time only device-based sources are supported
  though. The audio track source is not associated with any peer
  connection in particular, and can be shared amongst any number of
  local audio tracks, from the same peer connection or not.
- The local audio track, which is associated with a particular peer
  connection, and which is a slim bridge between an audio track source
  and the audio transceiver of the peer connection.

This change adds support to the native and C# libraries only.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/AudioTrackSource.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/AudioTrackSource.cs
@@ -1,0 +1,176 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Microsoft.MixedReality.WebRTC.Interop;
+
+namespace Microsoft.MixedReality.WebRTC
+{
+    /// <summary>
+    /// Configuration to initialize capture on a local audio device (microphone).
+    /// </summary>
+    public class LocalAudioDeviceInitConfig
+    {
+        /// <summary>
+        /// Enable automated gain control (AGC) on the audio device capture pipeline.
+        /// </summary>
+        public bool? AutoGainControl = null;
+    }
+
+    /// <summary>
+    /// Audio source for WebRTC audio tracks.
+    /// 
+    /// The audio source is not bound to any peer connection, and can therefore be shared by multiple audio
+    /// tracks from different peer connections. This is especially useful to share local audio capture devices
+    /// (microphones) amongst multiple peer connections when building a multi-peer experience with a mesh topology
+    /// (one connection per pair of peers).
+    /// 
+    /// The user owns the audio track source, and is in charge of keeping it alive until after all tracks using it
+    /// are destroyed, and then dispose of it. The behavior of disposing of the track source while a track is still
+    /// using it is undefined. The <see cref="Tracks"/> property contains the list of tracks currently using the
+    /// source.
+    /// </summary>
+    /// <seealso cref="LocalAudioTrack"/>
+    public class AudioTrackSource : IDisposable
+    {
+        /// <summary>
+        /// A name for the audio track source, used for logging and debugging.
+        /// </summary>
+        public string Name
+        {
+            get
+            {
+                // Note: the name cannot change internally, so no need to query the native layer.
+                // This avoids a round-trip to native and some string encoding conversion.
+                return _name;
+            }
+            set
+            {
+                AudioTrackSourceInterop.AudioTrackSource_SetName(_nativeHandle, value);
+                _name = value;
+            }
+        }
+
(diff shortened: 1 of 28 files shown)


libs/Microsoft.MixedReality.WebRTC/AudioTrackSource.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                AudioTrackSourceInterop.AudioTrackSource_SetName(_nativeHandle, value);
+                _name = value;
+            }
+        }
+
+        /// <summary>
+        /// List of local audio tracks this source is providing raw audio frames to.
+        /// </summary>
+        public List<LocalAudioTrack> Tracks { get; private set; } = new List<LocalAudioTrack>();
```
fibann, line 60: This should be an `IEnumerable` property (or users might add stuff to the list) and possibly contain an instance of some read-only enumerable internal class (or users might cast the property value to a List and still wreak havoc - a bit defensive but the only way to be sure).

djee-ms: Seems `IReadOnlyList<T>` is even better as it exposes `[]` and `Count`, which are always handy to have for the user.

djee-ms: Fixed.

libs/mrwebrtc/include/audio_track_source_interop.h
```diff
@@ -0,0 +1,76 @@
...
+/// Register a custom callback to be called when the audio track source produced
+/// a frame.
+/// WARNING: The default platform source internal implementation currently does
+/// not hook those callbacks, and therefore the callback will never be called.
+/// This is a limitation of the underlying implementation.
+/// See https://bugs.chromium.org/p/webrtc/issues/detail?id=11602
+MRS_API void MRS_CALL mrsAudioTrackSourceRegisterFrameCallback(
+    mrsAudioTrackSourceHandle source_handle,
+    mrsAudioFrameCallback callback,
+    void* user_data) noexcept;
```
fibann, line 74: If this is never called, can we leave it out/commented? It's public API and we'd better not have people taking a dependency on it, if there's the risk that we'll have to write it differently when/if it actually works!

djee-ms: I can comment it out.

djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/LocalAudioTrack.cs
```diff
@@ -77,63 +82,70 @@ public bool Enabled
...
-        /// <remarks>
-        /// On UWP this requires the "microphone" capability.
-        /// See <see href="https://docs.microsoft.com/en-us/windows/uwp/packaging/app-capability-declarations"/>
-        /// for more details.
-        /// </remarks>
-        public static Task<LocalAudioTrack> CreateFromDeviceAsync(LocalAudioTrackSettings settings = null)
+        /// <param name="source">The track source which provides the raw audio frames to the newly created track.</param>
+        /// <param name="initConfig">Configuration to initialize the track being created.</param>
+        /// <returns>Asynchronous task completed once the track is created.</returns>
+        public static Task<LocalAudioTrack> CreateFromSourceAsync(AudioTrackSource source, LocalAudioTrackInitConfig initConfig)
```
fibann, line 93: Does this still need to be async?

libs/mrwebrtc/src/media/audio_track_source.h
```diff
@@ -0,0 +1,116 @@
...
+  SourceState state() const override { return state_; }
+  bool remote() const override { return false; }
+
+  //
+  // AudioSourceInterface
+  //
+
+  // Sets the volume of the source. |volume| is in  the range of [0, 10].
+  // TODO(tommi): This method should be on the track and ideally volume should
+  // be applied in the track in a way that does not affect clones of the track.
```
fibann, line 45: Remove TODO (also below)

libs/Microsoft.MixedReality.WebRTC/LocalAudioTrack.cs
djee-ms: No; that's one of the next reviews after this one :)

libs/mrwebrtc/src/media/audio_track_source.h
djee-ms: Fixed.

PR #379, opened 2020-05-25
Add a glossary to the documentation

djee-ms:
Add a new documentation page containing a glossary of common WebRTC
concepts explained.

### DIFF
--- a/docs/manual/glossary.md
+++ b/docs/manual/glossary.md
@@ -25,7 +25,7 @@ A local media track is a [_media track_](#media-track) whose source is local to 
 
 ### Media
 
-The term _media_ refers to either or both of _audio_ and _video_. It does **not** encapsulate [_data channels_](#data-channel).
+The term _media_ refers to either or both of _audio_ and _video_. It does **not** include [_data channels_](#data-channel).
 
 ### Media track
 
--- a/docs/manual/glossary.md
+++ b/docs/manual/glossary.md
@@ -1,0 +1,78 @@
+---
+uid: glossary
+---
+
+# Glossary
+
+## D
+
+### Data channel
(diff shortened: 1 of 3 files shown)


djee-ms:
This makes it easier to reference the explanations from the various tutorials rather than repeat them in each tutorial.

image

docs/manual/glossary.md
```diff
@@ -0,0 +1,78 @@
+---
+uid: glossary
+---
+
+# Glossary
+
+## D
```
fibann, line 7: nit: do we really need letter sections? There are < 10 entries and just one for (almost) every letter.

docs/manual/glossary.md
```diff
@@ -0,0 +1,78 @@
...
+> See also
+>
+> - [_media track_](#media-track)
+> - [_remote media track_](#remote-media-track)
+
+## M
+
+### Media
+
+The term _media_ refers to either or both of _audio_ and _video_. It does **not** encapsulate [_data channels_](#data-channel).
```
fibann, line 28:
```suggestion
The term _media_ refers to either or both of _audio_ and _video_. It does **not** include [_data channels_](#data-channel).
```
Note sure "encapsulate" is correct here.

docs/manual/glossary.md
djee-ms: I find readability much better. This also prevents the right-side menu from being one long flat list. Glossary generally have per-letter grouping/section too. I'm also hopeful we can add more definitions as needed.

docs/manual/glossary.md
djee-ms: Fixed.

PR #380, opened 2020-05-26
Add split video track source

djee-ms:
Split the video track source from the video track, to allow sharing
across peer connections.

This change splits the former local video track into 2 separate
components:
- The video track source, which generates the raw video frames,
  generally from a local vdeo capture device (webcam) but not
  necessarily. The video track source is not associated with any peer
  connection in particular, and can be shared amongst any number of
  local video tracks, from the same peer connection or not.
- The local video track, which is associated with a particular peer
  connection, and which is a slim bridge between an video track source
  and the video transceiver of the peer connection.

This change does not touch yet the external video track, which should
eventually be converted to a video track source instead, as this is a
more natural fit.


djee-ms:
Second part for video, leaving out the change of external video source from a track to a proper source itself. This will be followed up by Unity changes too.

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
```diff
@@ -0,0 +1,361 @@
...
+
+        /// <summary>
+        /// Backing field for <see cref="Tracks"/>.
+        /// </summary>
+        private List<LocalVideoTrack> _tracks = new List<LocalVideoTrack>();
+
+        private event I420AVideoFrameDelegate _videoFrameReady;
+
+        /// <summary>
+        /// Create an video track source using a local video capture device (webcam).
```
fibann, line 182:
```suggestion
        /// Create a video track source using a local video capture device (webcam).
```

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
```diff
@@ -0,0 +1,361 @@
...
+        /// 
+        /// The video track source produces raw video frames by capturing them from a capture device accessible
+        /// from the local host machine, generally a USB webcam or built-in device camera. The video source
+        /// initially starts in the capturing state, and will remain live for as long as the source is alive.
+        /// Once the source is not live anymore (ended), it cannot be restarted. A new source must be created to
+        /// use the same video capture device again.
+        /// 
+        /// The source can used to create one or more local video tracks (<see cref="LocalVideoTrack"/>), which
```
fibann, line 190:
```suggestion
        /// The source can be used to create one or more local video tracks (<see cref="LocalVideoTrack"/>), which
```

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
```diff
@@ -0,0 +1,361 @@
...
+        /// once added to a video transceiver allow the video frames to be sent to a remote peer. The source itself
+        /// is not associated with any peer connection, and can be used to create local video tracks from multiple
+        /// peer connections at once, thereby being shared amongst those peer connections.
+        /// 
+        /// The source is owned by the user, who muse ensure it stays alive while being in use by at least one local
```
fibann, line 195:
```suggestion
        /// The source is owned by the user, who must ensure it stays alive while being in use by at least one local
```

libs/mrwebrtc/src/media/video_track_source.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include "api/mediastreaminterface.h"
+
+namespace Microsoft {
+namespace MixedReality {
+namespace WebRTC {
+
+class VideoTrackSource;
+
+/// Adapter for a local audio source backing one or more local audio tracks.
```
fibann, line 20:
```suggestion
/// Adapter for a local video source backing one or more local video tracks.
```

libs/mrwebrtc/src/media/video_track_source.h
```diff
@@ -0,0 +1,122 @@
...
+
+  //
+  // VideoTrackSourceInterface
+  //
+
+  // Indicates that parameters suitable for screencasts should be automatically
+  // applied to RtpSenders.
+  // TODO(perkj): Remove these once all known applications have moved to
+  // explicitly setting suitable parameters for screencasts and don't need this
+  // implicit behavior.
```
fibann, line 48: Delete (also below)

libs/mrwebrtc/test/video_track_source_tests.cpp
```diff
@@ -0,0 +1,114 @@
...
+  mrsLocalVideoDeviceInitConfig config{};
+  mrsVideoTrackSourceHandle source_handle{};
+  ASSERT_EQ(mrsResult::kSuccess,
+            mrsVideoTrackSourceCreateFromDevice(&config, &source_handle));
+  ASSERT_NE(nullptr, source_handle);
+  mrsVideoTrackSourceRemoveRef(source_handle);
+}
+
+// TODO - Don't use device, run outside MRSW_EXCLUDE_DEVICE_TESTS
+TEST_P(VideoTrackSourceTests, Name) {
```
fibann, line 37: nit: 80 lines of test for every function that returns a string looks a bit silly, especially considering that the tested code is essentially the same. It would be nicer to have a `TestFunctionReturningString` helper.

libs/mrwebrtc/src/media/video_track_source.h
djee-ms: Fixed.

libs/mrwebrtc/src/media/video_track_source.h
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
djee-ms: Fixed.

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
djee-ms: Fixed.

Issue #388, opened 2020-05-31
MediaPlayer renders video flipped vertically on OpenGL-like platforms


djee-ms:
**Describe the bug**
When using OpenGL-like graphics APIs (`UNITY_UV_STARTS_AT_TOP=0`), the `MediaPlayer` component of the Unity library renders video frames flipped vertically.

**To Reproduce**
1. Use a Standalone build (Windows Desktop) with forced OpenGLES3 renderer, or deploy to an Android device which by default uses OpenGLES 2/3.

**Expected behavior**
The video is displayed with the correct orientation.

**Environment**
 - Platform: Any platform using an OpenGL-like graphics API where `UNITY_UV_STARTS_AT_TOP=0`
 - Architecture: any
 - Unity version (if applicable): any
 - Target device: any

**Additional context**
After some testing it appears that [`Texture2D.LoadRawTextureData()`](https://docs.unity3d.com/ScriptReference/Texture2D.LoadRawTextureData.html) always expects the image data in "the default convention which is OpenGL" (quote from some other part of the docs I cannot recall), that is bottom-up with first byte in memory being the bottom-left corner. Conversely, WebRTC video frames are always top-down, with the first byte corresponding to the top-left corner. Therefore the `MediaPlayer` component _always_ uploads the texture flipped vertically, whatever the graphics API used. However the YUV and ARGB feed shaders provided flips that texture again on non-OpenGL platforms (`UNITY_UV_STARTS_AT_TOP=1`), compensating for the bug. On OpenGL-like platforms however this compensation is not active and the video renders vertically flipped.

PR #389, opened 2020-05-31
Fix vertical flip of video on OpenGL platforms

djee-ms:
Fix video being rendered vertically flipped by the MediaPlayer component
when using the YUVFeedShader(Unlit) or ARGBFeedShader(Unlit) on
platforms using the OpenGL graphics API.

The Unity Texture2D.LoadRawTextureData() method used by the MediaPlayer
component always expects bottom-up image data, even when using Direct3D,
but the WebRTC video frames are always top-down. Instead of using a
costly CPU transform, simply upload the video frames inverted into the
YUV planar textures or the ARGB texture, and use the shader to flip the
texture back before rendering.

Bug: #388

### DIFF
--- a/libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
@@ -179,15 +179,15 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             var videoSrc = (IVideoSource)VideoSource;
             switch (videoSrc.FrameEncoding)
             {
-                case VideoEncoding.I420A:
-                    _i420aFrameQueue = new VideoFrameQueue<I420AVideoFrameStorage>(frameQueueSize);
-                    videoSrc.RegisterCallback(I420AVideoFrameReady);
-                    break;
+            case VideoEncoding.I420A:
+                _i420aFrameQueue = new VideoFrameQueue<I420AVideoFrameStorage>(frameQueueSize);
+                videoSrc.RegisterCallback(I420AVideoFrameReady);
+                break;
 
-                case VideoEncoding.Argb32:
-                    _argb32FrameQueue = new VideoFrameQueue<Argb32VideoFrameStorage>(frameQueueSize);
-                    videoSrc.RegisterCallback(Argb32VideoFrameReady);
-                    break;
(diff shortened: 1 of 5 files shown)


djee-ms:
Tested on Standalone (Windows Desktop) with Direct3D11 and OpenGLES3, and on Android device with OpenGLES3.

libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
```diff
@@ -330,7 +330,10 @@ private void TryProcessI420AFrame()
 
                 // Copy data from C# buffer into system memory managed by Unity.
                 // Note: This only "looks right" in Unity because we apply the 
-                // "YUVFeedShader" to the texture (converting YUV planar to RGB).
+                // "YUVFeedShader(Unlit)" to the texture (converting YUV planar to RGB).
+                // Note: Texture2D.LoadRawTextureData() expects some bottom-up texture data but
+                // the WebRTC vdeo frame is top-down, so the image is uploaded vertically flipped,
```
fibann, line 335:
```suggestion
                // the WebRTC video frame is top-down, so the image is uploaded vertically flipped,
```

libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
```diff
@@ -378,6 +381,9 @@ private void TryProcessArgb32Frame()
                 }
 
                 // Copy data from C# buffer into system memory managed by Unity.
+                // Note: Texture2D.LoadRawTextureData() expects some bottom-up texture data but
+                // the WebRTC vdeo frame is top-down, so the image is uploaded vertically flipped,
```
fibann, line 385:
```suggestion
                // the WebRTC video frame is top-down, so the image is uploaded vertically flipped,
```

libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
djee-ms: Fixed.

libs/unity/library/Runtime/Scripts/Media/MediaPlayer.cs
djee-ms: Fixed.

Issue #248, opened 2020-03-31
Merge the two Android archives into a single one


djee-ms:
This is a TODO task once #193 is merged to `master` to merge the 2 archives produced by the Android build (`libwebrtc.aar` and `mrwebrtc.aar`) into a single archive `mrwebrtc.aar`, for convenience purpose and to benefit from static linking and code stripping (see also #247).

djee-ms:
Fixed by #390.

Issue #247, opened 2020-03-31
Reduce Android archives size


djee-ms:
This is a TODO task to investigate and reduce the size of the 2 Android archives (`libwebrtc.aar` and `mrwebrtc.aar`) that the Android build produce once #193 is merged to `master`.

djee-ms:
Fixed by merging `libwebrtc.aar` and `mrwebrtc.aar`. There is little chance to find any other significant size reduction, so closing.

PR #390, opened 2020-05-31
Merge Android archives

djee-ms:
Merge the two Android archives (`*.aar`) previously needed for Android support in Unity into a single archive `mrwebrtc.aar`. This drastically reduces the storage footprint of a Unity Android project by shipping a single archive with a single version of `libwebrtc`.

Bugs: #247, #248


djee-ms:
This is missing a way to import `UnityUtility.java` which is not compiled in `libwebrtc.aar` by default. I had a local change, need to find a way around it by _e.g._ having our own utility class.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
```diff
@@ -126,6 +126,12 @@ public void Unmute()
         /// <inheritdoc/>
         protected async Task OnEnable()
         {
+#if PLATFORM_ANDROID
```
fibann, line 129: If we #if out the whole `Android` class we then have to #if out all the places of call. We should #if out the method bodies instead, and possibly rename `Android` to `Platform` or something like this.

tools/build/android/webrtc-native/build.gradle
```diff
@@ -66,18 +66,27 @@ dependencies {
...
         task.finalizedBy copyToUnitySample
-    } else if (task.name == 'externalNativeBuildDebug') {
+        copyToUnitySample.dependsOn(mergeLibwebrtcArchive)
+    } else if (task.name == 'bundleDebugAar') {
         task.finalizedBy copyToUnitySample
-    } else if (task.name == 'externalNativeBuildFulldebug') {
+        copyToUnitySample.dependsOn(mergeLibwebrtcArchive)
+    } else if (task.name == 'bundleFulldebugAar') {
         task.finalizedBy copyToUnitySample
+        copyToUnitySample.dependsOn(mergeLibwebrtcArchive)
```
fibann, line 90: I don't know the gradle syntax, but it seems to me that the body of all 3 branches is the same 😄 Merge?

djee-ms: Yeah I was lazy, didn't want to figure out Gradle. I guess some `task.name.beginsWith("externalNativeBuild")` would be best, if that's possible. I'll have a look at the Gradle docs.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
djee-ms: I appreciate the first part, which is what Unity does I think; I guess you can `using Unity.Android` on Windows and it does nothing. On the second part I disagree because this is really Android-specific.

tools/build/android/webrtc-native/build.gradle
djee-ms: Hum.. actually I was looking at the old code. I will just add a single `if`.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
djee-ms: Fixed.

tools/build/android/webrtc-native/build.gradle
djee-ms: Fixed.

djee-ms:
Fixed with rebase to keep separate commits for individual works. Changes since last iteration are:
- Moved `#if UNITY_ANDROID` inside `Android.Initialize()` and removed it from other places.
- Fixed the tasks in Gradle (last 3 blocks in `build.gradle`)

tools/build/android/webrtc-native/src/main/cpp/jni_onload.cpp
```diff
@@ -2,25 +2,91 @@
...
+  // completely different JVM/JNI set of C++ references
+  // (modules/utility/source/jvm_android.cc), only used by the audio module.
+  // Apparently it seems this was to become the "new way" but got cancelled and
+  // the change partially reverted.
+  // https://bugs.chromium.org/p/webrtc/issues/detail?id=8067
+  // Now, in addition, when intializing WebRTC from Java/Unity, this is actually
+  // automatically called from
+  // JNI_PeerConnectionFactory_InitializeAndroidGlobals() which is called from
+  // Java via the Unity wrapper code in Android.Initialize(), so cannot be
+  // called manually here otherwise it asserts.
```
fibann, line 57: Very nice!

PR #392, opened 2020-06-02
Remove abstraction of ExternalVideoTrackSource

djee-ms:
Remove the internal abstraction of the ExternalVideoTrackSource class,
which used to exhibit a clean C++ interface but has no reason to be now
that only a pure C interop API is exported. This simplifies the internal
implementation by removing a level of indirection.


libs/mrwebrtc/src/media/external_video_track_source.h
```diff
@@ -15,6 +15,47 @@ namespace WebRTC {
 
 class ExternalVideoTrackSource;
 
+namespace detail {
+
+/// Adapater for the frame buffer of an external video track source,
```
fibann, line 20:
```suggestion
/// Adapter for the frame buffer of an external video track source,
```

djee-ms: I like "Adapater" though... ;)

PR #395, opened 2020-06-03
Polymorphic API for unified object base

djee-ms:
Introduce some polymorphism into the API by having all objects derive
from a base class which contains some common functionalities, and have a
single set of API functions to manipulate those. This prevents
duplicating code to e.g. set an object name into each single type of API
object.

- mrsObject is the base class for all objects, and contains the name
  and user data storage and manipulation API.
- mrsRefCountedObject is the base class for the subset of objects which
  are reference-counted through the API, which are objects the user owns
  as opposed to objects owned by another object.


libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
```diff
@@ -2,18 +2,22 @@
...
-using System.Diagnostics;
 using System.Runtime.InteropServices;
 using static Microsoft.MixedReality.WebRTC.Interop.AudioTrackInterop;
 
 namespace Microsoft.MixedReality.WebRTC.Interop
 {
     internal class RemoteAudioTrackInterop
     {
-        internal class ReadBufferHandle : SafeHandle
+        internal sealed class RemoteAudioTrackHandle : ObjectHandle
```
fibann, line 12: nit: not strictly related, but we have a lot of `MyLongClassNameInterop.MyLongClassName_DoThis` and here even `MyLongClassNameInterop.MyLongClassNameHandle`. It would be nice to have the class name just once in the qualified name.

libs/mrwebrtc/include/interop_api.h
```diff
@@ -80,38 +80,41 @@ struct mrsRemoteVideoTrackConfig;
...
-using mrsRemoteVideoTrackHandle = void*;
+using mrsRemoteVideoTrackHandle = mrsObjectHandle;
 
 /// Opaque handle to a native DataChannel interop object.
-using mrsDataChannelHandle = void*;
+using mrsDataChannelHandle = mrsObjectHandle;
 
 /// Opaque handle to a native ExternalVideoTrackSource interop object.
-using mrsExternalVideoTrackSourceHandle = void*;
+using mrsExternalVideoTrackSourceHandle = mrsRefCountedObjectHandle;
```
fibann, line 117: nit: not incredibly useful since they are all `void*` 😄 

A nicer trick would be:
```cpp
#define DEFINE_HANDLE(NAME) \
  struct NAME ## Impl {}; \
  using NAME ## Handle = const NAME ## Impl*

#define DEFINE_HANDLE_DERIVED(NAME, PARENT) \
  struct NAME ## Impl : public PARENT ## Impl{}; \
  using NAME ## Handle = const NAME ## Impl*

DEFINE_HANDLE(mrsObject);
DEFINE_HANDLE_DERIVED(mrsRefCountedObject, mrsObject);
...
```

C# would still pass raw void pointers in a completely unsafe way, but at least any C++ client would be type safe.

libs/mrwebrtc/src/pch.h
```diff
@@ -93,3 +93,32 @@
...
+    RTC_DCHECK(obj->GetObjectType() == ObjectType::k##TYPE);             \
+    const std::string name = obj->GetName();                             \
+    if (obj->RemoveRef() == 0) {                                         \
+      RTC_LOG(LS_VERBOSE) << "Destroyed " #TYPE " \"" << name.c_str()    \
+                          << "\" (0 ref).";                              \
+    }                                                                    \
+  } else {                                                               \
+    RTC_LOG(LS_WARNING) << "Trying to remove reference from NULL " #TYPE \
+                           " object.";                                   \
+  }
```
fibann, line 124: nit: I am not a fan of random stuff in pchs. Move to tracked_object.h?

libs/mrwebrtc/include/interop_api.h
djee-ms: Yes, I agree but I wanted to avoid too many unrelated refactors on a non-`master` branch. Arguably this is already too much.

libs/Microsoft.MixedReality.WebRTC/Interop/RemoteAudioTrackInterop.cs
djee-ms: Will leave as a clean-up for `master`.

libs/mrwebrtc/src/pch.h
djee-ms: Actually, this is now a proper function so will delete those unused macros

djee-ms: Fixed.

libs/mrwebrtc/include/interop_api.h
```diff
@@ -80,38 +80,41 @@ struct mrsRemoteVideoTrackConfig;
 struct mrsDataChannelConfig;
```
fibann, line 80: There are a bunch of mentions of `mrsXRemoveRef` (and possibly AddRef too) in the comments, need search&replace.

libs/mrwebrtc/include/interop_api.h
fibann: Yeah, IMHO it's not incredibly useful as it is redundant with the actual class definitions and API comments and might fall out of sync, so I was thinking, if we want to get some good out of this handles definition let's go all the way! But it doesn't harm so can stay as it is.

libs/mrwebrtc/include/interop_api.h
djee-ms: Fixed.

PR #398, opened 2020-06-04
Make C# external video track source a proper source

djee-ms:
Change ExternalVideoTrackSource to derive from the same base class as
the device-based video track source, which is renamed to
DeviceVideoTrackSource for clarity. The base class is VideoTrackSource.


libs/mrwebrtc/src/media/external_video_track_source.h
```diff
@@ -170,14 +171,17 @@ class ExternalVideoTrackSource : public TrackedObject,
...
+      RefPtr<GlobalFactory> global_factory,
+      std::unique_ptr<detail::BufferAdapter> adapter,
+      rtc::scoped_refptr<detail::CustomTrackSourceAdapter> source);
   // void Run(rtc::Thread* thread) override;
   void OnMessage(rtc::Message* message) override;
 
+  // Keep a second reference (in addition of VideoTrackSource::source_) as a
+  // pointer to the specific type of adapter source, to simplify access and
+  // avoid having to cast everywhere.
   rtc::scoped_refptr<detail::CustomTrackSourceAdapter> track_source_;
```
fibann, line 185: nit: can we have a helper function do the cast rather than storing a separate pointer? Just to avoid pointers getting out of sync.

djee-ms: Fixed.

Issue #393, opened 2020-06-02
Error in AAR LibWebRTC merge.


rygo6:
I am trying to run the build process for android and am having an issue when it gets to this new code that was added here:
https://github.com/microsoft/MixedReality-WebRTC/commit/25dd9b5dd00794571febeada7b2269a425440345

This is my resulting error:

```
$ ./gradlew.bat assembleRelease

> Task :webrtc-native:externalNativeBuildRelease
Build mrwebrtc arm64-v8a
ninja: no work to do.

> Task :webrtc-native:mergeLibwebrtcArchive FAILED
./merge-archives.sh exited with code 1. Run with -v to get more info.
ERROR: input archive C:\Developer\MixedReality-WebRTC      ooluildndroid\webrtc-native/build/outputs/aar/mrwebrtc.aar is not a file

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':webrtc-native:mergeLibwebrtcArchive'.
> Process 'command 'bash'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
25 actionable tasks: 2 executed, 23 up-to-date
```

The path to the .aar seems a bit garbled from the echo -e output. But I checked in the merge-archives.sh and the actual path under $MRWEBRTC_AAR is not garbled and does point to a file.

Not sure if this new sh script needs debugging or if something else could be wrong? The android build process was working before this.

djee-ms:
Hi @rygo6,

What's the actual path that gets garbled? I have no issue myself locally. You are compiling on Windows with WSL2, right? I tested only with an actual Linux machine.

rygo6:
I am trying to compile with msys on windows. Should I be doing this with WSL2?

The actual path is this:
C:\Developer\MixedReality-WebRTC\tools\build\android\webrtc-native\build\outputs\aar\mrwebrtc.aar

I tried just hardcoding that path and also:
/c/Developer/MixedReality-WebRTC/tools/build/android/webrtc-native/build/outputs/aar/mrwebrtc.aar

Both had the same issue.

rygo6:
File definitely exists and the test command works in msys just fine to find the file.

```
[...]

PR #399, opened 2020-06-05
Fix for merge-archives.sh when building on windows.

rygo6-MS:
#393

### DIFF
--- a/tools/build/android/webrtc-native/merge-archives.sh
+++ b/tools/build/android/webrtc-native/merge-archives.sh
@@ -131,6 +131,11 @@ while getopts l:m:o:vh OPTION; do
     esac
 done
 
+# Replace C:/ for /mnt/c/ if using gradle in windows
+MRWEBRTC_AAR=${MRWEBRTC_AAR/C://mnt/c}
+MRWEBRTC_AAR=${MRWEBRTC_AAR//\\//}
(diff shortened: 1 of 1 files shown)


tools/build/android/webrtc-native/merge-archives.sh
```diff
@@ -131,6 +131,11 @@ while getopts l:m:o:vh OPTION; do
     esac
 done
 
+# Replace C:/ for /mnt/c/ if using gradle in windows
```
djee-ms, line 134: Not everyone builds on C:, this should be more generic.

tools/build/android/webrtc-native/merge-archives.sh
```diff
@@ -131,6 +131,11 @@ while getopts l:m:o:vh OPTION; do
...
+MRWEBRTC_AAR=${MRWEBRTC_AAR/C://mnt/c}
+MRWEBRTC_AAR=${MRWEBRTC_AAR//\\//}
+echo $MRWEBRTC_AAR
```
djee-ms, line 137: Can you please log a proper message rather than a value without context? It makes it difficult for users to understand what is this path they see. See `lib.sh` for many examples of output with `echo -e` and some ANSI escape for coloring if needed.

PR #400, opened 2020-06-05
Add Windows DLL signing to Unity UPM release pipeline

djee-ms:
Make the Unity UPM release pipeline sign the Windows DLLs before packaging.

### DIFF
--- a/tools/ci/templates/jobs-unity-package.yaml
+++ b/tools/ci/templates/jobs-unity-package.yaml
@@ -7,7 +7,7 @@ parameters:
 - name: buildAgent
   type: string
   default: ''
-- name: npmPackageVersion
+- name: upmPackageVersion
   type: string
   default: ''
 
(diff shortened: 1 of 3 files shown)


tools/ci/templates/jobs-unity-package.yaml
```diff
@@ -46,92 +46,158 @@ jobs:
...
+    inputs:
+      path: '$(Build.SourcesDirectory)/libs/unity/samples/$(PackedSamplesFilename)'
+      artifact: 'upm_samples'
 
-  # Delete .npmrc
-  - pwsh: |
-      Remove-Item -Path '$(Build.SourcesDirectory)/libs/unity/samples/.npmrc' -Force -ErrorAction Ignore
-    displayName: 'Delete .npmrc (samples)'
-    condition: always()
+  # # Write .npmrc
```
fibann, line 155: What's the deal with the commented code?

djee-ms: That's for directly pushing to a NPM register. It was difficult to figure out some proper auth. On the other hand we might want to not push directly from the pipeline because currently YAML build pipelines don't support manual gates (manual approval button) and it's safer to be able to inspect the build and eye it one time before actually publishing definitely to a public repository. In short: that's a memo for if/when ADO supports manual gating, and in the meantime we'll publish manually for safety with the code above it.

fibann: Add a comment explaining this (or delete)

djee-ms: Fixed.

PR #404, opened 2020-06-08
Update the Unity library to handle split sources

djee-ms:
Convert `WebcamSource` and `MicrophoneSource` to source objects shareable
among multiple peer connections.

Because the `SceneVideoSender` component uses the ARGB32 callback, also re-add
that callback, and add a TODO to remove all ARGB32 callbacks in favor of explicit
conversion utility functions, for clarity about the performance implication of the
conversion.

This change also splits the `VideoChatDemo` and `SceneCaptureDemo` for clarity,
leaving each demo with a single audio and video track per peer connection.
The `SceneCaptureDemo` now has a placeholder showing the capture camera
position. Multiple tracks per peer connection are demo-ed in the `StandaloneDemo`.

image

All demos are also updated to show essential components as Unity objects in the
scene, instead of adding multiple of them on the same `GameObject`. This helps with
discoverability.

image


libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
```diff
@@ -76,6 +76,34 @@ public string Name
...
+        /// Event that occurs when a video frame has been produced by the source.
+        /// </summary>
+        public event Argb32VideoFrameDelegate ARGB32VideoFrameReady
+        {
+            // TODO - Remove ARGB callbacks, use I420 callbacks only and expose some conversion
+            // utility to convert from ARGB to I420 when needed (to be called by the user).
+            add
+            {
+                bool isFirstHandler = (_argb32videoFrameReady == null);
+                _argb32videoFrameReady += value;
```
fibann, line 89: This should be under lock (same for the bit in `remove`).

libs/unity/library/Editor/PeerConnectionEditor.cs
```diff
@@ -209,6 +212,23 @@ public override void OnInspectorGUI()
...
+                {
+                    SettingsService.OpenProjectSettings("Project/Player");
+                }
+                if (GUILayout.Button("Add Microphone Capability"))
+                {
+                    PlayerSettings.WSA.SetCapability(PlayerSettings.WSACapability.Microphone, true);
+                }
+            }
+#endif
+
```
fibann, line 231: nit: why both here and in MicrophoneSource?

djee-ms: As explained in the help message, because of some "bug"/"feature" in the Audio Module implementation which currently checks audio devices on `libwebrtc` library startup, which crashes on UWP if audio permission is not granted. In comparison the Video Module correctly only accesses the webcam when creating a video track source, so no need to force that permission.

libs/unity/library/Runtime/Scripts/Media/VideoRenderer.cs
```diff
@@ -15,23 +16,27 @@ namespace Microsoft.MixedReality.WebRTC.Unity
...
-        [SerializeField]
-        protected MediaSource AudioSource;
+        // Note - Here what we really want is to serialize some reference to an IVideoSource.
+        // Unfortunately Unity before 2019.3 does not support serialized interfaces, and more generally ignores
+        // polymorphism in serialization. And although there are ways to make that work with a custom inspector,
+        // this disables support for object picker, so only drag-and-drop works, which is very impractical.
+        // So we use a class deriving from MonoBehaviour, which is the only entity for which Unity handles polymorphism.
+        // This makes the UI slightly confusing because we cannot filter audio vs. video sources, since C# supports only
+        // single inheritance and we already need to divide senders from receivers, so cannot also split by media kind.
+        // Instead we use OnValidate() to make sure the media source is a video source.
```
fibann, line 29: There are a bit too many classes IMHO and this seems to be one of the causes.

Can we split the hierarchy of track sources and have:
```
MonoBehaviour
|-VideoProducer
  |-VideoTrackSource
  |-VideoReceiver
|-AudioTrackSource
| AudioReceiver
```
? This would take out `MediaTrackSource` (which is almost empty) and `MediaReceiver` (which could be replaced by an interface maybe?) AFAICS these are only used by MediaLine, which could use MonoBehavior for serialized fields instead and cast to interfaces to call methods... But this doesn't sound too great. 

Any better ideas?

djee-ms: You're basically throwing away any type safety for the entire transceiver API.

fibann: How exactly? Isn't the UI for MediaLine gating assignments of sources/receivers? I admit serializing a MonoBehavior looks ugly but AFAICS the potential for introducing errors is limited (I might be wrong, haven't looked at all the code yet).

My main issue is that splitting logic across a long hierarchy of tiny classes with similar names is confusing since responsibilities are not clearly defined. If we need `MediaProducer` and `MediaTrackSource` for the sake of type-checking these serialized fields, maybe we can make this explicit and remove all logic from them, making them explicit placeholders? Specific suggestions on the class contents:
- the `MediaKind` field seems unused, and if used it could simply be an abstract property rather than a field with a chain of constructors initializing it
- the logic in `MediaProducer` handles a work queue, has nothing to do with media; maybe rename to WorkQueue and have `MediaProducer (empty) : WorkQueue : MonoBehaviour`?
- the logic in `MediaTrackSource` is superfluous - it would be simpler to define OnEnable()/OnDisable() in the subclasses directly and keep this class empty.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -3,43 +3,99 @@
...
-                    trackName = trackName
+                    AutoGainControl = _autoGainControl,
                 };
-                Track = await LocalAudioTrack.CreateFromDeviceAsync(trackSettings);
-
-                // Synchronize the track status with the Unity component status
-                Track.Enabled = enabled;
+                Source = await WebRTC.AudioTrackSource.CreateFromDeviceAsync(initConfig);
             }
         }
```
fibann, line 75: The flow of control going up in the base class and down into this class is more confusing than necessary. See other thread, it would be better to flatten `OnEnable` in the concrete class, maybe call into a helper function in `MediaTrackSource` doing the event raising.

libs/unity/library/Runtime/Scripts/Media/WebcamSource.cs
```diff
@@ -179,28 +215,27 @@ protected override async Task CreateLocalVideoTrackAsync()
...
 
             // Check H.264 requests on Desktop (not supported)
-#if !ENABLE_WINMD_SUPPORT
-            if (PreferredVideoCodec == "H264")
-            {
-                Debug.LogError("H.264 encoding is not supported on Desktop platforms. Using VP8 instead.");
-                PreferredVideoCodec = "VP8";
-            }
-#endif
+            //#if !ENABLE_WINMD_SUPPORT
```
fibann, line 218: nit: add TODO/FIXME or delete

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -0,0 +1,418 @@
...
+        /// If <see cref="Transceiver"/> is valid, that is a first session negotiation has already been completed,
+        /// then changing this value raises a <see xref="WebRTC.PeerConnection.RenegotiationNeeded"/> event on the
+        /// peer connection of <see cref="Transceiver"/>.
+        /// </summary>
+        public MediaTrackSource Source
+        {
+            get { return _source; }
+            set
+            {
+                if (_source != value)
```
fibann, line 80: There is a cast in `UpdateSenderOnSourceChanged`, but should we check the value type for correctness explicitly here for clarity?

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -0,0 +1,418 @@
...
+        /// If <see cref="Transceiver"/> is valid, that is a first session negotiation has already been conducted,
+        /// then changing this value raises a <see xref="WebRTC.PeerConnection.RenegotiationNeeded"/> event on the
+        /// peer connection of <see cref="Transceiver"/>.
+        /// </summary>
+        public MediaReceiver Receiver
+        {
+            get { return _receiver; }
+            set
+            {
+                if (_receiver != value)
```
fibann, line 122: This should check that the passed receiver has the correct type.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
```diff
@@ -1,16 +1,20 @@
...
-    /// with the intention to send them to the remote peer.
+    /// Base class for media senders acting as automated bridges between a media track source
+    /// (frame producer) and a transceiver of a peer connection. The media sender is managed
+    /// by the peer connection automatically, so users typically do not use this class or one
+    /// of its derived classes directly.
     /// </summary>
-    public abstract class MediaSender : MediaSource
+    /// <seealso cref="AudioSender"/>
+    /// <seealso cref="VideoSender"/>
+    public abstract class MediaSender
```
fibann, line 17: Any chance we can get rid of this and subclasses? They don't "send" anything, AFAICS they are boilerplate around a Track now. If we need them around maybe put into MediaLine (they are only used there) and/or make `internal`?

libs/unity/library/Editor/PeerConnectionEditor.cs
fibann: Right, didn't read the message carefully enough.

libs/unity/library/Editor/PeerConnectionEditor.cs
```diff
@@ -209,6 +212,23 @@ public override void OnInspectorGUI()
         {
             serializedObject.Update();
 
+#if UNITY_WSA
+            if (!PlayerSettings.WSA.GetCapability(PlayerSettings.WSACapability.Microphone))
+            {
+                EditorGUILayout.HelpBox("The UWP player is missing the Microphone capability. Currently on UWP the native WebRTC implementation always tries to"
+                    + " open the microphone while initialzing the audio subsystem at startup. Not granting access will fail initialization, and generally crash the app."
```
fibann, line 219:
```suggestion
                    + " open the microphone while initializing the audio subsystem at startup. Not granting access will fail initialization, and generally crash the app."
```

libs/unity/library/Runtime/Scripts/Media/VideoRenderer.cs
fibann: > You're basically throwing away any type safety for the entire transceiver API.

Also note that type safety is already far from watertight - you can pass an Audio producer to VideoRenderer and an AudioTrackSource to a video MediaLine and vice versa 😛

libs/unity/library/Runtime/Scripts/Media/WebcamSource.cs
djee-ms: Fixed.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
djee-ms: Fixed; the setter now correctly checks the type of `value`.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
djee-ms: Fixed; the setter now correctly checks the type of value.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
djee-ms: First pass to see what it looks like : made `MediaSender` _et al._ `internal` for now, but it is still convenient to have a storage for all the properties of the sender track instead of munching them into the `MediaLine`. However looking more closely the only 2 properties of interest are the track name and the transceiver's codec preferences, so maybe they can move to the `MediaLine` and we move all logic code to maintain the local track into `MediaLine` too. Thoughts?

libs/unity/library/Editor/PeerConnectionEditor.cs
djee-ms: Fixed.

libs/unity/library/Runtime/Scripts/Media/MediaSender.cs
djee-ms: Now removed completely in latest iteration.

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
```diff
@@ -263,13 +274,19 @@ internal void OnTracksRemovedFromSource(IEnumerable<LocalVideoTrack> tracks)
...
         }
 
         internal void OnArgb32FrameReady(Argb32VideoFrame frame)
         {
             MainEventSource.Log.Argb32LocalVideoFrameReady(frame.width, frame.height);
-            _argb32videoFrameReady?.Invoke(frame);
+            lock (_videoFrameReadyLock)
+            {
+                _argb32videoFrameReady?.Invoke(frame);
+            }
```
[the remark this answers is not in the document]
djee-ms: I prefer to leave the lock, because the first handler registered is used to determine when to register the native callback, and this is not atomic, so has more potential to cause hard-to-debug issues than a deadlock which is plain and visible. I documented instead that handlers must not add/remove handlers to the event while running.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -61,17 +61,25 @@ protected override async void OnEnable()
...
 #endif
 
-        protected override async Task CreateAudioTrackSourceAsyncImpl()
+        protected virtual async Task OnEnable()
```
fibann, line 64: Why virtual?

djee-ms: Why indeed ¯\_(ツ)_/¯
I think that might have been an automated refactor removing the implementation in the base class, because I don't see why I'd have added that in a leaf class.
Also I just noted there are 2 implementations of `OnEnable()` now on UWP, will merge.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -78,46 +56,86 @@ public IMediaTrackSource Source
...
+        /// https://tools.ietf.org/html/draft-ietf-mmusic-msid-05#section-2, which in
+        /// particular constraints the set of allowed characters to those allowed for a
+        /// 'token' element as specified in https://tools.ietf.org/html/rfc4566#page-43:
+        /// - Symbols [!#$%'*+-.^_`{|}~] and ampersand &amp;
+        /// - Alphanumerical characters [A-Za-z0-9]
+        /// 
+        /// Users can manually test if a string is a valid SDP token with the utility
+        /// method <see cref="SdpTokenAttribute.Validate(string, bool)"/>.
+        /// </remarks>
+        /// <seealso cref="SdpTokenAttribute.Validate(string, bool)"/>
```
fibann, line 113: nit: can we test that is a valid SDP token on `set`?

fibann: Hmm is this done when assigning `_senderTrackName`?

libs/Microsoft.MixedReality.WebRTC/VideoTrackSource.cs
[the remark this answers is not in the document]
fibann: There has to be a lock on add/remove because of the non-atomicity as you say, I was suggesting to avoid the lock _just_ here in the callbacks. `_argb32videoFrameReady?.Invoke(frame)` should read the EventHandler atomically and thus shouldn't need to be under lock. The only issue with this is that (see linked article) if a handler is `remove`d while a call to Invoke is ongoing, the Invoke thread will hold the handler alive until the end of the function, and also call the handler, which might be surprising and/or make disposal logic a mess.

Anyway fine by me as it is right now.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
djee-ms: `_senderTrackName` has the attribute, so this will make some error message appear when editing the field in the Unity editor if the SDP token is not valid. We can add some extra validation in the setter too for runtime.

djee-ms: Fixed.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
djee-ms: Fixed.

PR #407, opened 2020-06-10
Add Android build to Unity release pipeline

djee-ms:
Add building the Android archive `mrwebrtc.aar` and packing it into the UPM package of the MixedReality-WebRTC Unity library.

### DIFF
--- a/tools/build/libwebrtc/config.sh
+++ b/tools/build/libwebrtc/config.sh
@@ -66,7 +66,7 @@ function verify-arguments() {
 # Main
 
 # Read command line
-while getopts d:b:t:c:vsuh OPTION; do
+while getopts d:b:t:c:vsufh OPTION; do
     case ${OPTION} in
     d) WORK_DIR=$OPTARG ;;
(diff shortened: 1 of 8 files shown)


tools/build/libwebrtc/config.sh
```diff
@@ -7,7 +7,7 @@
 
 set -o errexit
 set -o nounset
-set -o pipefail
+#set -o pipefail # This breaks checkout on ADO
```
fibann, line 10: If possible explain how it breaks it exactly

djee-ms: Will document. This is related to bad handling in bash which makes in pop in some contexts (like ADO) and not others (like your local bash session). Not sure on the intricate details on why this happens though, but SIGTRAP seems to be actually informational mostly.
- Details : http://www.pixelbeat.org/programming/sigpipe_handling.html
- SIGPIPE bash limitation : https://lists.gnu.org/archive/html/bug-bash/2015-02/msg00052.html

PR #417, opened 2020-06-18
Fix invalid use of size_t instead of uint64_t

djee-ms:
API calls require the use of uint64_t by convention to avoid the
architecture-dependent size of size_t. Using size_t generates a cast
error on x86 architectures.


djee-ms:
This is currently breaking CI, and is only affecting the tests (not the library).
https://dev.azure.com/aipmr/MixedReality-WebRTC-CI/_build/results?buildId=12029&view=results

PR #420, opened 2020-06-22
Split audio device source and base audio source

djee-ms:
Split the audio track source base class from the actual device-based
implementation, and make the former abstract. This allow extensibility
in the future to add e.g. an external audio track source, and makes the
API consistent with video track sources.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/Interop/DeviceAudioTrackSourceInterop.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/Interop/DeviceAudioTrackSourceInterop.cs
@@ -1,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Runtime.InteropServices;
+
+namespace Microsoft.MixedReality.WebRTC.Interop
+{
+    /// <summary>
+    /// Handle to a native device audio track source object.
+    /// </summary>
+    internal sealed class DeviceAudioTrackSourceHandle : AudioTrackSourceHandle { }
+
+    internal class DeviceAudioTrackSourceInterop
+    {
+        /// <summary>
+        /// Marshaling struct for initializing settings when opening a local audio device.
+        /// </summary>
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+        internal ref struct LocalAudioDeviceMarshalInitConfig
+        {
+            public mrsOptBool AutoGainControl;
+
+            /// <summary>
+            /// Constructor for creating a local audio device initialization settings marshaling struct.
+            /// </summary>
+            /// <param name="settings">The settings to initialize the newly created marshaling struct.</param>
+            /// <seealso cref="DeviceAudioTrackSource.CreateAsync(LocalAudioDeviceInitConfig)"/>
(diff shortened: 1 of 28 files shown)


libs/Microsoft.MixedReality.WebRTC/Interop/DeviceAudioTrackSourceInterop.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
```
fibann, line 1: nit: given that these are internal and not a lot of code, I'd keep them in the original file, just for the sake of not scrambling the line history.

libs/mrwebrtc/include/audio_track_source_interop.h
```diff
@@ -7,20 +7,6 @@
 
 extern "C" {
```
fibann, line 8: This file is probably worth deleting now.

libs/mrwebrtc/src/interop/device_audio_track_source_interop.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "interop/global_factory.h"
+#include "media/device_audio_track_source.h"
+
+using namespace Microsoft::MixedReality::WebRTC;
+
+mrsResult MRS_CALL mrsDeviceAudioTrackSourceCreate(
+    const mrsLocalAudioDeviceInitConfig* init_config,
+    mrsDeviceAudioTrackSourceHandle* source_handle_out) noexcept {
+  if (!source_handle_out) {
+    RTC_LOG(LS_ERROR) << "Invalid NULL audio track source handle.";
```
fibann, line 18:
```suggestion
    RTC_LOG(LS_ERROR) << "Invalid NULL source_handle_out.";
```
"Invalid [...] handle" sounds like _the handle_ is invalid rather than the out pointer to it.

libs/mrwebrtc/include/audio_track_source_interop.h
djee-ms: I actually started implementing external audio sources, which is why a common base class is needed. But there's a fundamental issue with that, so I can't submit that change. So in the meantime I wanted to merge at least the prerequisite structural change to make external audio sources an incremental non-breaking change. Also I see no reason why we can't enable local audio callbacks on sources and that would go there.

libs/Microsoft.MixedReality.WebRTC/Interop/DeviceAudioTrackSourceInterop.cs
djee-ms: I'd rather keep things consistent so we can more easily find where things are, than randomly merge some classes into some files while keeping others in separate files. Again, I intended to have external audio sources as a second derived class from that same base `AudioTrackSource`, and cramming everything in a single file sounds confusing. Hopefully I can add that change soon, but waiting on guidance from Google.

fibann: If we plan to put things in AudioTrackSourceInterop this should be fine then.

> I'd rather keep things consistent so we can more easily find where things are, than randomly merge some classes into some files while keeping others in separate files.

I see your point, though in general I believe symmetry in the file system is not that important (for me at least) when one browses the code mostly using search and jumping from point of use to definition.

PR #424, opened 2020-06-24
Fix remote audio bugs

fibann:


libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
```diff
@@ -158,6 +162,22 @@ protected void OnDestroy()
             AudioSettings.OnAudioConfigurationChanged -= OnAudioConfigurationChanged;
         }
 
+        protected new void Update()
+        {
+            base.Update();
+
+            // Check if _track has been changed by OnPaired/OnUnpaired and
+            // we need to start/stop streaming.
+            if (_track != null && !IsStreaming)
```
djee-ms, line 171: This is essentially replacing an event-based implementation with a polling one, why?

libs/mrwebrtc/src/media/audio_track_read_buffer.h
```diff
@@ -19,11 +20,12 @@ struct AudioFrame;
 class PeerConnection;
 
 /// Implementation of |mrsAudioTrackReadBufferHandle|.
-class AudioTrackReadBuffer : public webrtc::AudioTrackSinkInterface {
+class AudioTrackReadBuffer : public TrackedObject, webrtc::AudioTrackSinkInterface {
```
djee-ms, line 23: Why is this needed?

libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
fibann: Note that this is just reapplying changes from #359.

The event-based implementation is not really so, as the handlers just enqueue operations on the Update thread. This doesn't work well if the state changes while the component is disabled - multiple actions are enqueued and then run all at once when the component is enabled again. It's difficult to keep the state consistent at each step while doing so. Putting the logic in Update() is straightforward.

libs/mrwebrtc/src/media/audio_track_read_buffer.h
fibann: The destructor calls a WebRTC function and needs the signaling thread to be alive at that point.

djee-ms: It's unfortunate we (or rather, I) mixed the concepts of ref-counted and tracked into the same class 😢

libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
```diff
@@ -158,6 +162,22 @@ protected void OnDestroy()
             AudioSettings.OnAudioConfigurationChanged -= OnAudioConfigurationChanged;
         }
 
+        protected new void Update()
```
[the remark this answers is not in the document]
fibann: Fixed

PR #429, opened 2020-06-25
Update the C# UWP tutorial to reflect the new API

djee-ms:
Bug: #244


docs/manual/cs/helloworld-cs-connection-uwp.md
```diff
@@ -15,9 +15,11 @@ Continue editing the `OnLoaded()` method and append after the [`InitializeAsync(
...
+   The [`Connected`](xref:Microsoft.MixedReality.WebRTC.PeerConnection.Connected) event is raised when the peer connection is established, that is when the underlying media transports are writable. There is a subtelty here, in that this does not necessarily correspond to an offer/answer pair exchange being completed; indeed on the callee (answering peer) once the answer is created everything is done and the transports are writable, but the answer has not been sent yet, so the caller (offering peer) needs to wait until it received and applied that answer for its transports to be writable, and therefore its [`Connected`](xref:Microsoft.MixedReality.WebRTC.PeerConnection.Connected) event will be raised much later than the one of the callee (answering peer).
```
fibann, line 18:
```suggestion
   The [`Connected`](xref:Microsoft.MixedReality.WebRTC.PeerConnection.Connected) event is raised when the peer connection is established, that is when the underlying media transports are writable. There is a subtlety here, in that this does not necessarily correspond to an offer/answer pair exchange being completed; indeed on the callee (answering peer) once the answer is created everything is done and the transports are writable, but the answer has not been sent yet, so the caller (offering peer) needs to wait until it received and applied that answer for its transports to be writable, and therefore its [`Connected`](xref:Microsoft.MixedReality.WebRTC.PeerConnection.Connected) event will be raised much later than the one of the callee (answering peer).

```

PR #430, opened 2020-06-25
Init/close PeerConnection on enable/disable

### LINKED ISSUE #375: Cannot stop a call and start a new call
**Describe the bug**
Have built an application in unity using the master branch where i start a call between 2 peers and everything works fine.  I then end the call and try to start a new call and I get the following error when trying to start the new call and no video or audio starts to pass but connection is established:

ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
System.ThrowHelper.ThrowArgumentOutOfRangeException (System.ExceptionArgument argument, System.ExceptionResource resource) (at <437ba245d8404784b9fbab9b439ac908>:0)
System.ThrowHelper.ThrowArgumentOutOfRangeException () (at <437ba245d8404784b9fbab9b439ac908>:0)
System.Collections.Generic.List`1[T].get_Item (System.Int32 index) (at <437ba245d8404784b9fbab9b439ac908>:0)
Microsoft.MixedReality.WebRTC.Unity.PeerConnection.StartConnection () (at Assets/Microsoft.MixedReality.WebRTC.Unity/Scripts/PeerConnection.cs:912)

In my project I call PeerConnection.StartConnection() to start the call and PeerConnection.Uninitialize() and PeerConnection.Close() to stop the connection. 

**To Reproduce**
Start a Peer to Peer call. End a Peer to Peer call and then try to start a new peer to peer call.

**Expected behavior**
Should be able to start and stop calls without having to fully quit application

**Environment**

 - Platform: Unity Editor to WIndows Standalone build
 - Architecture:  Unity Editor is x64 on Windows and build is x64
 - Unity version (if applicable): 2019.3.9f1
 - Target device: Windows Desktop

fibann:
Ensure proper shutdown across MediaLines & co.

Removed AutoInitializeOnStart, and made Initialize/Uninitialize private,
now PC should be turned on/off by enabling/disabling.

We now create/keep local tracks only when there are both a live source and a
connection. Therefore, local tracks do not persist after the PeerConnection
component is disabled. This is intended, as we cannot keep the track and
attach it to a different PeerConnection if the component is re-enabled.

Removed listeners to tracks removed. Unpairing is handled manually on
renegotiation and on shutdown (MediaLine.UnpairTransceiver).

Fix #375.


libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
```diff
@@ -308,6 +308,10 @@ void IMediaReceiverInternal.DetachFromTransceiver(Transceiver transceiver)
         {
             Debug.Assert((_transceiver == null) || (_transceiver == transceiver));
             _transceiver = null;
+            _track = null;
```
djee-ms, line 311: This change and the one in `VideoReceiver.cs` seem quite different, why are things so asymmetrical?

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -322,26 +319,28 @@ internal void DetachTrack()
...
-            Debug.Assert(transceiver != null);
-            bool wantsRecv = (Receiver != null);
-            bool wasReceiving = (_pairedReceiver != null);
-            bool isReceiving = (transceiver.RemoteTrack != null);
-            // Note the extra "isReceiving" check, which ensures that when the remote track was
-            // just removed by OnUnpaired(RemoteTrack) from the TrackRemoved event then it is not
-            // immediately re-added by mistake.
-            var receiverInternal = (_receiver as IMediaReceiverInternal);
-            if (wantsRecv && isReceiving && !wasReceiving)
+            if (_receiver != null)
```
djee-ms, line 322: This seems to break the logic : if a user removes its receiver object from the MediaLine of the peer connection, we should clear the receiver track no? Or is it precisely that we cannot remove a receiver track without renegotiation, so changing the Receiver on the MediaLine should not have any effect until it's renegotiated?

libs/unity/library/Runtime/Scripts/PeerConnection.cs
```diff
@@ -558,75 +544,57 @@ public async Task HandleConnectionMessageAsync(SdpMessage message)
...
             if (AutoLogErrorsToUnityConsole)
             {
                 OnError.AddListener(OnError_Listener);
             }
-
-            if (AutoInitializeOnStart)
-            {
-                InitializeAsync();
-            }
+            InitializeAsync();
```
djee-ms, line 588: I didn't test it but is there any incentive to awaiting here? Is Unity changing the `enabled` state before or after the `OnEnable()` was called?

libs/unity/library/Runtime/Scripts/PeerConnection.cs
```diff
@@ -135,15 +135,6 @@ public class PeerConnection : WorkQueue
...
         #region Behavior settings
 
-        /// <summary>
-        /// Initialize the peer connection on <a href="https://docs.unity3d.com/ScriptReference/MonoBehaviour.Start.html">MonoBehaviour.Start()</a>.
-        /// If this field is <c>false</c> then the user needs to call <see cref="InitializeAsync(CancellationToken)"/>
-        /// to manually initialize the peer connection before it can be used for any purpose.
-        /// </summary>
-        [Tooltip("Automatically initialize the peer connection on Start().")]
-        [Editor.ToggleLeft]
-        public bool AutoInitializeOnStart = true;
```
djee-ms, line 145: We should probably document in the class header this new behavior.

libs/unity/library/Runtime/Scripts/PeerConnection.cs
fibann: The docs say it's called "just after the object is enabled" so I'd expect after. awaiting here seems pointless - if this becomes async there is probably nothing subscribing to the returned Task in the caller - and waiting seems harmful. So I guess users should not rely on `enabled` implying `Peer != null` and should wait for `OnInitialize`

libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
fibann: VideoReceiver doesn't seem to _do_ anything except exposing the `Track` callback, so there isn't much that it can do on Enable/Disable, and the logic around it is absent.

This raises some interesting questions:
- should the Transceiver direction be controlled by the enabled state of the connected receiver? At the moment it isn't, but it's unfortunate that we keep streaming in a track that we aren't rendering.
- can we improve this? At the moment `AudioReceiver` merges the feature of a "receiver" (plugs into a MediaLine and forwards its frames) and a "renderer" (plays the audio), while for video this is done by two separate classes. So we could make things more coherent by:
  - splitting `AudioReceiver` into two classes like for video; clear but not great (we need two different components for audio too, we have a long chain of classes already)
  - merging `VideoRenderer` and `VideoReceiver` -  have an `AudioRenderer` (rename from `AudioReceiver`) which renders audio automatically and a `VideoRenderer` that renders video automatically and all direct access to frames can done in code (which is already the case) by accessing the remote track in the MediaLine. Sounds nice - allows us to fix the hierarchy quirks around VideoRendererSource - but there are a few issues: how do we attach two renderers to the same track? what UI we make to tell "I want to receive this track but don't have a Renderer object to attach"?
  - make `XReceiver` just expose the callbacks and have `XRenderer` inherit from it - then you plug a base receiver into the MediaLine if you only want the callbacks, or a renderer if you want the stream to be rendered automatically (or you write your own renderer class). Still difficult to attach multiple renderers though.

Thoughts?

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: > Or is it precisely that we cannot remove a receiver track without renegotiation, so changing the Receiver on the MediaLine should not have any effect until it's renegotiated?

This seems to be correct - in any case sounds unrelated, if a user removes its receiver object from the MediaLine they won't go through this method, even before my change.

Re this specific change, this has the only possible effect of calling methods on `_receiver` so it makes sense to skip if there is no receiver. I guess the name `UpdateReceiver` is ambiguous, will rename to `UpdateReceiverPairingIfNeeded`.

libs/unity/library/Runtime/Scripts/Media/AudioReceiver.cs
fibann: This has been refactored in a different PR.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: This has been refactored in a different PR.

libs/unity/library/Runtime/Scripts/PeerConnection.cs
fibann: Fixed

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
```diff
@@ -1,16 +1,61 @@
...
+            {
+                pc.enabled = true;
+            }
+
+            // Check the event was raised
+            var timeout = DateTime.Now + TimeSpan.FromSeconds(10);
+            yield return new WaitUntil(() => isInitialized || DateTime.Now > timeout);
+            pc.OnInitialized.RemoveListener(listener);
+
+            Assert.IsNotNull(pc.Peer);
```
djee-ms, line 40:
```suggestion
            pc.OnInitialized.RemoveListener(listener);
            Assert.IsTrue(isInitialized);
            Assert.IsNotNull(pc.Peer);
```

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
```diff
@@ -1,16 +1,61 @@
...
+            pc.OnShutdown.AddListener(listener);
+
+            pc.enabled = false;
+            Assert.IsNull(pc.Peer);
+
+            // Check the event was raised
+            var timeout = DateTime.Now + TimeSpan.FromSeconds(10);
+            yield return new WaitUntil(() => isShutdown || DateTime.Now > timeout);
+            pc.OnShutdown.RemoveListener(listener);
+        }
```
djee-ms, line 57:
```suggestion
            pc.OnShutdown.RemoveListener(listener);
            Assert.IsTrue(isShutdown);
        }
```

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
```diff
@@ -28,25 +73,118 @@ public IEnumerator SimplePeerConnection()
...
+            ml2.Receiver = receiver2;
+
+            // Init/quit twice.
+            for (int i = 0; i < 2; ++i)
+            {
+                // Initialize
+                var init1 = InitializeAndWait(pc1);
+                var init2 = InitializeAndWait(pc2);
+                yield return init1;
+                yield return init2;
```
djee-ms, line 139:
```suggestion
                yield return InitializeAndWait(pc1);
                yield return InitializeAndWait(pc2);
```

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
```diff
@@ -28,25 +73,118 @@ public IEnumerator SimplePeerConnection()
...
+                // Shutdown peer #1
+                pc1.enabled = false;
+                Assert.IsNull(pc1.Peer);
+
+                // The source is not impacted, but tracks and transceiver are gone.
+                Assert.IsTrue(source1.IsLive);
+                Assert.IsFalse(receiver1.IsLive);
+                Assert.IsNull(ml1.SenderTrack);
+                Assert.IsNull(ml1.Transceiver);
+
```
djee-ms, line 174: Add some checks about the influence of shutting down the remote peer connection but not the local one? I assume the connection will be closed, but transceivers may still exist as well as tracks. Though I don't know if there's a good way to be notified when the connection is closed, and this might only be timeout-based, which is difficult to test.

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
djee-ms: Other places too; it's unusual to see a `IEnumerator` captured instead of just returning it / yielding it. Although technically I guess there's nothing wrong with this.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -280,43 +277,60 @@ private void UpdateTransceiverDesiredDirection()
             }
         }
 
-        internal void AttachTrack()
+        // Initializes and attaches a local track if all the preconditions are satisfied.
+        private void StartLocalTrackIfPossible()
```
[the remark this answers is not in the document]
fibann: It's "create and attach to transceiver" that I conflated into "start". I can use "create" if that sounds better no strong opinions.

libs/unity/library/Runtime/Scripts/PeerConnection.cs
```diff
@@ -487,12 +478,20 @@ public async Task HandleConnectionMessageAsync(SdpMessage message)
...
+                    {
+                        Debug.Assert(tr == mediaLine.Transceiver);
+                    }
 
                     // Associate the transceiver with the media line, if not already done, and associate
                     // the track components of the media line to the tracks of the transceiver.
                     try
                     {
-                        mediaLine.UpdateOnReceiveOffer(tr);
+                        mediaLine.UpdateAfterSDPMessage();
```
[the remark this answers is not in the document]
fibann: Yes, will change.

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
fibann: The idea is to start the two async jobs together then wait for them. Though chances are that in this case they will be run on the same thread, so it doesn't make much difference, I'll go with the two-liner.

libs/unity/library/Tests/Runtime/PeerConnectionTests.cs
fibann: I tried waiting for a frame and the local peer (not shutdown) didn't report any change. So yeah, I am not sure if there is a way to reliably detect disconnection. I'll leave it like this and add a comment.

fibann:
> this design prevents the user from hot-swapping a track on an active transceiver, right?

I have changed `Source.set` to swap the new track before disposing the old one so that should make track swapping fast (we can't persist the tracks themselves but recreating them should be cheap enough).

There turned out to be a bug there, fix is now up (plus tests).

PR #431, opened 2020-06-26
Always add MediaLines to track source

### LINKED ISSUE #426: MediaLines on track source are not always initialized
They are only set when setting `Source` on a MediaLine at runtime. 

This causes an exception if there is a call active and the shutdown order is such that the source is destroyed before the PC. This in turn seems to cause dangling resources with consequent freezing Unity editor on close/play again.

fibann:
The fix is in MediaLine.OnAfterDeserialize. The rest is small refactorings/simplifications.

This fixes #426. The behavior is still a bit weird - we are removing the source from the MediaLine when it is disabled, but have no way to re-add it when it is re-enabled.

### DIFF
--- a/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
@@ -25,7 +25,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity
     /// inspector window, though direct manipulation via code is also possible.
     /// </summary>
     [Serializable]
-    public class MediaLine
+    public class MediaLine : ISerializationCallbackReceiver
     {
         /// <summary>
         /// Kind of media of the media line and its attached transceiver.
@@ -58,13 +58,9 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             {
                 if (value == null)
                 {
-                    if (_source != null)
-                    {
-                        (_source as IMediaTrackSourceInternal).OnRemoveFromMediaLine(this);
-                        _source = null;
-                        UpdateSenderOnSourceChanged();
-                        UpdateTransceiverDesiredDirection();
-                    }
+                    (_source as IMediaTrackSourceInternal)?.OnRemoveFromMediaLine(this);
+                    _source = null;
+                    DestroySenderIfNeeded();
                 }
                 else
                 {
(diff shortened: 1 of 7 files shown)


libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -552,18 +515,22 @@ internal void UpdateOnReceiveAnswer()
...
+
+        public void OnBeforeSerialize() {}
+
+        public void OnAfterDeserialize()
+        {
+            if (_source)
             {
-                videoTrack.Dispose();
+                // Fill the list of media lines for the source.
+                ((IMediaTrackSourceInternal)_source).OnAddedToMediaLine(this);
```
djee-ms, line 532: `OnAddedToMediaLine()` asserts that the media line list doesn't contain this source already, but `OnAfterDeserialize()` is called in many places, including apparently when cloning an object, and not necessarily only when the current object is created/loaded the first time (see [docs](https://docs.unity3d.com/ScriptReference/ISerializationCallbackReceiver.html)). So `OnAddedToMediaLine()` should probably be updated to avoid asserting.

fibann: It shouldn't be called twice on the same object though, right?

djee-ms: No idea, the docs seem to say it is...

fibann: The docs say
> Serialization can occur during all kinds of operations. For example, when using Instantiate() to clone an object, Unity serializes and deserializes the original object

IMHO they seem to suggest to be aware that those callbacks might be called at runtime, in addition to when files are loaded from disk. They don't seem to imply that callbacks might be called arbitrarily on an object that has already been deserialized. So I see no problem here.

djee-ms: Fine. "Unity serializes and deserializes the original object" tells me they do are calling it, but on the other hand that makes no sense to deserialize the source object of a clone, so let's assume it's poor wording.

fibann: I agree it sounds weird but it's reasonable to assume that they mean "serializes the original object and deserializes the serialized original object" 😄

Issue #246, opened 2020-03-31
Add video capture on Android


djee-ms:
This is a TODO task to add video capture on Android once #193 is merged to `master`.

PR #433, opened 2020-06-30
Android video capture

djee-ms:
Add video capture support on Android via a dedicated Java class
`AndroidCameraHelper` which does the interop with the utilities of `libwebrtc`.
The implementation is straightforward, with the notable difference that the
Java camera object is owned by the video track source and needs to be kept
alive by it.

This change slightly refactors the video capture device and format enumerations
by moving them into the `DeviceVideoTrackSource` class for consistency, and use
some structs for parameter passing in the interop for better extensibility.

Bug: #246


tools/build/android/webrtc-native/build.gradle
```diff
@@ -62,6 +62,12 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'androidx.appcompat:appcompat:1.1.0'
+    // See e.g. examples/aarproject/app/build.gradle
+    // Note: com.google.code.findbugs is for @Nullable,
+    // see https://stackoverflow.com/questions/19030954/cant-find-nullable-inside-javax-annotation
+    // FIXME - This uses published packages, not locally-built ones!!!
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
+    implementation 'org.webrtc:google-webrtc:1.0.+'
```
djee-ms, line 70: This needs to be fixed before merging, but should have no influence on the rest of the code, so opening for review already in the meantime.

libs/mrwebrtc/src/utils.h
```diff
@@ -67,6 +68,40 @@ std::string ObjectToString(TrackedObject* obj);
 bool IsValidAudioTrackBufferPadBehavior(
     mrsAudioTrackReadBufferPadBehavior pad_behavior);
 
+/// Callback-based asynchronous enumerator utility.
+///
+/// The utility takes a mandator enumeration callback, which is called each time
```
fibann, line 73:
```suggestion
/// The utility takes a mandatory enumeration callback, which is called each time
```

tools/build/android/webrtc-native/build.gradle
djee-ms: Fixed.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -75,6 +107,47 @@ protected async void OnEnable()
...
+            // If focus is restored after a pending request, check the permission again
+            if (_androidRecordAudioRequestPending)
+            {
+                _androidRecordAudioRequestPending = false;
+
+                if (Permission.HasUserAuthorizedPermission(Permission.Microphone))
+                {
+                    // If now authorized, start capture as if just enabled
+                    Debug.Log("User granted authorization to access microphone, starting MicrophoneSource now...");
+                    OnEnable();
```
fibann, line 127: nit: extract video initialization in its own function rather than calling OnEnable again - shouldn't make much difference but flow looks clearer IMHO.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -75,6 +107,47 @@ protected async void OnEnable()
...
+                }
+                else if (Time.time <= _androidRecordAudioRequestRetryUntilTime)
```
fibann, line 129: It seems that if we take this branch then we'll get stuck (won't enable the source nor log the error message) until the application loses focus and gets it again, is that correct? Rather than relying on focus, should we check every e.g. 1s with a coroutine?

libs/mrwebrtc/src/media/device_video_track_source.cpp
```diff
@@ -252,29 +296,449 @@ ErrorOr<RefPtr<DeviceVideoTrackSource>> DeviceVideoTrackSource::Create(
...
     return Error(Result::kUnknownError);
   }
 
   // Create the wrapper
   RefPtr<DeviceVideoTrackSource> wrapper =
-      new DeviceVideoTrackSource(global_factory, std::move(video_source));
+      new DeviceVideoTrackSource(global_factory, std::move(video_source)
+#if defined(MR_SHARING_ANDROID)
+                                                     ,
+                                 java_video_capturer
```
fibann, line 378: nit: weird formatting

libs/mrwebrtc/src/media/device_video_track_source.cpp
```diff
@@ -252,29 +296,449 @@ ErrorOr<RefPtr<DeviceVideoTrackSource>> DeviceVideoTrackSource::Create(
...
     : VideoTrackSource(std::move(global_factory),
                        ObjectType::kDeviceVideoTrackSource,
-                       std::move(source)) {}
+                       std::move(source))
+#if defined(MR_SHARING_ANDROID)
+      ,
+      java_video_capturer_(java_video_capturer)
+#endif  // defined(MR_SHARING_ANDROID)
+{
+}
```
fibann, line 722: nit: it would probably be nicer to have the Android logic in a subclass of DeviceTrackSource (or at least to #ifdef the whole constructor here)

libs/unity/library/Runtime/Scripts/Media/WebcamSource.cs
```diff
@@ -212,6 +245,47 @@ protected async void OnEnable()
                 }
             }
 #endif
+#if PLATFORM_ANDROID
```
fibann, line 248: nit: merge with previous `#if PLATFORM_ANDROID` section?

libs/mrwebrtc/src/media/device_video_track_source.cpp
djee-ms: **Coding style rules**
Rule 1/1 : `clang-format` is always right.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
djee-ms: For how long do you suggest to check? There is no way to tell if the user already denied access and there's no dialog, or if it is the first time and the dialog is still open and the user didn't select an option yet; Java has the distinction but Unity doesn't expose it and return the same result in both cases. The latter can take an indeterminate amount of time, and having a timeout on a user dialog is pretty bad. But if it's the former, then you will continue to check forever every second for no reason, which is almost worse.

With the current approach:
- If the user already denied, we never check again. Granted we lose the error message; we could maybe have a warning when we try the first time in `OnEnable()` just for debugging. But at least we likely will never check again, which is what should happen since there's no chance we'll get permission.
- If the dialog is open, then everything works as intended since the focus will change and we'll get the timeout.

fibann: > For how long do you suggest to check?

300s? 😄 

Let me re-frame the problem. I agree everything works fine if the user accepts the dialog and permission is granted (first branch here). If the user waits 5 min and then denies permission, it also works fine (third branch here).

If the user denies permission before 5 min (second branch), capture silently fails. An error message is printed only when (if) a focus switch happens after 5 min. Also the component being disabled or not depends on when the focus switch happens. So the behavior can be improved here.

I'd say we should:
- keep checking for permission granted (and only for that) in `OnApplicationFocus`
- check on a 5min timeout in a coroutine started from `OnEnable` rather than in `OnApplicationFocus`, so whether a denial is detected doesn't depend on focus switches
- avoid disabling the component on permission denied? Since we can't reliably detect denial I'd avoid influencing the app behavior.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
djee-ms: I did it like this on purpose to make sure no one introduces some code in `OnEnable()` that would not be run through this workaround. I agree with you usually I would have extracted a separate function, but here I want to avoid future troubles.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
djee-ms: `MicrophoneSource` and `WebcamSource` always disable themselves if failing to create a track, as we tie the track lifetime to the enable/disable state. Failing to do so only in one specific case and only on Android will cause confusion.

djee-ms: (well at least after #436 which I was about to open)

fibann: > only in one specific case and only on Android

Of course we'd need to adopt the same logic on UWP too. And I'd avoid disabling the component on failure to create a track too, for the reason below.

> we tie the track lifetime to the enable/disable state

Not really - starting the track is asynchronous so there is inevitably a lapse of time where the component is enabled but there is no track. Users should use `IsStreaming`/`XStreamStarted` to detect whether there is an active track. My proposal is that the component would stay in that state forever on failure, which is less surprising IMHO than staying there for a few seconds or for 5 min and then automatically reverting to disabled. But I guess this might be subjective.

djee-ms: Alright. I guess the argument is also that we need to stick to `OnEnable()` for creating the track to make sure that if the track exists then the `Update()` method is called, so a separate public `StartCaptureAsync()` is too dangerous for that reason.

One thing bothering me though : what happen if you enable then disable then enable quickly, so that the first track has not yet finished being created. I suspect that because of `await` you end up trying to create a second track while the first is not done yet unfortunately, so we cannot just rely on `OnDisable()` always occurring after the track is created, nor `OnEnable()` after the track was destroyed, and need more synchronizing.

libs/mrwebrtc/src/media/device_video_track_source.cpp
djee-ms: Separated the #ifdef for clarity.

libs/unity/library/Runtime/Scripts/Media/WebcamSource.cs
djee-ms: This is the Android variant for the auto format block. I merged `#endif` + `#if` into `#elif` for clarity, but that should stay before the variable declaration. And the block above is about permissions. I prefer to have them clearly separated to show the steps. That's mostly a matter of taste though.

fibann: No strong preferences. I guess putting the steps into separate methods would be the clearest way to show the steps though!

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
djee-ms: Logged as #439 to look at that issue specifically.

PR #448, opened 2020-07-09
Reintroduce MediaTrackSource

fibann:
Now that there is no VideoRendererSource we don't need it to be an interface anymore.


libs/unity/library/Runtime/Scripts/Media/MediaTrackSource.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        private readonly List<MediaLine> _mediaLines = new List<MediaLine>();
+
+        internal void OnAddedToMediaLine(MediaLine mediaLine)
+        {
+            Debug.Assert(!_mediaLines.Contains(mediaLine));
+            _mediaLines.Add(mediaLine);
+        }
+
+        internal void OnRemoveFromMediaLine(MediaLine mediaLine)
```
djee-ms, line 34: nit: for symmetry
```suggestion
        internal void OnRemovedFromMediaLine(MediaLine mediaLine)
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -51,40 +51,35 @@ public class MediaLine : ISerializationCallbackReceiver
...
+                    if (_source != value)
                     {
-                        if (_source != mediaTrackSource)
+                        if (_source != null)
                         {
-                            (_source as IMediaTrackSourceInternal)?.OnRemoveFromMediaLine(this);
-                            _source = mediaTrackSource;
-                            (_source as IMediaTrackSourceInternal).OnAddedToMediaLine(this);
-                            CreateSenderIfNeeded();
+                            _source.OnRemoveFromMediaLine(this);
```
djee-ms, line 78: nit: `_source?.OnRemoveFromMediaLine(this);` without the explicit `if`?

fibann: `_source` is now a MonoBehaviour and `?.` will give a warning

djee-ms: Right, nevermind.

PR #449, opened 2020-07-09
Expose WebRTC logging via user callback

djee-ms:
Allow the user to register a callback invoked each time a message is
logged by the internal WebRTC implementation. This allows better error
reporting and easier debugging without the need to connect a debugger or
even have debug symbols available (PDBs).

### DIFF
--- a/libs/mrwebrtc/include/logging_interop.h
+++ b/libs/mrwebrtc/include/logging_interop.h
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "interop_api.h"
+
+extern "C" {
+
+/// Severity of a log message.
+enum class mrsLogSeverity : int32_t {
+  kUnknown = -1,  // Could not assign a severity level
+  kVerbose = 1,
+  kInfo = 2,
+  kWarning = 3,
+  kError = 4,
+  kNone = 5,
+};
+
+struct mrsLogSinkHandleImpl {};
+
+/// Handle to a registered log sink, used to unregister it.
+using mrsLogSinkHandle = mrsLogSinkHandleImpl*;
+
+/// Callback invoked when a log message is received. The callback receives the
+/// log message severity and a string with the message content. The first
+/// argument is the opaque |user_data| parameter passed during registration.
+using mrsLogMessageCallback = void(MRS_CALL*)(void*,
+                                              mrsLogSeverity,
+                                              const char*);
+
+/// Register a log message sink in the form of a callback invoked when a log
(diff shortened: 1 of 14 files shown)


djee-ms:
Note that:
- There is no Unity library change; didn't think about whether there's something needed or not yet. This change only expose to the C and C# libraries.
- This requires patching the Google m71 branch due to a "bug" (missing severity argument) fixed in a later version, hence the change in submodules. The change has already been applied to `webrtc-uwp-sdk`.
- Removing the `branch` argument from `.gitmodules` which is confusing (does not do what most people expect, and anyway we don't use the `releases/m71` branch anymore but `mrwebrtc` which is branched from it).

libs/mrwebrtc/include/logging_interop.h
```diff
@@ -0,0 +1,46 @@
...
+/// Register a log message sink in the form of a callback invoked when a log
+/// message is produced. The callback is invoked for any message with a severity
+/// great than or equal to the specified |min_severity|.
+MRS_API mrsLogSinkHandle MRS_CALL
+mrsLoggingAddSink(mrsLogSeverity min_severity,
+                  mrsLogMessageCallback callback,
+                  void* user_data) noexcept;
+
+/// Unregister a previously registered log message sink.
+MRS_API void MRS_CALL mrsLoggingRemoveSink(mrsLogSinkHandle handle) noexcept;
```
fibann, line 41: nit: the others are `mrsLog`, this one is `mrsLogging`?

libs/mrwebrtc/src/interop/logging_interop.cpp
```diff
@@ -0,0 +1,118 @@
...
+using namespace Microsoft::MixedReality::WebRTC;
+
+namespace {
+
+class InteropLogSink : public rtc::LogSink {
+ public:
+  using MessageCallback = Callback<mrsLogSeverity, const char*>;
+
+  static InteropLogSink* Create(MessageCallback callback) noexcept {
+    try {
```
fibann, line 20: Is the try/catch defensive, or is there something specific we are looking out for?

djee-ms: It's defensive against the various `std` methods throwing, since this is a `noexcept` function.

libs/mrwebrtc/include/logging_interop.h
djee-ms: `mrsLoggingLogMessage()` sounded like a mouthful. I know it breaks consistency. Although in general when you have some logging feature the configuration is from a logging class or functions, but the logging itself is generally a macro or a very short function name, so I don't feel too bad about this particular one.

PR #453, opened 2020-07-13
Add Python build script for Windows builds

djee-ms:
Add a Python build.py script for Windows builds which automatically
clone the Google repository and eventually applies the UWP patches if
needed, then build the given variant. This greatly simplifies building
for users.

### DIFF
--- a/README.undock.md
+++ b/README.undock.md
@@ -1,41 +1,48 @@
 # Experimental undock branch
 
-## Building
+## Checkout
+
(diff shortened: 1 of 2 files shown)


README.undock.md
```diff
@@ -1,41 +1,48 @@
 # Experimental undock branch
```
fibann, line 1: Can we delete the PS script now?

djee-ms: It's not on parity yet, but as soon as it is yes, that's the plan.

PR #457, opened 2020-07-14
Add support for multi-config build to build.py

djee-ms:
Enable building multiple configurations at once from the same command
call.


djee-ms:
Add what `build.ps1` was missing : multi-build.
```sh
python build.py --target win,winuwp --cpu x86,x64,arm --release
```
will build all valid configs and correctly skip invalid ones (_e.g._ `win-arm-release`).

image

PR #458, opened 2020-07-14
Fix edits of receiver pairing state

fibann:
Limit Source and Receiver changes to main thread for simplicity.

### DIFF
--- a/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
@@ -362,7 +362,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             // automatically by the implementation during SetRemoteDescription() in recvonly
             // mode (per the WebRTC spec). So the SetDirection(sendonly) triggers the TrackRemoved
             // event, but the pairing was never done because SetDirection() is called before
-            // the received is updated.
+            // the receiver is updated.
 
             // Callbacks must be called on the main thread.
             _peer.InvokeOnAppThread(() =>
--- a/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
@@ -60,15 +60,6 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             get { return (_source as MediaTrackSource); }
             set
             {
-                // If the connection is active, ensure that this is run on the main thread
-                // and doesn't race with the signaling handlers.
-                // If the connection is not awake we cannot use EnsureIsMainAppThread, but
-                // there should be no race on shared state, so skip the check.
-                if (_peer.isActiveAndEnabled)
-                {
-                    _peer.EnsureIsMainAppThread();
-                }
-
                 if (value == null)
                 {
                     if (_source != null)
@@ -173,15 +164,6 @@ namespace Microsoft.MixedReality.WebRTC.Unity
                     throw new ArgumentException("Wrong media kind", nameof(Receiver));
                 }
 
-                // If the connection is active, ensure that this is run on the main thread
-                // and doesn't race with the signaling handlers.
-                // If the connection is not awake we cannot use EnsureIsMainAppThread, but
-                // there should be no race on shared state, so skip the check.
-                if (_peer.isActiveAndEnabled)
-                {
-                    _peer.EnsureIsMainAppThread();
-                }
-
                 if (_receiver != null)
                 {
                     if (_remoteTrack != null)
@@ -193,11 +175,11 @@ namespace Microsoft.MixedReality.WebRTC.Unity
                 _receiver = value;
                 if (_receiver != null)
                 {
+                    _receiver.OnAddedToMediaLine(this);
                     if (_remoteTrack != null)
                     {
                         _receiver.OnPaired(_remoteTrack);
                     }
-                    _receiver.OnAddedToMediaLine(this);
                 }
 
                 // Whatever the change, keep the direction consistent.
@@ -281,6 +263,8 @@ namespace Microsoft.MixedReality.WebRTC.Unity
             if (Transceiver != null)
             {
                 // Avoid races on the desired direction by limiting changes to the main thread.
+                // Note that EnsureIsMainAppThread cannot be used if _peer is not awake, so only
+                // check when there is a transceiver (meaning _peer is enabled).
                 _peer.EnsureIsMainAppThread();
 
                 bool wantsSend = (_source != null);
@@ -354,8 +338,6 @@ namespace Microsoft.MixedReality.WebRTC.Unity
         internal void OnUnpaired(MediaTrack track)
         {
             Debug.Assert(track != null);
-            Debug.Assert(Transceiver != null);
-            Debug.Assert(Transceiver.RemoteTrack == null); // already removed
             // This is called by the TrackRemoved event, which can be fired sometimes even
             // though we did not have any opportunity yet to pair. So only unpair if we did.
             // In details, the case is the answering peer being in sendonly mode, yet created
@@ -372,6 +354,7 @@ namespace Microsoft.MixedReality.WebRTC.Unity
                     bool wasReceiving = _remoteTrack != null;
                     if (wasReceiving)
                     {
+                        Debug.Assert(_remoteTrack == track);
                         _receiver.OnUnpaired(_remoteTrack);
                     }
                 }
--- a/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
@@ -37,6 +37,8 @@ namespace Microsoft.MixedReality.WebRTC.Unity
 
         /// <summary>
         /// Media source producing the media to send through the transceiver attached to this media line.
+        /// </summary>
+        /// <remarks>
         /// This must be an instance of a class derived from <see cref="AudioTrackSource"/> or <see cref="VideoTrackSource"/>
         /// depending on whether <see cref="MediaKind"/> is <see xref="Microsoft.MixedReality.WebRTC.MediaKind.Audio"/>
         /// or <see xref="Microsoft.MixedReality.WebRTC.MediaKind.Video"/>, respectively.
@@ -50,12 +52,23 @@ namespace Microsoft.MixedReality.WebRTC.Unity
         /// If <see cref="Transceiver"/> is valid, that is a first session negotiation has already been completed,
         /// then changing this value raises a <see xref="WebRTC.PeerConnection.RenegotiationNeeded"/> event on the
         /// peer connection of <see cref="Transceiver"/>.
-        /// </summary>
+        ///
+        /// Must be changed on the main thread.
+        /// </remarks>
         public MediaTrackSource Source
         {
             get { return (_source as MediaTrackSource); }
             set
             {
+                // If the connection is active, ensure that this is run on the main thread
+                // and doesn't race with the signaling handlers.
+                // If the connection is not awake we cannot use EnsureIsMainAppThread, but
+                // there should be no race on shared state, so skip the check.
+                if (_peer.isActiveAndEnabled)
+                {
+                    _peer.EnsureIsMainAppThread();
+                }
+
                 if (value == null)
                 {
                     if (_source != null)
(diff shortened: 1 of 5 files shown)


libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -50,12 +52,23 @@ public class MediaLine : ISerializationCallbackReceiver
         /// If <see cref="Transceiver"/> is valid, that is a first session negotiation has already been completed,
         /// then changing this value raises a <see xref="WebRTC.PeerConnection.RenegotiationNeeded"/> event on the
         /// peer connection of <see cref="Transceiver"/>.
-        /// </summary>
+        ///
+        /// Must be changed on the main thread.
```
djee-ms, line 56: Avoid confusion with _e.g._ the UWP main UI thread:
```suggestion
        /// Must be changed on the main Unity app thread.
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -50,12 +52,23 @@ public class MediaLine : ISerializationCallbackReceiver
...
             set
             {
+                // If the connection is active, ensure that this is run on the main thread
+                // and doesn't race with the signaling handlers.
+                // If the connection is not awake we cannot use EnsureIsMainAppThread, but
+                // there should be no race on shared state, so skip the check.
+                if (_peer.isActiveAndEnabled)
+                {
+                    _peer.EnsureIsMainAppThread();
+                }
```
djee-ms, line 70: This makes little sense, you can only access `isActiveAndEnabled` from the main Unity app thread, otherwise you get an exception I believe.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -141,36 +156,48 @@ public MediaTrack SenderTrack
         /// If <see cref="Transceiver"/> is valid, that is a first session negotiation has already been conducted,
         /// then changing this value raises a <see xref="WebRTC.PeerConnection.RenegotiationNeeded"/> event on the
         /// peer connection of <see cref="Transceiver"/>.
-        /// </summary>
+        ///
+        /// Must be changed on the main thread.
```
djee-ms, line 160:
```suggestion
        /// Must be changed on the main Unity app thread.
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -141,36 +156,48 @@ public MediaTrack SenderTrack
...
                 {
-                    if (_receiver != null)
-                    {
-                        _receiver.OnRemovedFromMediaLine(this);
-                        _receiver = null;
-                    }
+                    return;
                 }
-                else
+                if (value!= null && value.MediaKind != MediaKind)
```
djee-ms, line 171: nit:
```suggestion
                if (value != null && value.MediaKind != MediaKind)
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -141,36 +156,48 @@ public MediaTrack SenderTrack
...
+                }
+
+                // If the connection is active, ensure that this is run on the main thread
+                // and doesn't race with the signaling handlers.
+                // If the connection is not awake we cannot use EnsureIsMainAppThread, but
+                // there should be no race on shared state, so skip the check.
+                if (_peer.isActiveAndEnabled)
+                {
+                    _peer.EnsureIsMainAppThread();
+                }
```
djee-ms, line 183:
```suggestion
                _peer.EnsureIsMainAppThread();
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -141,36 +156,48 @@ public MediaTrack SenderTrack
...
-                        if (_receiver != null)
-                        {
-                            _receiver.OnRemovedFromMediaLine(this);
-                        }
-
-                        _receiver = value;
-                        _receiver.OnAddedToMediaLine(this);
+                        _receiver.OnPaired(_remoteTrack);
                     }
+                    _receiver.OnAddedToMediaLine(this);
```
djee-ms, line 200: nit: reverse order for consistency with the unpair/remove above?
```
_receiver.OnAddedToMediaLine(this);
if (_remoteTrack != null) {
    _receiver.OnPaired(_remoteTrack);
}
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -248,10 +276,13 @@ internal MediaLine(PeerConnection peer, MediaKind kind)
             _mediaKind = kind;
         }
 
-        protected void UpdateTransceiverDesiredDirection()
+        private void UpdateTransceiverDesiredDirection()
         {
             if (Transceiver != null)
             {
+                // Avoid races on the desired direction by limiting changes to the main thread.
+                _peer.EnsureIsMainAppThread();
```
djee-ms, line 284: Move out if `if()`? Having a precondition on the function that depends on some other property is weird. Do we really need to call that method with `Transceiver == null` from another thread than the main Unity app thread?

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -294,30 +325,30 @@ internal void DetachTrack()
             }
         }
 
-        internal void UpdateReceiver()
+        internal void UpdateReceiverPairingIfNeeded()
         {
+            Debug.Assert(Transceiver != null);
+
+            // Callbacks must be called on the main thread.
```
djee-ms, line 332:
```suggestion
            // Callbacks must be called on the main Unity app thread.
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -332,11 +363,20 @@ internal void OnUnpaired(MediaTrack track)
             // mode (per the WebRTC spec). So the SetDirection(sendonly) triggers the TrackRemoved
             // event, but the pairing was never done because SetDirection() is called before
             // the received is updated.
```
djee-ms, line 365:
```suggestion
            // the receiver is updated.
```

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
```diff
@@ -332,11 +363,20 @@ internal void OnUnpaired(MediaTrack track)
...
-            if (_isReceiverPaired)
+
+            // Callbacks must be called on the main thread.
+            _peer.InvokeOnAppThread(() =>
```
djee-ms, line 368: I'd move the 2 asserts related to `Transceiver` above in that app thread delegate (can't leave comment on them, scroll up), that ensures we're not accessing them concurrently, no?

libs/unity/library/Runtime/Scripts/Media/WorkQueue.cs
```diff
@@ -37,7 +44,14 @@ public void EnsureIsMainAppThread()
...
         public void InvokeOnAppThread(Action action)
         {
-            _mainThreadWorkQueue.Enqueue(action);
+            if (_mainAppThread != null && IsMainAppThread)
+            {
+                action();
+            }
+            else
+            {
+                _mainThreadWorkQueue.Enqueue(action);
```
djee-ms, line 53: Document that calls before `Awake` will enqueue until awoken even if on the correct thread, and not run synchronously?

libs/unity/library/Tests/Runtime/MediaLineTests.cs
```diff
@@ -175,10 +175,12 @@ public void SetSource()
 
             // Set an invalid source (wrong media kind)
             Assert.Throws<ArgumentException>(() => mediaLine.Source = pc_go.AddComponent<DummyAudioSource>());
+
+            yield return null;
```
djee-ms, line 179: Why? Need comment, especially as this is the last line of the test.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
```diff
@@ -219,6 +221,8 @@ public void SetReceiver()
 
             // Set an invalid source (wrong media kind)
             Assert.Throws<ArgumentException>(() => mediaLine.Receiver = pc_go.AddComponent<AudioReceiver>());
+
+            yield return null;
```
djee-ms, line 225: Same

libs/unity/library/Tests/Runtime/VideoSenderTests.cs
```diff
@@ -80,6 +81,8 @@ public void CaptureStartsOnActivate()
             // TODO fix after media source refactoring
             //// Confirm the source stops streaming
             //Assert.IsFalse(source.IsStreaming);
+
+            yield return null;
```
djee-ms, line 85: Same

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: The issue here is that `EnsureIsMainAppThread` will throw if called before the PC is awake (which will happen normally - it happens in tests now - if the PC is added to an inactive object), but setting Source/Receiver on a "sleeping" PC should be fine.

If `isActiveAndEnabled` throws consistently we can use that in `EnsureIsMainAppThread` rather than checking the (possibly uninitialized) local thread variable.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
fibann: This needs to be run on the Unity app thread, and it needs at least one `yield return` to be a valid coroutine, right? Is there a nicer way to make a valid UnityTest?

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
djee-ms: I don't know if it throws consistently, I would rely on consistently doing something forbidden by Unity and expecting Unity will catch that and throw.

In any case the current code doesn't work, because either `isActiveAndEnabled` is accessed from a background thread and it's garbage/inconsistent/throws, so the test `EnsureIsMainAppThread()` might not even run, or it is from the main Unity app thread, and the test `EnsureIsMainAppThread()` is useless.

If there's no better way, then just delete everything and don't test. Nobody expects users to be able to modify Unity component properties from a background thread anyway, so it's mostly about making sure our code does the right thing.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: Conflating into https://github.com/microsoft/MixedReality-WebRTC/pull/458#discussion_r454987005

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: The purpose of that code was to throw if called from outside the app thread and not throw when called on the app thread, which it served perfectly. `isActiveAndEnabled` does throw, it seems. I could have just called that and discarded the result, but I felt the above seemed more readable, even though I agree it looks weird at an accurate analysis.

Anyway I agree we shouldn't rely on a specific behavior of `isActiveAndEnabled`, so I have rearranged/removed the checks.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: Fixed

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: I replaced them with an assert on the track which is more relevant.

libs/unity/library/Runtime/Scripts/Media/MediaLine.cs
fibann: `EnsureIsMainAppThread` will throw if the peer is not awake, even if the thread is correct. So we are checking only when we know that it is awake (if it isn't, the method will be a nop so no one cares). Added a comment.

libs/unity/library/Runtime/Scripts/Media/WorkQueue.cs
fibann: Fixed.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
djee-ms: I guess not. Can you document that this is the reason please? Because I would have not thought about it.

PR #459, opened 2020-07-15
Explain yield return null

fibann:


fibann:
#458 somehow didn't fell like merging this.

PR #462, opened 2020-07-17
[Unity] Fix several prerelease bugs

djee-ms:
- Don't write BOM when updating package.json otherwise this will prevent Unity from loading the package.
- Upgrade TestNetCoreConsole to .NET Core 3.1 to avoid Azure agent warning about 3.0 being EOL/deprecated.
- Move some test utility classes used in tests into the main Unity library so it's independent of the samples one.
- Fix some Editor test not correctly waiting for the end of the SDP exchange.
- Fix some `.dll.meta` file producing some duplicate deploy on Android due to eager globbing.


libs/unity/library/Tests/Editor/EditorTests.cs
```diff
@@ -57,33 +55,27 @@ public async void PeerConnectionLocalConnect()
...
                     pc1.IceCandidateReadytoSend += (IceCandidate candidate) => pc2.AddIceCandidate(candidate);
                     pc2.IceCandidateReadytoSend += (IceCandidate candidate) => pc1.AddIceCandidate(candidate);
 
                     // Connect
-                    var conn1 = new ManualResetEventSlim(initialState: false);
-                    var conn2 = new ManualResetEventSlim(initialState: false);
-                    pc1.Connected += () => conn1.Set();
-                    pc2.Connected += () => conn2.Set();
                     pc1.CreateOffer();
-                    WaitForSdpExchangeCompleted(conn1, conn2);
```
fibann, line 86: Does this mean that `Connected` is raised _before_ the SDP exchange is completed? This is a bit surprising, does it mean that we can't rely on `Connected` to know when transceivers etc are ready?

djee-ms: > Does this mean that Connected is raised before the SDP exchange is completed? 

Yes, it has always been like that.

> does it mean that we can't rely on Connected to know when transceivers etc are ready?

No. Connected means the connection is established, which means the transports are writable, which means media can start flowing between the two peers. That does not mean that the negotiation is done. This may sound counter-intuitive, but this is due to the fact the connection is asymmetric, and is designed to allow warm startup by allowing the callee to start sending the media before the exchange is even completed, so that when the caller receives the answer and accepts it then it possibly receives immediately after (or has already started receiving, and has been discarding it for now because the exchange was pending) the media and can immediately use it, instead of waiting for something else to tell the callee that the caller received its answer (which by the way I am not sure even exists). Also note that from the point of view of the callee, once it created a local answer, even before it starts sending it to the caller, the exchange is done (since there's no receive ACK from the caller), whereas for the caller the exchange is still pending until it receives and applies the answer. But in testing we want to wait for both, which in practice means waiting for the caller to receive and apply the answer.

PR #467, opened 2020-07-21
Connect MediaLines only on PeerConnection awake

### LINKED ISSUE #446: Fix Unity track sources behavior on Enable/Disable
We aren't creating a track if we enable the source after plugging it into a MediaLine, we are unplugging the source on Disable which is weird, `OnAddedToMediaLine` asserts when editing MediaLines, `MediaLine.CreateSenderIfNeeded` is accessing `isActiveAndEnabled` from arbitrary threads, and probably more. We should define the behavior well and lock access to shared state.

fibann:
Connecting in OnAfterDeserialize wasn't correct:
- source/receiver might not be deserialized yet when OnAfterDeserialize is
called
- OnAfterDeserialize is called multiple times for the same object, which
triggered asserts
- OnAfterDeserialize is not called on the update thread (which isn't great
for calling an external callback)

Fixes #446.

### DIFF
--- a/libs/unity/library/Tests/Runtime/MediaLineTests.cs
+++ b/libs/unity/library/Tests/Runtime/MediaLineTests.cs
@@ -141,7 +141,6 @@ namespace Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime
         {
             // Create the peer connections
             var pc_go = new GameObject("pc1");
-            pc_go.SetActive(false); // prevent auto-activation of components
             var pc = pc_go.AddComponent<PeerConnection>();
 
             // Create some video track sources
@@ -193,7 +192,6 @@ namespace Microsoft.MixedReality.WebRTC.Unity.Tests.Runtime
         {
             // Create the peer connections
             var pc_go = new GameObject("pc1");
-            pc_go.SetActive(false); // prevent auto-activation of components
             var pc = pc_go.AddComponent<PeerConnection>();
 
             // Create some video track sources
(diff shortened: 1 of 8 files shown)


libs/unity/library/Tests/Runtime/MediaLineTests.cs
```diff
@@ -235,5 +233,66 @@ public IEnumerator SetReceiver()
...
+
+            // Add a media line while inactive.
+            VideoTrackSource source1 = go.AddComponent<UniformColorVideoSource>();
+            VideoReceiver receiver1 = go.AddComponent<VideoReceiver>();
+            MediaLine ml1 = pc.AddMediaLine(MediaKind.Video);
+            ml1.Source = source1;
+            ml1.Receiver = receiver1;
+
+            // Media lines have not been set yet.
+            Assert.IsEmpty(source1.MediaLines);
```
djee-ms, line 253: That's quite unintuitive. So I do `AddMediaLine()` and sometimes it adds it to `MediaLines` and sometimes not?

libs/unity/library/Tests/Runtime/MediaLineTests.cs
```diff
@@ -235,5 +233,66 @@ public IEnumerator SetReceiver()
...
+            pc.enabled = false;
+
+            // Add a media line while disabled.
+            VideoReceiver receiver3 = go.AddComponent<VideoReceiver>();
+            MediaLine ml3 = pc.AddMediaLine(MediaKind.Video);
+            ml3.Source = source1;
+            ml3.Receiver = receiver3;
+
+            // Media line #3 is connected.
+            Assert.AreEqual(source1.MediaLines[2], ml3);
```
djee-ms, line 282: And with active but not enabled it does add to `MediaLines`? This is so confusing.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
fibann: TBC `AddMediaLine` always adds a media line to `PeerConnection.MediaLines`. I assume you are talking about `mediaLine.Source = source`.

The problem is that:
- setting a connection in the editor won't call `OnAddedToMediaLine`. I am not sure if there is a reliable way to do it. This implies that there is inevitably going to be a state (fresh out of load, not awake) where the object is accessible but `mediaLine` is not in `source.MediaLines`
- we can't serialize the connection since `MediaLine` is not a Unity object and therefore Unity won't let you share a reference to the same `MediaLine` across two objects

`OnAfterDeserialize` seemed to suit our case except for the problems above. So is seems some inconsistency in state is inevitable. `Awake` seems to be the next best option. Any better ideas?

libs/unity/library/Tests/Runtime/MediaLineTests.cs
fibann: _Awake_ but not enabled. This is kind of arbitrary, we could do it the other way (only call the callbacks if the peer is enabled rather than if it's awake). I thought it might be better to limit the cases where `MediaLines` is not populated to when the peer is not awake which is kind of a niche situation, so this way looked a bit better. No strong feelings about it.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
djee-ms: Ok I see the problem, I have no better solution then.

libs/unity/library/Tests/Runtime/MediaLineTests.cs
djee-ms: See other comment, leave as is.

PR #468, opened 2020-07-21
Fix VideoChatDemo scene

fibann:
References to some UI script had got corrupted along the way


libs/unity/samples/Runtime/Scenes/VideoChatDemo/VideoChatDemo.unity
```diff
@@ -1372,12 +1367,12 @@ MonoBehaviour:
     m_PersistentCalls:
       m_Calls: []
   _mediaLines:
-  - _peer: {fileID: 114452440398351871}
+  - _peer: {fileID: 0}
```
djee-ms, line 1370: revert

PR #479, opened 2020-07-23
Change Windows NuGet package names

djee-ms:
Change the names of the Windows NuGet packages to reflect the fact the packages now contain `mrwebrtc.dll` which exposes a pure C interface, unlike the v1 packages which were exposing a C++ one.


tools/ci/templates/jobs-mrwebrtc-release-pack.yaml
```diff
@@ -29,9 +29,28 @@ jobs:
...
+  - task: PythonScript@0
+    displayName: 'Compute package name'
+    inputs:
+      scriptSource: inline
+      script: |
+        import sys;
+        platform = sys.argv[1];
+        print("Platform : %s" % platform);
+        if (platform == 'Desktop'):
+          packageName = "mrwebrtc"
```
fibann, line 44: nit: should it be `mrwebrtc_win32` to be consistent with the project name?

djee-ms: Looking at nuget.org, native non-UWP packages don't have any suffix. So trying to be consistent with that.

PR #486, opened 2020-07-28
Fix issues with frame ready callbacks

### LINKED ISSUE #151: Redundant conversion to ARGB consumes a lot of CPU (HL2 cannot stream > 720p with hardware H.264)
This is a regression starting from https://github.com/microsoft/MixedReality-WebRTC/commit/ab67d06c5ff3fe216ad63ac58d092773b1ace98a (so impacts releases starting from v1.0.1). It's not incredibly clear what's happening but the best guess is that adding the callbacks triggers the conversion of every frame to RGB, which starves the encoding thread.

This is going to be fixed as part of #152. If the RGB frames are not needed, a simple workaround is to revert the commit above.

fibann:
Fixes #151, a deadlock caused by subscription and frame-ready event trying to acquire the callback lock and the WebRTC worker thread at the same time, wrong behavior when using both I420A and ARGB32 callbacks, and a non-compiling test.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/Interop/VideoTrackSourceInterop.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/Interop/VideoTrackSourceInterop.cs
@@ -11,7 +11,7 @@ namespace Microsoft.MixedReality.WebRTC.Interop
     /// </summary>
     internal class VideoTrackSourceHandle : RefCountedObjectHandle { }
 
-    internal class VideoTrackSourceInterop
+    internal static class VideoTrackSourceInterop
     {
         // TODO - Remove ARGB callbacks, use I420 callbacks only and expose some conversion
         // utility to convert from ARGB to I420 when needed (to be called by the user).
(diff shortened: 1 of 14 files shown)


libs/Microsoft.MixedReality.WebRTC/Interop/VideoTrackSourceInterop.cs
```diff
@@ -42,24 +42,28 @@ internal class VideoTrackSourceInterop
 
         #endregion
 
-        public class InteropCallbackArgs
+        /// Callback arguments to ensure delegates registered with the native layer don't go out of scope.
+        internal static readonly I420AVideoFrameUnmanagedCallback I420AFrameCallback = I420AFrameCallbackFunction;
+        internal static readonly Argb32VideoFrameUnmanagedCallback Argb32FrameCallback = Argb32FrameCallbackFunction;
+
+        // Implemented by video source wrappers to receive frames from the interop layer.
+        internal interface IVideoSource
```
djee-ms, line 50: That's a lot neater than the previous pattern, nice!

libs/Microsoft.MixedReality.WebRTC/LocalVideoTrack.cs
```diff
@@ -259,30 +296,32 @@ public override void Dispose()
...
             // but will be handled at some point anyway, even if the managed instance is gone.
             _nativeHandle.Dispose();
         }
 
-        internal void OnI420AFrameReady(I420AVideoFrame frame)
+        void VideoTrackSourceInterop.IVideoSource.OnI420AFrameReady(I420AVideoFrame frame)
         {
             MainEventSource.Log.I420ALocalVideoFrameReady(frame.width, frame.height);
-            I420AVideoFrameReady?.Invoke(frame);
+            lock(_videoFrameReadyLock)
```
djee-ms, line 312: Did you check the lock was not adding any performance overhead? I don't think so, but since this is the critical path...

libs/mrwebrtc/src/video_frame_observer.h
```diff
@@ -127,6 +127,8 @@ class VideoFrameObserver : public rtc::VideoSinkInterface<webrtc::VideoFrame> {
   /// This is not exclusive and can be used along another I420 callback.
   void SetCallback(Argb32FrameReadyCallback callback) noexcept;
 
+  bool HasCallbacks() const { return i420a_callback_ || argb_callback_; }
```
djee-ms, line 130: nit:
```suggestion
  bool HasAnyCallback() const { return i420a_callback_ || argb_callback_; }
```

tests/Microsoft.MixedReality.WebRTC.Tests/LocalVideoTrackTests.cs
```diff
@@ -488,6 +408,21 @@ public void SimpleExternalArgb32()
...
+            var track_config = new LocalVideoTrackInitConfig
+            {
+                trackName = "custom_i420a"
+            };
+            using (var source = ExternalVideoTrackSource.CreateFromI420ACallback(
+                VideoTrackSourceTests.CustomI420AFrameCallback))
+            using (var track = LocalVideoTrack.CreateFromSource(source, track_config))
+            {
+                VideoTrackSourceTests.TestFrameReadyCallbacks(track);
+            }
```
djee-ms, line 423: Braces please, this is a recipe for mismerge / refactor disaster.

libs/Microsoft.MixedReality.WebRTC/LocalVideoTrack.cs
fibann: I didn't, I would expect it to be in the order of 10-100 ns if uncontested, so irrelevant.

I am inclined to take out this lock anyway, it can cause deadlocks and the only side effect of removing it is that a handler may be called once (at most) after it has been removed. This is an issue with any free-threaded C# event, so IMHO users should be aware of this and handle any necessary synchronization (we should point it out in the comments ofc as it's not obvious). Otherwise for consistency we should implement this pattern in all our free-threaded events which would be impractical.

djee-ms: Alright, that makes sense. If there's not much benefit then let's keep it simple and remove it.

Issue #159, opened 2020-01-03
Missing capture formats from GetVideoCaptureFormatsAsync on HL2


eirikhollis:
On HL2, when calling `PeerConnection.GetVideoCaptureDevicesAsync()`, the only capture device found is the "QC Back Camera".
If the capture device is inserted into `PeerConnection.GetVideoCaptureFormatsAsync(deviceId)`, the only resolutions returned are (W x H x FPS)
```
2272 x 1278 x 30.0000019073486
896 x 504 x 30.0000019073486
2272 x 1278 x 15.0000009536743
896 x 504 x 15.0000009536743
```

Forcing the capture resolution to 720p does indeed work, which indicates that there are missing capture formats returned from `GetVideoCaptureFormatsAsync`.

djee-ms:
The problem is not missing formats, the problem is that either we are missing an argument to specify the video profile, or we should return a collection of all available video profiles and their associated video capture formats. Here the list you get is the list of video capture formats for the default video profile.

djee-ms:
As a **temporary** workaround (and please keep it temporary, because this is an implementation detail, so I won't guarantee this will always work) you can directly call methods of the UWP `MediaCapture` API like [`MediaCapture.FindKnownVideoProfiles()`](https://docs.microsoft.com/en-us/uwp/api/windows.media.capture.mediacapture.findknownvideoprofiles) to list the profiles, and then access [`MediaCaptureVideoProfile.SupportedRecordMedia`](https://docs.microsoft.com/en-us/uwp/api/windows.media.capture.mediacapturevideoprofile.supportedrecordmediadescription) to get the capture formats.

eirikhollis:
Ah, I understand. Thanks for the suggestion for the temporary workaround.

eirikhollis:
As a suggestion, I would probably have video profile as an optional parameter, such as:
```
GetVideoCaptureFormatsAsync(string deviceId, VideoProfileKind videoProfile = 0);
```
where if you do not specify a video profile (Unspecified), it will return all formats and their associated video profile, but if you set a specific profile, then it will only return the formats related to that profile.

djee-ms:
I had a quick look at it and it's not a trivial change because although we expose the profile ID for capture, the enumeration is still done generically via the Google API which doesn't have any concept of video profile. So this requires some change to transition to explicit Media Foundation based format enumeration

eirikhollis:
[...]

PR #487, opened 2020-07-29
Enumerate video profiles and filter capture formats

djee-ms:
Add the ability to enumerate the video profiles of a video capture device, and
to retrieve the video capture formats for any profile. This fixes an issue
where only video capture formats for the default profile were returned, by
allowing the user to specifically enumerate profiles and their associated
capture formats.

Bug: #159

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/DeviceVideoTrackSource.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/DeviceVideoTrackSource.cs
@@ -178,7 +178,7 @@ namespace Microsoft.MixedReality.WebRTC
         /// 
         /// This is equivalent to:
         /// <code>
-        /// GetCaptureProfilesAsync(deviceId, VideoProfileKind.Unknown);
+        /// GetCaptureProfilesAsync(deviceId, VideoProfileKind.Unspecified);
         /// </code>
         /// </remarks>
         /// <seealso cref="GetCaptureProfilesAsync(string, VideoProfileKind)"/>
--- a/libs/Microsoft.MixedReality.WebRTC/DeviceVideoTrackSource.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/DeviceVideoTrackSource.cs
@@ -76,9 +76,20 @@ namespace Microsoft.MixedReality.WebRTC
         /// </summary>
         /// <remarks>
         /// This is compared by strict equality, so is best left unspecified or to an exact value
-        /// retrieved by <see cref="DeviceVideoTrackSource.GetCaptureFormatsAsync"/>.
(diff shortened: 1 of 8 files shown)


libs/mrwebrtc/include/interop_api.h
```diff
@@ -195,22 +258,45 @@ using mrsVideoCaptureFormatEnumCallback = void(
...
-/// Enumerate the video capture formats asynchronously.
+/// Asynchronously enumerate the capture formats of a video capture device.
+///
+/// The capture formats are enumerated for the given video capture device:
+/// - If the device does not support video profiles, |profile_id| and
+/// |profile_kind| are both ignored, and all capture formats supported by the
+/// specified device are enumerated.
+/// - If the device supports video profiles, enumeration depends on the values
+/// of the profile ID and kind. The profile ID string |profile_id| can be null
+/// or an empty string, meaning no specified profile ID. The profile kind
```
fibann, line 269: `null` as "unspecified" is standard, but empty string is a bit weird. I'd rather just check for null here and treat empty string as any other string.

libs/Microsoft.MixedReality.WebRTC/DeviceVideoTrackSource.cs
```diff
@@ -150,7 +161,96 @@ public static Task<List<VideoCaptureDevice>> GetCaptureDevicesAsync()
...
+        /// capture formats of, as retrieved from the <see cref="VideoCaptureDevice.id"/> field of
+        /// a capture device enumerated with <see cref="GetCaptureDevicesAsync"/>.</param>
+        /// <returns>The list of available video profiles for the specified video capture device.</returns>
+        /// <remarks>
+        /// If the video capture device does not support video profiles, the function succeeds
+        /// and returns an empty list.
+        /// 
+        /// This is equivalent to:
+        /// <code>
+        /// GetCaptureProfilesAsync(deviceId, VideoProfileKind.Unknown);
```
fibann, line 181:
```suggestion
        /// GetCaptureProfilesAsync(deviceId, VideoProfileKind.Unspecified);
```

libs/mrwebrtc/include/interop_api.h
djee-ms: That makes sense indeed.

djee-ms: Actually this is annoying because we convert to `absl::string_view` so lose the distinction. Also in general I don't think we make much difference between `nullptr` and empty strings, I think we use the `IsStringNullOrEmpty()` helper on most string arguments in the API, and since there's no empty string profile ID then I would be for treating it the same as `nullptr`.

Issue #492, opened 2020-07-31
Video capture format enumeration returns a single format on UWP when specifying a video profile


djee-ms:
Video capture format enumeration returns a single capture format on UWP when specifying a video profile, instead of a complete list. This does not affect other platforms, nor devices not supporting video profiles.

This is due to some internal limitation and bug around the use of MediaCapture vs. WebRTC, which have some different mechanism to select a video capture format when multiple are available. To work around this, when starting capture the implementation filters the list of capture format to a single one to ensure consistency. Unfortunately this code is reused internally when simply listing the available formats.

PR #494, opened 2020-07-31
Work around broken getSupportedFormats() in UWP

### LINKED ISSUE #488: DeviceVideoTrackSource.GetCaptureFormatsAsync crashes on HL2
**Describe the bug**
On latest master branch with newest code for enumerating of devices and profiles, the function `DeviceVideoTrackSource.GetCaptureFormatsAsync` crashes on HL2 ARM. I haven't had a lot of time to test, but I tried with just a `VideoCaptureDevice.id` and also with `VideoCaptureDevice.id` and `VideoProfileKind.VideoConferencing`.

It does not crash in Unity.

**Environment**
 - Platform: UWP
 - Architecture: ARM
 - Unity version: 2019.4.5
 - Target device: HoloLens 2

djee-ms:
Implement a custom video capture format enumeration based on
MediaCapture instead of relying on the broken webrtc-uwp-sdk
implementation, which is difficult to fix without some large refactor of
the video capturer.

This incidentally fixes #488 where a nullptr string was sometimes passed
to the std::string constructor leading to a crash. The offending code is
not used anymore.

Bug: #488, #492

### DIFF
--- a/libs/mrwebrtc/src/media/device_video_track_source.cpp
+++ b/libs/mrwebrtc/src/media/device_video_track_source.cpp
@@ -7,9 +7,45 @@
 #include "media/device_video_track_source.h"
 #include "video_track_source_interop.h"
 
+#if defined(WINUWP)
+#include <winrt/Windows.Foundation.Collections.h>
+#include <winrt/Windows.Media.Capture.h>
+#endif
+
 namespace {
 
 using namespace Microsoft::MixedReality::WebRTC;
+
+#if defined(WINUWP)
+uint32_t FourCCFromMFSubType(winrt::hstring const& subtype) {
+  using namespace winrt::Windows::Media::MediaProperties;
+  uint32_t fourcc;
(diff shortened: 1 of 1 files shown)


djee-ms:
@eirikhollis FYI just confirmed by adding this loop somewhere and running VideoChatDemo on HL2 that it doesn't crash, so fixes #488:

```cs
UnityEngine.Debug.Log("Listing devices...");
var devices = await DeviceVideoTrackSource.GetCaptureDevicesAsync();
foreach (var device in devices)
{
    UnityEngine.Debug.Log("Device: " + device.id);
    var formats = await DeviceVideoTrackSource.GetCaptureFormatsAsync(device.id);
    foreach (var format in formats)
    {
        UnityEngine.Debug.Log($"+ Format: {format.width}x{format.height}@{format.framerate}");
    }
}
```

Also tested with TestAppUWP that when specifying a video profile then all its capture formats are listed, which fixes #492.

PR #504, opened 2020-08-05
Update helloworld-cs-mediatracks-uwp.md

KarenConsoli:
There was a command that was not correct in the code example. Please see the difference.

image


djee-ms:
Merged, thanks @KarenConsoli!

Issue #124, opened 2019-11-11
Unity crashes on Windows when permission to the microphone is denied


HugoKampsSB:
Hi!

I have a working RTC setup in Unity on my Windows laptop. One of my colleagues cloned the project but when pressing Play Unity closed down (even without the crash helper), without any logs. After some digging I found out it was due to the following error:

`Failed to complete start recognition system (hr = 0x80070005)`

Using this error I traced it to the Microphone access on this computer being disabled for all apps. 
![image](https://user-images.githubusercontent.com/37140671/68577396-4fce7480-0470-11ea-9e6e-1312716aa454.png)

After enabling it, it worked perfectly again :) Sadly I don't have any crash logs, because Unity closed down in a different way. 

I hope this helps!

djee-ms:
Unfortunately there is nothing we can do about that. When the internal Google implementation is initializing its audio module, on the very first call to any library API, it tries to find a microphone (any one) and will fail if none found. Right after that the caller asserts on the audio module pointer being NULL. This is admittedly a bug, or at least a questionable design decision that a microphone must be available, unless I am missing something here.

I tried adding the following checks, and they all failed to prevent the crash:
- Check in `LocalAudioSource.cs` that `Microphone.devices.Length` is non-zero; unfortunately Unity will report all available microphones, even if access to them is forbidden.
- Check `Application.HasUserAuthorization(UserAuthorization.WebCam)`, which incorrectly returns `true` (probably mapped to UWP capabilities, and returns always `true` on Desktop).

I am not happy about this answer, but unfortunately it seems the only way forward is to log a bug with Google and see what their devs have to say about this. Which may anyway pose some issue to integrate, because I have concerns they will not be willing to backport the fix to M71 (too old).

But more than that, the CoreAudio implementation of WebRTC is failing if **any** device fails to initialize. In particular in this case **all** devices are failing with `E_ACCESSDENIED` as seen in the debugger logs in Debug. I cannot find the reference again, but I already observed that questionable behavior and tried to report it, only to find that there was already a bug closed as "Won't fix" because the CoreAudio(1) is being deprecated in favor of a new CoreAudio2 rewrite, so devs do not want to spend resources on fixing the legacy one. Which makes it even less likely to get a fix in for the current issue. And CoreAudio2 is currently marked as experimental, so I am not very keen on switching to that for production code.

[...]

PR #507, opened 2020-08-05
Switch to ADM2 on Windows Desktop

djee-ms:
Switch audio implementation on Windows Desktop to the new Core Audio
based audio device module (ADM2), which avoids a crash in ADM1 when one
device is not supported by that module (even if unused), like Azure
Kinect DK.

Add a new function |mrsLibraryUseAudioDeviceModule2()| to enable users
to switch back to ADM1 in case there would be issues with ADM2, as
support and maturity of ADM2 are unclear, and the module is not always
well tested, though is the recommended way from Google going forward.

Bug: #124

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/Library.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/Library.cs
@@ -7,6 +7,23 @@ using System;
 namespace Microsoft.MixedReality.WebRTC
 {
     /// <summary>
+    /// Audio device module for Windows Desktop platform.
+    /// </summary>
+    public enum AudioDeviceModule : byte
+    {
+        /// <summary>
+        /// Legacy audio device module (ADM1); not recommended unless there is an issue
+        /// with the new default one.
+        /// </summary>
+        LegacyModule = 1,
+
+        /// <summary>
+        /// New CoreAudio based audio device module (ADM2). This is the default.
+        /// </summary>
+        DefaultModule = 2
+    }
+
+    /// <summary>
     /// Container for library-wise global settings of MixedReality-WebRTC.
     /// </summary>
     public static class Library
@@ -21,6 +38,26 @@ namespace Microsoft.MixedReality.WebRTC
         public static uint ReportLiveObjects()
         {
             return Utils.LibraryReportLiveObjects();
+        }
+
+        /// <summary>
+        /// Select the audio device module to use on Windows Desktop.
+        /// </summary>
+        /// <param name="adm">The module to use.</param>
+        /// <remarks>
+        /// The ADM must be configured before any peer connection is created/initialized;
+        /// otherwise an <see cref="InvalidOperationException"/> is raised.
+        /// 
+        /// This has no effect on UWP or non-Windows platforms.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        /// The peer connection factory was already initialized, therefore the audio device
+        /// module cannot be changed anymore.
+        /// </exception>
+        public static void UseAudioDeviceModule(AudioDeviceModule adm)
+        {
+            uint res = Utils.LibraryUseAudioDeviceModule(adm);
+            Utils.ThrowOnErrorCode(res);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


libs/mrwebrtc/include/interop_api.h
```diff
@@ -24,6 +24,24 @@ enum class mrsOptBool : int8_t { kTrue = -1, kFalse = 0, kUnset = 0b01010101 };
...
+/// https://github.com/microsoft/MixedReality-WebRTC/issues/124
+///
+/// This function needs to be called before the peer connection factory is
+/// initialized, and therefore before any other library call, as most calls will
+/// initialize the peer connection factory. This cannot be toggled ON and OFF
+/// after that, and will fail if called too late.
+///
+/// This has no effect on UWP and non-Windows platforms, and will always
+/// succeed if called appropriately.
+MRS_API mrsResult MRS_CALL mrsLibraryUseAudioDeviceModule2(bool use) noexcept;
```
fibann, line 43: Pass an enum instead? E.g. `enum class AudioDeviceModule { ADM1, ADM2 }`

djee-ms: Fixed.

Also added missing C# projection.

libs/Microsoft.MixedReality.WebRTC/Library.cs
```diff
@@ -6,6 +6,23 @@
...
+        /// <summary>
+        /// Legacy audio device module (ADM1); not recommended unless there is an issue
+        /// with the new default one.
+        /// </summary>
+        LegacyModule = 1,
+
+        /// <summary>
+        /// New CoreAudio based audio device module (ADM2). This is the default.
+        /// </summary>
+        DefaultModule = 2
```
fibann, line 23: nit: any reason why the enum values don't have the same name here and in the C API?

libs/mrwebrtc/include/interop_api.h
```diff
@@ -24,6 +24,36 @@ enum class mrsOptBool : int8_t { kTrue = -1, kFalse = 0, kUnset = 0b01010101 };
...
+};
+
+/// Enable or disable the use of the new Audio Device Module (ADM2) on Windows
+/// Desktop.
+///
+/// This mainly enables working around some unsupported devices plugged in which
+/// make the legacy module (ADM1) fail its initialization even if said devices
+/// are not later used. Therefore ADM2 is used by default, and this function is
+/// mainly used to disable it and fall back to the legacy ADM1 if needed.
+/// https://github.com/microsoft/MixedReality-WebRTC/issues/124
```
fibann, line 45: Rewrite the comment now that we pass an enum.

> This mainly enables working around some unsupported devices plugged in which make the legacy module (ADM1) fail its initialization

This reads as if _calling this function_ enables working around unsupported devices - but it's the opposite. It should probably be something like "ADM2 is the default. ADM1 is deprecated as it fails initialization if etc.... but if you want you can force its usage by calling this function".

libs/Microsoft.MixedReality.WebRTC/Library.cs
djee-ms: No good ones; I changed my mind when doing the C# projection and thought it was better to be more explicit about what to expect from those (default/legacy) than having some history-related meaningless names (ADM1/ADM2) that do not convey any information for most users. I can change the C API ones too if you prefer.

fibann: In general I'd prefer explicit naming, what's considered "legacy" and "default" might change over time, but this is not a likely issue in this case, so we can go with these. As long as they match - them being different seems to suggest that they have a reason to be different.

libs/mrwebrtc/include/interop_api.h
djee-ms: Fixed, and renamed enum values to match C# too.

libs/Microsoft.MixedReality.WebRTC/Library.cs
djee-ms: Fixed.

JacopoMarullo:
Is this working for UWP x64 for Windows 10 apps?

djee-ms:
No, this is Windows Desktop, not UWP.

PR #521, opened 2020-08-07
Improve build.py

djee-ms:
- Ensure build.py is run from Python 3.x and not an earlier version,
before using 3.x features that make the script throw some cryptic error
- Add encoding to remove warning when running with Python 2.7
- Default to all targets/cpus if not specified, to avoid some cryptic
error if any of those is not provided

### DIFF
--- a/tools/build/build.py
+++ b/tools/build/build.py
@@ -1,12 +1,22 @@
 #!/usr/bin/env python
+# coding=utf-8
+#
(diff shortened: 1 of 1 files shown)


djee-ms:
I was trying to see if we need to port the fix in #518, but we don't invoke MSBuild directly on undock. Instead I found a couple of minor improvements.

PR #524, opened 2020-08-07
Handle OnDisable while resource is initializing

### LINKED ISSUE #464: Disable only works on PeerConnection/MediaTrackSource if object has finished initializing
Initialization in PeerConnection/sources is asynchronous. Currently OnDisable doesn't check if there is an initialization in progress and won't do anything in that case.

fibann:
If a component containing a WebRTC resource with async initialization is
disabled before the resource has finished initializing, now we block the
OnDisable call and properly dispose the resource.

Refactored permission requests on Android to use async methods, so they
integrate better with the rest of the code.

Fixes #464.

### DIFF
--- a/libs/unity/library/Runtime/Scripts/AsyncInitHelper.cs.meta
+++ b/libs/unity/library/Runtime/Scripts/AsyncInitHelper.cs.meta
@@ -5,7 +5,7 @@ MonoImporter:
   serializedVersion: 2
   defaultReferences: []
   executionOrder: 0
-  icon: {instanceID: 0}
+  icon: {fileID: 2800000, guid: f845cb707dda5c84594956f49655100f, type: 3}
   userData: 
   assetBundleName: 
   assetBundleVariant: 
--- a/libs/unity/library/Runtime/Scripts/AsyncInitHelper.cs.meta
+++ b/libs/unity/library/Runtime/Scripts/AsyncInitHelper.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: d32d9bb3d083a194abff0902a5ce56a8
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
@@ -11,7 +11,7 @@ using UnityEngine.Android;
 #endif
 
 #if UNITY_WSA && !UNITY_EDITOR
-using Windows.Media.Capture;
+using global::Windows.Media.Capture;
 #endif
 
 namespace Microsoft.MixedReality.WebRTC.Unity
@@ -46,11 +46,9 @@ namespace Microsoft.MixedReality.WebRTC.Unity
 
         protected async Task<WebRTC.AudioTrackSource> InitAsync()
         {
-
-
             // Continue the task outside the Unity app context, in order to avoid deadlock
             // if OnDisable waits on this task.
-            bool accessGranted = await RequestAccessAsync().ConfigureAwait(false);
+            bool accessGranted = await RequestAccessAsync().ConfigureAwait(continueOnCapturedContext: false);
             if (!accessGranted)
             {
                 return null;
--- a/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
@@ -121,7 +121,11 @@ namespace Microsoft.MixedReality.WebRTC.Unity
 
         protected void OnDisable()
         {
-            _initHelper.AbortInitTask();
+            /// Wait synchronously for the end of the initialization task, so that after the component
+            /// has been disabled the device is released and free to be used again.
+            /// Note that the initialization task needs not to be continued on the app thread, or it
+            /// will deadlock.
+            _initHelper.AbortInitTask().Wait();
             DisposeSource();
         }
 
--- a/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
+++ b/libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Threading.Tasks;
 using Microsoft.MixedReality.WebRTC.Unity.Editor;
 using UnityEngine;
 
@@ -10,12 +11,7 @@ using UnityEngine.Android;
 #endif
 
 #if UNITY_WSA && !UNITY_EDITOR
-using System.Threading.Tasks;
-using global::Windows.UI.Core;
-using global::Windows.Foundation;
-using global::Windows.Media.Core;
-using global::Windows.Media.Capture;
-using global::Windows.ApplicationModel.Core;
+using Windows.Media.Capture;
 #endif
 
 namespace Microsoft.MixedReality.WebRTC.Unity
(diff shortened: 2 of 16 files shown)


libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -10,12 +11,7 @@
 #endif
 
 #if UNITY_WSA && !UNITY_EDITOR
-using System.Threading.Tasks;
-using global::Windows.UI.Core;
-using global::Windows.Foundation;
-using global::Windows.Media.Core;
-using global::Windows.Media.Capture;
-using global::Windows.ApplicationModel.Core;
+using Windows.Media.Capture;
```
djee-ms, line 14: Add `global::` prefix

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -36,77 +32,55 @@ public class MicrophoneSource : AudioTrackSource
...
-            }
-#endif
 
-#if UNITY_WSA && !UNITY_EDITOR
-            // Request access to audio capture. The OS may show some popup dialog to the
-            // user to request permission. This will succeed only if the user approves it.
-            try
+            // Continue the task outside the Unity app context, in order to avoid deadlock
+            // if OnDisable waits on this task.
+            bool accessGranted = await RequestAccessAsync().ConfigureAwait(false);
```
djee-ms, line 53: This is not what you are doing; you are allowing the task to run on any thread, you are not forbidding it to run on the Unity app thread. AFAIK the scheduler is allowed to decide to run on the same thread even with `ConfigureAwait(false)`; it's up to it to decide between this thread and another.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -36,77 +32,55 @@ public class MicrophoneSource : AudioTrackSource
...
             {
                 AutoGainControl = _autoGainControl,
             };
             try
             {
-                AttachSource(await DeviceAudioTrackSource.CreateAsync(initConfig));
+                var createTask = DeviceAudioTrackSource.CreateAsync(initConfig);
+                // Continue the task outside the Unity app context, in order to avoid deadlock
+                // if OnDisable waits on this task.
+                return await createTask.ConfigureAwait(continueOnCapturedContext: false);
```
djee-ms, line 68: Same, there's nothing preventing a deadlock.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
```diff
@@ -36,77 +32,55 @@ public class MicrophoneSource : AudioTrackSource
...
+            }
+        }
+
+        protected void Update()
+        {
+            var result = _initHelper.Result;
+            if (result != null)
+            {
+                AttachSource(result);
             }
```
djee-ms, line 84: Polling every frame for a one-off thing :cry:

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
```diff
@@ -0,0 +1,67 @@
...
+            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
+            {
+                // Request UWP access to audio capture. The OS may show some popup dialog to the
+                // user to request permission. This will succeed only if the user grants permission.
+                try
+                {
+                    // On UWP the app must have the "microphone" capability, and the user must allow microphone
+                    // access. So check that access before trying to initialize the WebRTC library, as this
+                    // may result in a popup window being displayed the first time, which needs to be accepted
+                    // before the microphone can be accessed by WebRTC.
```
djee-ms, line 34: Outdated comment

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
```diff
@@ -0,0 +1,67 @@
...
+        internal static async Task<bool> RequestAccessAsync(StreamingCaptureMode mode)
+        {
+            // Note that the UWP UI thread and the main Unity app thread are always different.
+            // https://docs.unity3d.com/Manual/windowsstore-appcallbacks.html
+            Debug.Assert(!UnityEngine.WSA.Application.RunningOnUIThread());
+            var permissionTcs = new TaskCompletionSource<bool>();
+            UnityEngine.WSA.Application.InvokeOnUIThread(() =>
+            {
+                // Request UWP access to audio capture. The OS may show some popup dialog to the
+                // user to request permission. This will succeed only if the user grants permission.
```
djee-ms, line 28:
```suggestion
                // Request UWP access to audio or video capture. The OS may show some popup dialog to the
                // user to request permission. This will succeed only if the user grants permission.
```

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
```diff
@@ -0,0 +1,67 @@
...
+                        else
+                        {
+                            permissionTcs.SetResult(true);
+                        }
+                    });
+                }
+                catch (Exception ex)
+                {
+                    // Log an error and prevent activation
+                    Debug.LogError($"Audio access failure: {ex.Message}.");
```
djee-ms, line 58:
```suggestion
                    Debug.LogError($"Media access failure: {ex.Message}.");
```

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
+guid: 45dccc74a2e70304bbe6f7be5eaa4c1b
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
```
djee-ms, line 8: +icon

libs/unity/library/Runtime/Scripts/AsyncInitHelper.cs.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
+guid: d32d9bb3d083a194abff0902a5ce56a8
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
```
djee-ms, line 8: +icon

libs/unity/library/Runtime/Scripts/PeerConnection.cs
```diff
@@ -278,26 +271,33 @@ public static Task<IReadOnlyList<VideoCaptureDevice>> GetVideoCaptureDevicesAsyn
...
-            {
-                return RequestAccessAndInitAsync(token);
-            }
-#if UNITY_WSA && !UNITY_EDITOR
-            else
-            {
-                UnityEngine.WSA.Application.InvokeOnUIThread(() => RequestAccessAndInitAsync(token), waitUntilDone: true);
-                return Task.CompletedTask;
-            }
+            // FIXME - Use ADM2 instead, this /maybe/ avoids this.
```
djee-ms, line 292: Did you try with ADM2 if this is still needed? You can forbid access manually from Settings and then run any demo, see if things smoothly error or the app flat out crashes. Last I checked on Desktop with the system-wide authorization (so this is different from the UWP microphone capability we're trying to handle here, though it's the same kind of issue) using ADM2 works around it and will not crash if system-wide microphone access is disabled.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
fibann: So, we want to run `AttachSource` on the Unity app thread. The simplest thing would be to do it at the end of the initialization task, but that's setup so that it doesn't complete on the app thread (to avoid deadlocking with OnDisable). 

Another thing we could do is capturing the Unity `SynchronizationContext` and posting the call to `AttachSource` on it (which btw is probably a better way to run tasks on the app thread than our `WorkQueue`) at the end of the initialization task but _only_ if the component hasn't been disable yet, but this requires non-trivial synchronization. Also using the Unity `SynchronizationContext` directly is not well documented and there might be pitfalls.

There shouldn't be too many device sources in a single app, so I'd say let's leave it like this

djee-ms: Good catch about the `SynchronizationContext`, but no idea if Unity supports that well nor at all. Let's leave as is for now if that works.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
fibann: Discussed in chat. According to https://devblogs.microsoft.com/dotnet/configureawait-faq/, `ConfigureAwait(false)` prevents the continuation from being scheduled on the current SynchronizationContext, thus effectively pushing it into the default thread pool. We agree that unless some unlikely circumstances (Unity app thread belongs to the default thread pool; thread pool has no available threads) are true there should be no deadlock.

libs/unity/library/Runtime/Scripts/Media/MicrophoneSource.cs
fibann: Fixed

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
fibann: Fixed

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
fibann: Fixed

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs
fibann: Fixed

libs/unity/library/Runtime/Scripts/Media/UwpUtils.cs.meta
fibann: Fixed (also below)

libs/unity/library/Runtime/Scripts/PeerConnection.cs
fibann: It seems to be fixed now, it will raise a few `winrt::hresult_error`s but doesn't seem to crash. I removed this block.

Issue #153, opened 2019-12-19
Blocky video from hardware H.264 encoder on HL2 


fibann:
On HL2 some configurations produce a stream that is much "blockier" than the result of the same configuration on HL1 or PC. A possible reason is that the HL2 encoder uses a higher QP range than other encoders by default. WebRTC has a way to specify the QP explicitly but the encoder implementation ignores the value at the moment.

fibann:
Fix https://github.com/webrtc-uwp/webrtc-windows/pull/74 has been integrated by 6d9b8a79abaed3d13dcd7443277bc3859014290c. Now it's possible to mitigate this issue by setting a QP lower than the default (51).

This can be done by:
- creating the SDP offer from the endpoint that is going to _receive_ the video, and
- adding `x-google-max-quantization=<value>` to `PeerConnection.PreferredVideoCodecExtraParams` before creating the offer.

The setting seems to only be applied to the encoder when appearing in the remote description, and at the moment `PreferredVideoCodecExtraParams` is only added to offers (not answers) - thus it only has effect if used in an offer that comes from the receiving endpoint. I suppose we could extend the extra-params mechanism to apply to answers too, and remove this limitation.

Note that setting a low max QP makes the framerate drop in scenes with movement going on. In the scenario I tested (1280x720@30, 1Mbps) a value of 45 seemed to bring some quality improvement while maintaining a stable framerate, but YMMV.

kspark-scott:
Thanks Filippo (@fibann).  We can try this out, though it may take a few days before I have the time to fit it in.  It will require a change to how we do signalling as our calling is not symmetric: calls can be initiated from HoloLens only.  So currently the HoloLens creates the SDP offer and then sends but does not receive video.  If I understand you correctly, we'll have to reverse that and have the call receiver create the offer and the HoloLens create the answer.  We do already have call/answer messages that are separate from the SDP offer/answer so it's not too disruptive, but is more than just adding an extra option to the codec params.

eirikhollis:
@fibann How much of a difference should it make? I'm setting the quantization value with
```
PreferredVideoCodecExtraParams = `"x-google-max-quantization=45"
```
but even with values ranging from 10, 20, 30, ..., 80, I see no difference in blockiness. Is there a way I can verify that it has been set?
CPU usage on HL2 stays consistent as well.

djee-ms:
[...]

PR #528, opened 2020-08-10
Expose H.264 encoder configuration 

fibann:
The configuration can be tweaked to get improved video quality. This is a
mitigation for #153.

Important: using a profile different from (Constrained) Baseline requires
specifying the correct `profile-level-id` parameter in the corresponding SDP
codec entry for other peers to correctly accept/reject the codec. This change
does not change SDP offer from the default (unspecified, equivalent to
Baseline). If using a higher profile, you must ensure that all peers support
encoding/decoding at that profile, or the behavior will be undefined.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
@@ -291,5 +291,25 @@ namespace Microsoft.MixedReality.WebRTC.Interop
             }
             return string.Join(";", streamIDs.ToArray());
         }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct H264Config
+        {
+            internal int Profile;
+            internal int RcMode;
+            internal int MaxQp;
+            internal int Quality;
+
+            internal H264Config(PeerConnection.H264Config config)
+            {
+                Profile = (int)config.Profile;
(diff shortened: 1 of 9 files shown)


libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
```diff
@@ -291,5 +291,25 @@ public static string EncodeTransceiverStreamIDs(List<string> streamIDs)
...
+            internal int RcMode;
+            internal int MaxQp;
+            internal int Quality;
+
+            internal H264Config(PeerConnection.H264Config config)
+            {
+                Profile = (int)config.Profile;
+                RcMode = config.RcMode.HasValue ? (int)config.RcMode : -1;
+                MaxQp = config.MaxQp.HasValue ? (int)config.MaxQp : -1;
+                Quality = config.Quality.HasValue ? (int)config.Quality : -1;
```
djee-ms, line 308: nit: Use `GetValueOrDefault(-1)` instead for clarity?

djee-ms:
That review will break `master` if you don't integrate the upstream change at the same time.

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
fibann: Fixed

PR #534, opened 2020-08-14
Use custom AudioMixer on UWP

### LINKED ISSUE #371: Prevent automatic remote audio output on UWP
The custom mixer that is already used on other platforms should be passed through the WinRT wrapper.

Note this could be automatically fixed by undocking from webrtc-uwp as well.

fibann:
Fixes #371.

### DIFF
--- a/libs/mrwebrtc/src/interop/global_factory.cpp
+++ b/libs/mrwebrtc/src/interop/global_factory.cpp
@@ -253,7 +253,7 @@ mrsResult GlobalFactory::InitializeImplNoLock() {
     factoryConfig->audioRenderingEnabled = true;
     factoryConfig->enableAudioBufferEvents = false;
     factoryConfig->customAudioMixer =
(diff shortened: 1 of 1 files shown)


libs/mrwebrtc/src/interop/global_factory.cpp
```diff
@@ -251,6 +252,8 @@ mrsResult GlobalFactory::InitializeImplNoLock() {
     factoryConfig->audioCapturingEnabled = true;
     factoryConfig->audioRenderingEnabled = true;
     factoryConfig->enableAudioBufferEvents = false;
+    factoryConfig->customAudioMixer =
+        reinterpret_cast<uint64_t>(custom_audio_mixer_.get());
```
djee-ms, line 256: Probably won't build in 32-bit archs

fibann: It turns out it is correct but produces a warning. Fixed.

PR #538, opened 2020-08-17
Fix issues with Android build

fibann:
UnityUtility has been changed into AndroidCameraInterop, so use that.


fibann:
There are still outstanding issues from #528 in some Windows flavors, but Android seems to be building now.

PR #586, opened 2020-09-22
Native Video Renderer

rygo6-MS:
I am opening this PR to start getting some pre-emptive feedback, and also so you can highlight anything more this would need in order to get merged. Also opening in case anyone wants to help test, or help figure out some of the missing pieces if needed.

This adds the ability to render the WebRTC video streams into an Unity texture through Native C++ calls, bypassing C# marshalling. A new component called `NativeVideoRenderer` can be used in place of the existing `VideoRenderer` component in order to use the new functionality.

There are a handful of new C# scripts in the Unity library, as well as a whole new C++ project which implements the native rendering calls. 

The two big issues I see as remaining:
- Currently this is tested to work only on Windows 10 DX11 and Android OpenGLES. There is a DX12 implementation but is untested. Vulkan and Metal haven't been started on. I am wondering do all graphics API's and platforms need an implementation before this could be merged?
- Currently does not support ARGB. Altspace didn't need it so I haven't put the time on it, but can if it is needed.


libs/mrwebrtc-unityplugin/src/PlatformBase.h
```diff
@@ -0,0 +1,60 @@
...
+// UNITY_METRO - WSA or UWP
+// UNITY_WEBGL - WebGL
+#if _MSC_VER
+	#define UNITY_WIN 1
+#elif defined(__APPLE__)
+	#if defined(__arm__) || defined(__arm64__)
+		#define UNITY_IPHONE 1
+	#else
+		#define UNITY_OSX 1
+	#endif
```
eirikhollis, line 23: With Apple moving over to ARM on their OSX devices, this code will define all new Macs as iPhones.

eirikhollis:
This looks very good! I'll be happy to test it out and see how it performs.
Does it support multiple remote video feeds?

Microsoft.MixedReality.WebRTC.sln
```diff
@@ -41,28 +41,37 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TestReceiveAV", "examples\T
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Implementation", "Implementation", "{5A873D0C-4D1E-4AAA-AE3A-BFC96E796431}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mrwebrtc-unityplugin-win32", "tools\build\mrwebrtc-unityplugin\win32\mrwebrtc-unityplugin-win32.vcxproj", "{3D2AAA9E-C000-4669-AC09-D755D5FAC7D3}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
fibann, line 48: Can you remove the `Any CPU` sln configuration? I assume this was unintended, I think VS slips one in automatically when you modify the sln...

libs/mrwebrtc/src/mrs_errors.h
```diff
@@ -5,8 +5,8 @@
 
 #include <string>
 
-#include "export.h"
-#include "result.h"
+#include "../include/export.h"
+#include "../include/result.h"
```
fibann, line 9: Why is this needed?

libs/mrwebrtc-unityplugin/src/PlatformBase.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
fibann, line 1: This file and the ones in Unity are taken from https://github.com/Unity-Technologies/NativeRenderingPlugin AFAICS. By policy they either need to be in a separate directory (`external/NativeRenderingPlugin` or something like that) or they need to be annotated with the appropriate copyright notice.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -0,0 +1,309 @@
...
+        }
+        break;
+
+      case VideoKind::kARGB:
+        // TODO
+        break;
+
+      case VideoKind::kNone:
+        // TODO
+        break;
```
fibann, line 152: These should return a failure through the API, and/or there should be a comment on the API function saying that the only supported format is I420.

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderingPluginUpdate.cs
```diff
@@ -0,0 +1,101 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.WebRTC.UnityPlugin
+{
+	public class NativeRenderingPluginUpdate : MonoBehaviour
```
fibann, line 11: This should be internal, since it should only be visible to `NativeVideoRenderer`. Also it needs some comments on what it does.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+using System;
+using UnityEngine;
+using Microsoft.MixedReality.WebRTC.UnityPlugin;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.WebRTC
+{
+	/// <summary>
+	/// This will render the video stream through native calls with DX11 or OpenGL, completely bypassing C# marshalling.
+	/// This provides a considerable performance improvement.
```
fibann, line 13:
```suggestion
	/// This provides a considerable performance improvement compared to <see cref="VideoRenderer"/>.
```

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+	/// </summary>
+    [AddComponentMenu("MixedReality-WebRTC/Native Video Renderer")]
+    public class NativeVideoRenderer : MonoBehaviour
+    {
+	    [SerializeField]
+	    private RawImage _rawImage;
```
fibann, line 19:
```suggestion
        [SerializeField]
        private RawImage _rawImage;
```
tabs -> spaces

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+                    break;
+                case VideoEncoding.Argb32:
+	                break;
+            }
+        }
+
+        private void I420AVideoFrameReady(I420AVideoFrame frame)
+        {
+	        // Clear the video display to not confuse the user who could otherwise
+	        // think that the video is still playing but is lagging/frozen.
```
fibann, line 81: Comment is obsolete, delete

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+	        _i420aFrameQueue.Enqueue(frame);
+        }
+
+        private void StartNativeRendering(int width, int height)
+        {
+	        // Subscription is only used to ge the frame dimensions to generate the textures. So Unsubscribe once that is done.
+	        _source.I420AVideoFrameReady -= I420AVideoFrameReady;
```
fibann, line 88: There seems to be a race condition here - the `I420AVideoFrameReady` runs in a different thread, so it might add another frame or more to the queue before being removed from the event. This means that `StartNativeRendering` could be called twice. The class should avoid this, e.g. checking if `_nativeRenderer != null` before calling `StartNativeRendering`.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+                case VideoEncoding.Argb32:
+	                break;
+            }
+        }
+
+        private void I420AVideoFrameReady(I420AVideoFrame frame)
+        {
+	        // Clear the video display to not confuse the user who could otherwise
+	        // think that the video is still playing but is lagging/frozen.
+	        _i420aFrameQueue.Enqueue(frame);
```
fibann, line 82:
```suggestion
	        _i420aFrameQueue?.Enqueue(frame);
```
`_i420aFrameQueue = null` at line 89 might be called while this is still running, so this should avoid crashing on null. Alternatively the queue alive rather than be null-ed it (maybe it can be cleared just in case).

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using UnityEngine;
+using Microsoft.MixedReality.WebRTC.UnityPlugin;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.WebRTC
```
fibann, line 9:
```suggestion
namespace Microsoft.MixedReality.WebRTC.Unity
```
For consistency with the other components in the package.

libs/unity/library/Runtime/Shaders/YUVFeedShaderUnlit.shader
```diff
@@ -1,90 +1,82 @@
...
-        }
-    }
+			fixed4 frag (v2f i) : SV_Target
+			{
+				float3 yuv;
+				yuv.x = tex2D(_YPlane, i.uv).r;
+				yuv.y = tex2D(_UPlane, i.uv).r;
+				yuv.z = tex2D(_VPlane, i.uv).r;
+				return fixed4(yuv2rgb(yuv), 1.0);
+			}
```
fibann, line 78: For curiosity, why this change (half -> float, fixed3 -> fixed4)? @djee-ms any reason why we were using half/fixed3?

tools/build/android/mrwebrtc-unityplugin/build.gradle
```diff
@@ -0,0 +1,69 @@
+apply plugin: 'com.android.library'
```
fibann, line 1: @djee-ms can you double-check the Android parts?

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
```diff
@@ -135,7 +135,7 @@ public static void ReleaseWrapperRef(IntPtr wrapperRef)
         }
 
         [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
-        internal struct SdpFilter
+        public struct SdpFilter
```
fibann, line 138: This shouldn't need to be public.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -740,6 +740,11 @@ public bool Initialized
             }
         }
 
+        /// <summary>
+        /// Accessor for the native peer connection handle.
+        /// </summary>
+        public PeerConnectionHandle NativeHandle => _nativePeerhandle;
```
fibann, line 746: This seems unused now.

libs/Microsoft.MixedReality.WebRTC/Interop/ObjectInterop.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.MixedReality.WebRTC.Interop
     /// <summary>
     /// Handle to a native object.
     /// </summary>
-    internal abstract class ObjectHandle : SafeHandle
```
fibann, line 13: This and `RefCountedObjectHandle` can go back to be internal, AFAICS we are exposing the `IntPtr` directly in `RemoteVideoTrack` so this is unneeded.

libs/Microsoft.MixedReality.WebRTC/RemoteVideoTrack.cs
```diff
@@ -124,6 +124,8 @@ public bool Enabled
         private event I420AVideoFrameDelegate _videoFrameReady;
         private event Argb32VideoFrameDelegate _argb32VideoFrameReady;
 
+        public IntPtr NativeHandle => _nativeHandle.DangerousGetHandle();
```
fibann, line 127: It would be better to expose a `SafeHandle` instead - that will keep the handle to the native object alive. Passing around the `IntPtr` will not.

Also this needs a comment specifying it's internal use only.

libs/mrwebrtc-unityplugin/include/api.h
```diff
@@ -0,0 +1,67 @@
...
+
+//// Register textures for remote video and start rendering it.
+MRS_API mrsResult MRS_CALL
+mrsNativeRenderer_EnableRemoteVideo(mrsRemoteVideoTrackHandle videoHandle,
+                                    VideoKind format,
+                                    TextureDesc textures[],
+                                    int textureCount) noexcept;
+
+/// Clear remote textures and stop rendering remote video.
+MRS_API mrsResult MRS_CALL mrsNativeRenderer_DisableRemoteVideo(mrsRemoteVideoTrackHandle videoHandle) noexcept;
```
fibann, line 52: It would be more consistent with the rest of the API if `mrsNativeRenderer_Create` returned a `mrsNativeRendererHandle` and the other methods took a `mrsNativeRendererHandle` rather than a `mrsRemoteVideoTrackHandle`. `mrsNativeRendererHandle` could just be a typedef of `NativeRenderer*` and you wouldn't need a global map of remote track to renderer. See the other mrs*Handle types, this is the same pattern used for them.

libs/mrwebrtc-unityplugin/include/export.h
```diff
@@ -0,0 +1,16 @@
...
+// P/Invoke uses stdcall by default. This can be changed, but Unity's IL2CPP
+// does not understand the CallingConvention attribute and instead
+// unconditionally forces stdcall. So use stdcall in the API to be compatible.
+#if defined(MR_SHARING_WIN)
+	#define MRS_API __declspec(dllexport)
+	#define MRS_CALL __stdcall
+#elif defined(MR_SHARING_ANDROID)
+	#define MRS_API __attribute__((visibility("default")))
+	#define MRS_CALL __attribute__((stdcall))
+#endif
```
fibann, line 16: This looks wrong, `MR_SHARING_WIN` (and `_ANDROID`) are used by mrwebrtc to turn on the function export, so if you use the same macro in mrwebrtc-unityplugin the compiler will compile in and export the same functions twice.

At the moment `export.h` in mrwebrtc lacks a configuration where the API is compiled as `dllimport`, so if you want to correctly use the mrwebrtc functions in mrwebrtc-unityplugin you should rewrite that file (and this file, with different macros) so that it handles that case, like the example in https://stackoverflow.com/a/2164853/6705316.

A quicker-and-dirtier alternative is not to include _any_ header from mrwebrtc in mrwebrtc-unityplugin and redefine the types you need (`mrsRemoteVideoTrackHandle`, `mrsResult`). Even in that case you should rename the macros in this file (`MR_SHARING_WIN`, `MRS_API`, `MRS_CALL`) to something different to avoid confusion.

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderer.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        public IntPtr texture;
+        public int width;
+        public int height;
+    }
+
+    /// <summary>
+    /// High-performance, low-level, native video rendering for Microsoft.MixedReality.WebRTC.PeerConnection objects.
+    /// </summary>
+    public class NativeRenderer : IDisposable
```
fibann, line 28: Same as `NativeRendererPluginUpdate`, this should be an internal class or nested in `NativeVideoRenderer` as it only needs to be visible there. Also, the name is a bit confusing since it's not clear what the difference is between this and `NativeVideoRenderer`. It should probably be `NativeRendererProxy` or something like that.

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderer.cs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.MixedReality.WebRTC.UnityPlugin
```
fibann, line 7: This is too similar with `Microsoft.MixedReality.WebRTC.Unity` and the difference is not clear. It should either be merged with `Microsoft.MixedReality.WebRTC.Unity` or it could be a sub-namespace, e.g. `Microsoft.MixedReality.WebRTC.Unity.Renderer`.

fibann (changes requested):
Awesome contribution @rygo6-MS! There are a handful of things to change, mainly to make the API more consistent, but it looks good!

Re support for Vulkan, etc. and ARGB: I'd say it can go in without (provided that the missing support is documented), it seems very easy to add that afterwards if needed.

Two more things that would be good to add:
- documentation on how to build and use the plugin
- support for attaching a native renderer to local video tracks/sources as well as remote tracks; this would take some tampering with types in the C API as at the moment there is no common base;

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -0,0 +1,219 @@
...
+        private void I420AVideoFrameReady(I420AVideoFrame frame)
+        {
+	        // Clear the video display to not confuse the user who could otherwise
+	        // think that the video is still playing but is lagging/frozen.
+	        _i420aFrameQueue.Enqueue(frame);
+        }
+
+        private void StartNativeRendering(int width, int height)
+        {
+	        // Subscription is only used to ge the frame dimensions to generate the textures. So Unsubscribe once that is done.
```
fibann, line 87: Note that _in theory_ WebRTC can rescale the stream and change the frame dimensions from one frame to another. In practice I haven't seen it happening in my tests. So it's fine if this supports constant dimensions for now, but we should take care to fail cleanly (or at the very least crash with a clear error in a clear callstack) if the dimensions happen to change.

djee-ms: I've seen it happening on 2 occasions:
- Network conditions degradation/recovering (hard to repro)
- On Android, simply rotate the phone from portrait to landscape, and that flips width/height.

eirikhollis: Frame dimensions can also vary when changing source without reconnecting the Peer, e.g. when going from webcam to a shared desktop screen. We change resolutions on average at least once in every call, so constant dimensions wouldn't work for us.

libs/mrwebrtc-unityplugin/src/PlatformBase.h
rygo6-MS: What define could be alternatively used?

Microsoft.MixedReality.WebRTC.sln
rygo6-MS: Deleted those lines.

libs/mrwebrtc/src/mrs_errors.h
rygo6-MS: Think that was an unnecessary remnant from when I was trying to get android build working. Deleted the ../include

libs/mrwebrtc-unityplugin/src/PlatformBase.h
rygo6-MS: What copyright notice is needed? 

Would this be enough?
```
// Copyright (c) Unity Technologies.
// Licensed under the MIT License.
https://github.com/Unity-Technologies/NativeRenderingPlugin
```

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
rygo6-MS: Added warning logs and note to api summary.

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderingPluginUpdate.cs
rygo6-MS: A monobehaviour can't be nested inside of another class, but I changed it to internal.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: Changed to your wording.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: deleted

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: Added null check for _nativeRenderer

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: added

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: Updated

libs/unity/library/Runtime/Shaders/YUVFeedShaderUnlit.shader
rygo6-MS: I changed it to half. I may have typed float out of habit. Its rare this makes a difference on modern hardware but probably better to use lowest precision anyways.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
rygo6-MS: deleted

libs/Microsoft.MixedReality.WebRTC/Interop/ObjectInterop.cs
rygo6-MS: changed to internal

libs/Microsoft.MixedReality.WebRTC/RemoteVideoTrack.cs
rygo6-MS: So I should return the `RemoteVideoTrackInterop.RemoteVideoTrackHandle` ? That would require `RemoteVideoTrackHandle` and `RemoteVideoTrackInterop` to become public. Which could happen, but the purpose of this property is so that the Unity plugin can retrieve the native pointer and send it down to C++.  The Unity plugin which retrieves this does not itself hold onto the managed pointer, it immediately passes it to C++, so it would not benefit from the SafeHandle keeping the object alive. Which with that, do you think it should still be changed to return `RemoteVideoTrackHandle`?

fibann: That's a good point. Since this is a public property I'd say it's still better to be safe than sorry and return a `SafeHandle`, just to prevent bugs if some code at some point decides to keep this around, but you can also just add a "Internal use" comment and it should be fine for now. Definitely not worth exposing `RemoteVideoTrackHandle`.

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
rygo6-MS: If I try to set this to internal I get this error on compiling:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0051	Inconsistent accessibility: parameter type 'Utils.SdpFilter' is less accessible than method 'Utils.SdpForceCodecs(string, Utils.SdpFilter, Utils.SdpFilter, StringBuilder, ref ulong)'	Microsoft.MixedReality.WebRTC	C:\mrwebrtc\libs\Microsoft.MixedReality.WebRTC\Interop\InteropUtils.cs	170	Active
```

But I notice it is like this in master? Not sure if I am missing something, or need to set some .NET setting differently, but it appears to me like the way it is in master wouldn't compile.

libs/mrwebrtc-unityplugin/include/export.h
rygo6-MS: This file I meant to delete. I switched it to using the export.h imported from the actual mrwebrtc project. Does doing it that way still cause any issue?

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderer.cs
rygo6-MS: Renamed NativeRender to NativeVideo and NativeRendererInterop to NativeVideoInterop. I think this naming will work better. You create a NativeVideo object, and the NativeVideoRenderer renders the NativeVideo.

libs/unity/library/Runtime/Scripts/NativeRender/NativeRenderer.cs
rygo6-MS: Changed all UnityPlugin namespaces to just Unity.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
fibann: Can you add a comment to `NativeVideoRenderer` and `mrsNativeRenderer_EnableRemoteVideo` saying that only I420A is supported at the moment?

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
fibann: I see what's happening, `SdpForceCodecs` is declared as public but `Utils` is an internal class on master, so the method is actually internal. Here you have made `Utils` a public class so `SdpForceCodecs` is considered public, and the compiler wants also `SdpFilter` to be public.

AFAICS you only need `Utils` to be public because `ThrowOnErrorCode` is used in `NativeVideo`, right? I think the correct way to fix this is to leave `Utils` internal and have the native renderer library define its own error codes. Anyway at the moment it seems that all the methods are always returning `kSuccess` so I'd say just comment the `ThrowOnErrorCode` calls and revert this file here.

libs/mrwebrtc-unityplugin/include/export.h
fibann: Using the same file or a copy-paste doesn't make any difference if you are still using the same macros for both DLLs. unityplugin is compiled as a different DLL from mrwebrtc, so it should use different macros. If they all use `MRS_API`, when you compile unityplugin:
- the functions that are defined in unityplugin are compiled as MRS_API = dllexport (correct)
- any of the API functions in mrwebrtc included by unityplugin will also be compiled as MRS_API = dllexport (wrong! unityplugin doesn't contain its definition, only the declaration, so the linker will fail; should be dllimport instead, because it is importing the function from mrwebrtc)

I am not sure how it is working now, I suppose that since unityplugin has an exports.def, all `__declspec(export)` are ignored (and `__declspec(dllimport)` is optional, so the mrwebrtc functions get imported correctly anyway). But this is bug-prone and confusing, so it should be done the correct way. Since you don't really need `dllimport` after all, I guess a simple fix is this:
- re-add export.h
- rename all the macros to something unique to unityplugin (e.g. `MR_SHARING_`, `MRS_` -> `MRWEBRTC_UNITYPLUGIN_`)
- use the renamed macros in `mrwebrtc-unityplugin/include/api.h` and in the preprocessor definitions in the `vcxproj`

rygo6-MS: Thank you for explaining all of that. Setting up dll exports like this on C++ libraries to go into unity is one of the areas I am not the most knowledgeable about.

I deleted the exports.def file so that it goes off al of __declspec(export) calls. But it still looks like it is correct in what it exports. Running dumpbin I get:
```
1    0 000010F0 OnGraphicsDeviceEvent = OnGraphicsDeviceEvent
2    1 00001260 UnityPluginLoad = UnityPluginLoad
3    2 000012B0 UnityPluginUnload = UnityPluginUnload
4    3 00001330 mrsNativeRenderer_Create = mrsNativeRenderer_Create
5    4 00001450 mrsNativeRenderer_Destroy = mrsNativeRenderer_Destroy
6    5 000015B0 mrsNativeRenderer_DisableRemoteVideo = mrsNativeRenderer_DisableRemoteVideo
7    6 00001470 mrsNativeRenderer_EnableRemoteVideo = mrsNativeRenderer_EnableRemoteVideo
8    7 000015D0 mrsNativeRenderer_GetVideoUpdateMethod = mrsNativeRenderer_GetVideoUpdateMethod
9    8 000012E0 mrsNativeRenderer_SetLoggingFunctions = mrsNativeRenderer_SetLoggingFunctions   
```

Does this mean that the unityplugin.dll is not compiling in any methods from mrwebrtc.dll? I suspect this is so because as it currently is, you must put both the unityplugin.dll and mrwebrtc.dll into Unity. The unityplugin.dll dynamically links to the mrwebrtc.dll

But, I will rename unity macros as you suggest, I assume that is more correct even if it is dynamically linking.

rygo6-MS: However I have found that I still need `MR_SHARING_WIN` defined in the preprocessor definitions for the mrwebrtc-unityplugin project. Is that an issue?

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: I made it output a warning if frame dimensions change after initialization. Some research will need to be done to figure out how to let this texture be resized from the C++ side.

libs/mrwebrtc-unityplugin/include/export.h
fibann: I am a bit surprised that it works - it might be that the linker figures out that the functions are imported from mrwebrtc on its own.

You should rename `MR_SHARING_WIN` and `MR_SHARING_ANDROID` too, they are supposed to be specific to the DLL too.

> Setting up dll exports like this on C++ libraries to go into unity is one of the areas I am not the most knowledgeable about.

I have always found this to be surprisingly badly documented, the MS docs explain what things do but fail IMHO to give a comprehensive usage picture. You can find a few examples on SO, e.g. https://stackoverflow.com/questions/538134/exporting-functions-from-a-dll-with-dllexport, maybe that will give you a few more info.

libs/Microsoft.MixedReality.WebRTC/Interop/InteropUtils.cs
fibann: `NativeVideo` still has a few calls to `ThrowOnErrorCode` that fail to compile now.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
fibann: This is good enough for the PR to go in, I will open an issue to improve this once that's done.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -0,0 +1,271 @@
...
+    std::lock_guard guard(m_lock);
+    m_remoteVideoFormat = format;
+    switch (format) {
+      case VideoKind::kI420:
+        if (textureDescCount == 3) {
+          m_remoteTextures.resize(3);
+          m_remoteTextures[0] = textureDescs[0];
+          m_remoteTextures[1] = textureDescs[1];
+          m_remoteTextures[2] = textureDescs[2];
+          mrsRemoteVideoTrackRegisterI420AFrameCallback(
```
fibann, line 104: Just realized this - only one callback can be registered with this function, so enabling the native renderer for the track will unregister previous callbacks and prevent `FrameReady` events from being called. So the native renderer will take exclusive control of the video track and break existing subscriptions. This should be documented in the comments on `mrsNativeRenderer_EnableRemoteVideo` and `NativeVideoRenderer`.

I'll also log an issue to improve this after merge.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+void NativeRenderer::DisableRemoteVideo() {
+  Log_Debug("NativeRenderer::DisableRemoteVideo");
+  {
+    // Instance lock
+    std::lock_guard guard(m_lock);
+    m_remoteTextures.clear();
+    m_remoteVideoFormat = VideoKind::kNone;
+  }
```
fibann, line 131: This should call `mrsRemoteVideoTrackRegisterI420AFrameCallback(m_handle, nullptr, nullptr)`.

rygo6-MS:
I believe all issues have been remedied. Except the documentation. Should that be put somewhere in this PR? Or should I make a PR to gh-pages branch and edit this page:
https://microsoft.github.io/MixedReality-WebRTC/manual/installation.html

libs/mrwebrtc-unityplugin/src/PlatformBase.h
fibann:
```suggestion
	#if TARGET_OS_TV
		#define UNITY_TVOS 1
	#elif TARGET_OS_IOS
		#define UNITY_IOS 1
	#else
		#define UNITY_OSX 1
	#endif
```
From https://github.com/Unity-Technologies/NativeRenderingPlugin/blob/master/PluginSource/source/PlatformBase.h - I haven't tested it but it looks sensible. Also the same file defines `UNITY_ANDROID` explicitly - why isn't this defined here?

tools/build/android/mrwebrtc-unityplugin/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+        versionName rootProject.ext.versionName
+        ndk {
+            abiFilters "arm64-v8a"
+        }
+        packagingOptions {
+            doNotStrip "*/arm64-v8a/*.so"
+        }
+        externalNativeBuild {
+            cmake {
+                cppFlags "-std=c++17 -DANDROID_STL=c++_shared -DMRS_USE_STR_WRAPPER -DMR_SHARING_ANDROID -DMR_UNITYPLUGIN_SHARING_ANDROID -DUNITY_ANDROID -DSUPPORT_OPENGL_ES=1"
```
fibann, line 20:
```suggestion
                cppFlags "-std=c++17 -DANDROID_STL=c++_shared  -DMR_UNITYPLUGIN_SHARING_ANDROID -DUNITY_ANDROID -DSUPPORT_OPENGL_ES=1"
```
`MRS_USE_STR_WRAPPER` is obsolete, can be removed. `MR_SHARING_ANDROID` should only be used when building mrwebrtc as from previous comment.

tools/build/android/webrtc-native/build.gradle
```diff
@@ -17,7 +17,7 @@ android {
         }
         externalNativeBuild {
             cmake {
-                cppFlags "-std=c++14 -DANDROID_STL=c++_shared -DMRS_USE_STR_WRAPPER -DMR_SHARING_ANDROID -DWEBRTC_POSIX -DWEBRTC_ANDROID"
+                cppFlags "-std=c++14 -DANDROID_STL=c++_shared -DMRS_USE_STR_WRAPPER -DMR_SHARING_ANDROID -DUNITY_ANDROID -DWEBRTC_POSIX -DWEBRTC_ANDROID -DSUPPORT_OPENGL_ES=1"
```
fibann, line 20: Why do you need to add these defines to this lib?

libs/mrwebrtc-unityplugin/src/pch.h
```diff
@@ -0,0 +1,52 @@
...
+#endif
+
+#include <windows.h>
+
+#elif defined(MR_SHARING_ANDROID)
+
+#define WEBRTC_POSIX 1
+#define WEBRTC_ANDROID 1
+
+#endif
```
fibann, line 39:
```suggestion
#if defined(MR_UNITYPLUGIN_SHARING_WIN)

#include "targetver.h"

#define WEBRTC_WIN 1

#ifndef WIN32_LEAN_AND_MEAN
#define WIN32_LEAN_AND_MEAN
#endif

#ifndef NOMINMAX
#define NOMINMAX
#endif

#include <windows.h>

#elif defined(MR_UNITYPLUGIN_SHARING_ANDROID)

#define WEBRTC_POSIX 1
#define WEBRTC_ANDROID 1

#endif
```
I am not sure you need the `#define WEBRTC_*` lines since this project is not including WebRTC code directly AFAICS (but they shouldn't hurt so fine to leave them).

fibann:
@rygo6-MS I have quickly checked the Android build pieces, see a few minor things to fix.

The documentation is the docs folder, the gh-pages content is autogenerated from the MD files in docs. The info about the native renderer should probably go into [docs/manual/unity/unity-mediaplayer.md](https://github.com/microsoft/MixedReality-WebRTC/blob/master/docs/manual/unity/unity-mediaplayer.md) . Also it's probably worth adding a link from [docs/manual/unity/unity-integration.md](https://github.com/microsoft/MixedReality-WebRTC/blob/master/docs/manual/unity/unity-integration.md). It can be done in a separate PR anyway, this can be merged as soon as the last comments are addressed.

JacopoMarullo:
Hi! Is it supposed to run also on HL2? can you tell something about the performance gain that it can provide? For example for a 1080p video stream how much milliseconds we can gain for each frame rendering? 
Just a curiosity. 

Thanks a lot!

tools/build/android/mrwebrtc-unityplugin/build.gradle
rygo6-MS: I removed MRS_USE_STR_WRAPPER but cannot remove MR_SHARING_ANDROID because the unityplugin references files in mrwebrtc that need that define.

tools/build/android/webrtc-native/build.gradle
rygo6-MS: These are needed for the unity graphics interfaces, it doesn't compile if they are removed.

rygo6-MS:
> Hi! Is it supposed to run also on HL2? can you tell something about the performance gain that it can provide? For example for a 1080p video stream how much milliseconds we can gain for each frame rendering?
> Just a curiosity.
> 
> Thanks a lot!

I have not tested if this works on HL2 yet, but I can't think of a reason why it wouldn't, and it should. I will be testing it this week.

tools/build/android/mrwebrtc-unityplugin/build.gradle
fibann: Right, it seems that some code is using MR_SHARING_ANDROID as a generic "is this Android" check rather than as "are we building mrwebrtc for Android". Will log an issue, but this can go ahead as long as it works.

tools/build/android/webrtc-native/build.gradle
fibann: But this file is used for mrwebrtc which doesn't include the Unity graphics interfaces, right?

fibann (approved):
Good to go! Will log issues for the remaining things

tools/build/android/mrwebrtc-unityplugin/build.gradle
fibann: Logged as #629.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
fibann: Logged as #630

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
fibann: Logged as #631

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
fibann: Logged support for ARGB as #632

PR #634, opened 2020-12-02
Include native dependencies in project when using NuGet with old style csprojs (package.config) or paket

sanderaernouts:
The `Microsoft.MixedReality.WebRTC` NuGet package uses the runtimes folder for the native dependencies (`mrwebrtc.dll`) however this does not work when the project consuming this package is using the "old style csproj" with `package.config` or when a project uses Paket and targets full .NET framework.

This `.props` file adds the correct bitness of `mrwebrtc.dll` as a hidden content item and ensures it gets copied to the output directory when a `package.config` or `paket.reference` file is found in the root of the project.

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
+++ b/libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
@@ -39,6 +39,13 @@
     <NativeUnityPluginSourceFiles Include="$(OutputPath)\Microsoft.MixedReality.WebRTC.dll;$(OutputPath)\Microsoft.MixedReality.WebRTC.pdb" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Include="package.props">
+      <Pack>true</Pack>
+      <PackagePath>build\$(PackageId).props</PackagePath>
+    </None>
+</ItemGroup>
+
   <Target Name="PostBuild" AfterTargets="PostBuildEvent">
     <Copy SourceFiles="@(NativeUnityPluginSourceFiles)" DestinationFolder="$(ProjectDir)..\unity\library\Runtime\Plugins\Win32\x86_64" />
   </Target>
--- a/libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
+++ b/libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
@@ -39,13 +39,6 @@
     <NativeUnityPluginSourceFiles Include="$(OutputPath)\Microsoft.MixedReality.WebRTC.dll;$(OutputPath)\Microsoft.MixedReality.WebRTC.pdb" />
   </ItemGroup>
 
-  <ItemGroup>
-    <None Include="package.props">
-      <Pack>true</Pack>
-      <PackagePath>build\$(PackageId).props</PackagePath>
-    </None>
-</ItemGroup>
-
   <Target Name="PostBuild" AfterTargets="PostBuildEvent">
     <Copy SourceFiles="@(NativeUnityPluginSourceFiles)" DestinationFolder="$(ProjectDir)..\unity\library\Runtime\Plugins\Win32\x86_64" />
   </Target>
--- a/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
+++ b/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
@@ -1,5 +1,5 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
+  <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
     <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Visible>false</Visible>
--- a/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
+++ b/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
@@ -1,5 +1,5 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup Condition="Exists('packages.config') OR Exists('$(MSBuildProjectName).packages.config') OR Exists('packages.$(MSBuildProjectName).config') OR Exists('paket.references')">
+  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
     <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Visible>false</Visible>
--- a/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
+++ b/tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
@@ -1,6 +1,6 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
-    <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
+    <Content Include="$(MSBuildThisFileDirectory)\..\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Visible>false</Visible>
       <Link>mrwebrtc.dll</Link>
--- a/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
+++ b/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
@@ -1,5 +1,5 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
+  <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
     <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Visible>false</Visible>
--- a/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
+++ b/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
@@ -1,0 +1,9 @@
+﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
+    <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Visible>false</Visible>
+      <Link>mrwebrtc.dll</Link>
+    </Content>
+  </ItemGroup>
+</Project>
--- a/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
+++ b/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
@@ -1,6 +1,6 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
-    <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
+    <Content Include="$(MSBuildThisFileDirectory)\..\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <Visible>false</Visible>
       <Link>mrwebrtc.dll</Link>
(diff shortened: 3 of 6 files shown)


sanderaernouts:
Wondering if it would be better to use `Condition="'$(UsingMicrosoftNETSdk)' == 'true'"` so that he .props file is only included for any non SDK style csproj instead of checking for the `packages.config` or `paket.references` files.

libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
```diff
@@ -39,6 +39,13 @@
     <NativeUnityPluginSourceFiles Include="$(OutputPath)\Microsoft.MixedReality.WebRTC.dll;$(OutputPath)\Microsoft.MixedReality.WebRTC.pdb" />
   </ItemGroup>
 
+  <ItemGroup>
```
fibann, line 42: Will this work even if the NuGet package is generated using a nuspec (which is the case for the official packages)?

fibann:
> 
> 
> Wondering if it would be better to use `Condition="'$(UsingMicrosoftNETSdk)' == 'true'"` so that he .props file is only included for any non SDK style csproj instead of checking for the `packages.config` or `paket.references` files.

Sounds much better to me

libs/Microsoft.MixedReality.WebRTC/Microsoft.MixedReality.WebRTC.csproj
sanderaernouts: @fibann I must have assumed you were using the SDK style CSPROJ way of building packages. I removed the `.props` file from the `.cproj` and moved it to the `nuget` folder. I'm not sure about the relative paths I placed in the `.nuspec` though. Tried to figure it out based on your YAML pipelines but couldn't really see if the .props file should be moved anywhere.

sanderaernouts:
> > Wondering if it would be better to use `Condition="'$(UsingMicrosoftNETSdk)' == 'true'"` so that he .props file is only included for any non SDK style csproj instead of checking for the `packages.config` or `paket.references` files.
> 
> Sounds much better to me

Done 👍

djee-ms:
Hi @sanderaernouts,

Many thanks for this contribution, although we technically don't support .NET Framework (nor packages.config) I know many users have tried to make the project work with it, and this looks like a nice and simple way to help those users, so I'm all for merging it.

It looks like the same change should probably be applied to the UWP package too https://github.com/microsoft/MixedReality-WebRTC/blob/master/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.nuspec, would you mind adding that to your PR please? I am on my side in a discussion with the NuGet team internally to confirm this is the right approach to managing native DLLs with `packages.config`-style projects, which is unfortunately not very well documented. Or do you have a reference maybe about this `.props`-based approach?

Thanks!

sanderaernouts:
@djee-ms sure thing, I have just pushed the same changes for [Microsoft.MixedReality.WebRTC.UWP.nuspec]( https://github.com/microsoft/MixedReality-WebRTC/blob/master/tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.nuspec).

>I am on my side in a discussion with the NuGet team internally to confirm this is the right approach to managing native DLLs with packages.config-style projects, which is unfortunately not very well documented. Or do you have a reference maybe about this .props-based approach?

I don't have a good reference for this approach, I have used a similar approach in a different scenario to reliably get DLL's into the bin folder. Another approach is to use MSBuild targets to copy the files into the `OutputDirectory` but I found this approach to be much simpler. It would be good to get a definitive answer from the NuGet team on this that would definitely also help me out in other (similar) cases.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
```diff
@@ -0,0 +1,9 @@
+﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
```
zivkan, line 2: non-SDK style .NET projects can use PackageReference, but it's a small percentage.

Having said that, I was looking thought [NuGet's targets file](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets), and I see that `RestoreProjectStyle` is set in the target `_GetRestoreProjectStyle`. However, I couldn't figure out how package authors (or even customers in their own props/targets, like Directory.Build.targets) can condition on this property value.

@nkolev92 @jeffkl any ideas? Should they use `CollectPackageReferences`?  I always assumed that was for PackageReference only, whereas this props file is trying to add a `Content` element only for non-PackageReference package usage.

nkolev92: I don't think you can do it. 

Also this targets file likely needs to be a target framework specific location in order to avoid triggering false compatibility (build/packageid.props makes it compatible with everything.)

djee-ms: @nkolev92 how do we do that? I'm really at my limit of understanding of NuGet here, barely following the conversation above. And is there an alternative then to drive a condition block on whether `PackageReference` or `packages.config` is used, that would be better than `$(UsingMicrosoftNETSdk)`?

zivkan: In the nuspec file, where it puts `Microsoft.MixedReality.WebRTC.props` in the package, instead of putting it in `build\`, put it in `build\netstandard2.0\`, since your package appears to contain only `netstandard2.0` assets.

If your package supported multiple TFMs, for example you had `lib\netstandard2.0\` and `lib\net5.0\`, then you'd need the msbuild files in both `build\netstandard2.0\` and `build\net5.0\`.

zivkan: Regarding "correctly" supporting PackageReference vs packages.config (or Paket), unfortunately we don't seem to have a solution. I don't have other ideas on how we can easily support all PackageReference and packages.config (include Paket) perfectly.

I'm guessing by the repo name that there might be significant usage from UWP apps?  My memory is that UWP projects are non-SDK style and might use PackageReference. So this PR will cause both this `<Content` to copy the file, in addition to NuGet's `PackageReference` feature, so there will be double writes of this file. I don't think there are any more serious consequences than double writing the file.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
```diff
@@ -0,0 +1,9 @@
...
+    <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
```
amaitland, line 3: From memory`$(PlatformTarget)` is defined after the `props` files are included, it's usually defined in a `PropertyGroup` based on `$(Platform)`. Perhaps this works on newer versions of `VS`, I'd be surprised if it worked on older versions.

I expect you'll run into the same issue as described in https://github.com/dotnet/project-system/issues/4158#issuecomment-645671777 where the file copy doesn't happen on a clean checkout as the `.props` file doesn't exist on disk and isn't evaluated as part of the project when using older `NonSdkStyle` project (packages.config).

> When the restore occurs the files will be written to disk. If we were to re-evaluate the project at that point we would > pick them up and the <Reference> items they contain. However, since we don't even know the <Import> exists we > don't know we need to re-evaluate. If we re-evaluate the project for some other reason (e.g., it is unloaded and > reloaded, or a .cs file or reference is added) we will then see the imported files.

I'd expect this to only work after closing/reopening the project/solution.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
amaitland: For `PackageReference` projects the `.csproj.nuget.g.props` file defines a `$(NuGetProjectStyle)` property. 

```
<NuGetProjectStyle Condition=" '$(NuGetProjectStyle)' == '' ">PackageReference</NuGetProjectStyle>
```

Under `VS2017/VS2019` this would appear to be set for `SDK Style` projects.
I also tested under `VS/2017/2019` for `non-SDK style` project using `PackageReference` and it's also set as expected. 

I'm unfamiliar with `Paket`. Does anyone know if this would be a better check than `$(UsingMicrosoftNETSdk)' != 'true'`? 

```
<ItemGroup  Condition=" '$(NuGetProjectStyle)' == ''">
```

Is anyone aware of any circumstances where this check would be incorrect?

I'm also interested in adding better support for `non-SDK style` projects to my own project [CefSharp](https://github.com/cefsharp/CefSharp) and have been following https://github.com/NuGet/docs.microsoft.com-nuget/issues/2070#issuecomment-767706280. 

Thanks for your time.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
sanderaernouts: @amaitland so if I understand you correctly there is no proper way to do this in the old project system apart from manually adding a props file similar to this to the project that includes this package?

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
zivkan: Good idea/find @amaitland. I'd suggest that `'$(NuGetProjectStyle)' != 'PackageReference'` would be a better condition, although I didn't test it.

` '$(NuGetProjectStyle)' == ''` will always evaluate to true in `Directory.Build.props` (but so will `!= 'PackageReference'`) because MSBuild evaluates top down, and this file is always imported before the project's `<project>.nuget.g.props`. In a project file, it will evaluate to `true` when the `obj/` folder is empty (before NuGet restore after `git clean -xdf`, or clean checkout on a CI agent). This is why it's important to use `msbuild -restore`, not `msbuild -t:restore,build` if trying to restore & build in a single command line when using msbuild.exe. The `dotnet` CLI does this automatically.

But, it should actually work really great in a package's props/targets file. If the project uses `PackageReference`, the package's props file will be imported by the file that defines this property, so it's guaranteed to be set, and set to `PackageReference`. For packages.config, the property won't be set, hence MSBuild treats it equal to the empty string. I also don't know how Paket works, but if they define `NuGetProjectStyle` to `PackageReference` then frankly it's their own fault for misrepresenting their tool. If they set it to `Paket`, then the `!= 'PackageReference'` condition will ensure it doesn't get skipped accidentally. Similarly, if NuGet starts setting `NuGetProjectStyle` to `PackagesConfig` for `packages.config` projects (if you check NuGet's `NuGet.targets` linked above, we already do this, but in a target that should not currently trigger in packages.config projects), the condition will also keep working as planned.

So yes, I recommend this based on @amaitland's suggestion:
```suggestion
  <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
```

tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
```diff
@@ -0,0 +1,9 @@
+﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Condition="'$(UsingMicrosoftNETSdk)' != 'true'">
```
zivkan, line 2: As per https://github.com/microsoft/MixedReality-WebRTC/pull/634#discussion_r567004196

```suggestion
  <ItemGroup Condition="'$(NuGetProjectStyle)' != 'PackageReference'">
```

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
amaitland: @zivkan Thanks for the additional insight 👍

tools/nuget/UWP/Microsoft.MixedReality.WebRTC.UWP.props
sanderaernouts: I applied the suggested change to both `.props` files 👍

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
sanderaernouts: > In the nuspec file, where it puts `Microsoft.MixedReality.WebRTC.props` in the package, instead of putting it in `build\`, put it in `build\netstandard2.0\`, since your package appears to contain only `netstandard2.0` assets.
> 
> If your package supported multiple TFMs, for example you had `lib\netstandard2.0\` and `lib\net5.0\`, then you'd need the msbuild files in both `build\netstandard2.0\` and `build\net5.0\`.

Good catch @zivkan! @djee-ms  I moved the props file into the framework-specific folders. Originally I had the props files as part of the CSPROJ and I did not check the target framework when adding them to the `.nuspec` files. My bad!

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
sanderaernouts: Though I guess is a "known issue" with the old project system, any NuGet package can add `.targets` or `.props` files that would suffer from the same problem.

zivkan (approved):
I think it would be a good idea to do some manual testing of installing the packages created by this PR into various different project types, but conceptually I think it looks good from a NuGet point of view.

tools/nuget/Desktop/Microsoft.MixedReality.WebRTC.props
amaitland: If you switch from a .props file to a .targets file the PlatformTarget will be accessible. 

I'm not aware of a perfect solution unfortunately. VS2017 will fail to load the props after a Nuget install/restore.  VS2013/2015 don't have the problem. VS2019 appears to be working, I had a few instances where the files weren't copied, might have been my fault as I was testing some changes, sample size of one so might be nothing. 

I've added an after build task with a condition to check if the file exists in the output folder, display a message in the build output, I was considering outputting an error so the build fails. You could do something similar. 

You could add a after build target to copy the file to the output directory, you'd probably need an after publish target also.

This is just from my experience, I'm not a definitive authority on the subject, just someone who's spent a lot of time hacking Nuget packages to try get things working. If someone has some other suggestions I'd be very interested in hearing them.

djee-ms: I just tried creating a new .NET Framework console app (4.7.2) and adding a manually-generated package with the config in this PR, after fixing the path (since the `.props` file is in `build\netstandard2.0`, so 2 folders down from the root, not 1):

```diff
- <Content Include="$(MSBuildThisFileDirectory)\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
+ <Content Include="$(MSBuildThisFileDirectory)\..\..\runtimes\win10-$(PlatformTarget)\native\mrwebrtc.dll">
```

(ping @sanderaernouts can you fix the PR please?)

...and the only issue I got is that by default `$(PlatformTarget)` is `AnyCPU` because VS only creates that target. That error _will_ fail the build and be rather self-explanatory I think. After manually adding an `x64` target (for example) and switching to that, the project builds and runs without issue. I am using Visual Studio 2019 16.8.

@amaitland noted, thanks for the detailed explanation. But at this point we don't really support VS 2017 officially, that would make a bit too much work to test also all VS versions, so we can only guarantee VS 2019. So I think we can ignore that issue for now to unblock this PR. That being said after that if you want to propose a change to add an after build task for VS 2017 I am happy to have a look at it.

sanderaernouts: @djee-ms I have updated the paths in the props files to go 2 levels up instead of 1.

> ...and the only issue I got is that by default $(PlatformTarget) is AnyCPU because VS only creates that target. That error will fail the build and be rather self-explanatory I think. After manually adding an x64 target (for example) and switching to that, the project builds and runs without issue. I am using Visual Studio 2019 16.8.

regarding this ☝️ , we could define a default when AnyCPU is used, either `x86` or `x64` but I don't know if that is correct? Another pattern I have seen with `AnyCPU` and native code is to include both versions in `x64` and `x86` subfolders but then the code that is invoking `mrwebrtc.dll` needs to decide on using `x64\mrwebrtc.dll` or `x86\mrwebrtc.dll` based on the actual platform it is running on.

djee-ms (approved):

djee-ms:
Thanks for your contribution @sanderaernouts and for sticking around through the (long) review! :)

PR #641, opened 2020-12-23
Fix bug with comparing ids

desto12:
After this change you can change in c# which camera to use.


fibann (approved):

fibann:
Good catch @desto12, thanks!

Issue #673, opened 2021-02-17
Support string in DataChannel.SendMessage 


torepaulsson:
I've tested to implement this somewhat in this draft PR: https://github.com/microsoft/MixedReality-WebRTC/pull/675

**Describe the problem**
I'm not 100% sure about this, but, once a DataChannel is setup there is these two methods to send raw bytes over the data channel:
```csharp
SendMessage(IntPtr message, ulong size)
SendMessage(byte[] message)
```

The only way to send messages right now is using raw bytes, but there should be another alternative according to the RFC (https://tools.ietf.org/html/rfc8831). The RFC states that the Payload Protocol Identifier (PPID) used in the data channel can be used to indicate raw bytes or UTF8 encoded data.

It would be great to have this available in the package as well, and might not be much trouble if the underlying support in the library is already there (hopefully).

**Describe the solution you'd like**
Overload the SendMessage method with another alternative:
```csharp
DataChannel.SendMessage(string message)
```
This would UTF8 encode the message and send it by with the PPID flag.

**Describe alternatives you've considered**
I see no other way to make this happen, there is no way to send the raw bytes as a UTF8 message since it is set by the underlying protocol, I think?

torepaulsson:
I've checked the C++ code and found the following in data_channel.cpp
![image](https://user-images.githubusercontent.com/10498076/108224422-315b2c80-713b-11eb-9b34-7114ed56a96a.png)

I don't have Visual Studio 2017 (yet) and am not that familiar with C++, but it seems this boolean is what I'm looking for!

torepaulsson:
Closed issue since the library now supports the flag using the SendMessageEx method and the MessageReceivedEx event.

mikhaylo1:
@torepaulsson would you mind to publish **mrwebrtc.dll** with that fix ? 
I have the same problem you described - the peer I am connecting to doesn't accept blob DataChannel messages from MixedReality-WebRTC 2.0.2 and expects text only

It looks like there is only win32-x64-Debug version published https://dev.azure.com/aipmr/MixedReality-WebRTC-CI/_build/results?buildId=19141&view=artifacts&pathAsName=false&type=publishedArtifacts

Very interested in win32-x64-Release build and failed to build it myself

torepaulsson:
[...]

PR #675, opened 2021-02-19
Add text message support for data channel send and receive callbacks.

torepaulsson:
Finally got around to this update. I've updated the solution to not be a breaking change. Added new method for sending text messages over the data channel.
Added a new event which is fired when non-binary data is received.

This will still be a change in the inner workings of the mr-webrtc dll. I've opened up an issue to discuss what can be done about it below!

### DIFF
--- a/libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
+++ b/libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
@@ -59,6 +59,22 @@ namespace Microsoft.MixedReality.WebRTC
         }
 
         /// <summary>
+        /// Type of message sent or received through the data channel.
+        /// </summary>
+        public enum MessageKind
+        {
+            /// <summary>
+            /// The message is a binary message.
+            /// </summary>
+            Binary = 1,
+
+            /// <summary>
+            /// The message is a text message (UTF-8 encoded string).
+            /// </summary>
+            Text = 2
+        }
+
+        /// <summary>
         /// Delegate for the <see cref="BufferingChanged"/> event.
         /// </summary>
         /// <param name="previous">Previous buffering size, in bytes.</param>
@@ -128,7 +144,7 @@ namespace Microsoft.MixedReality.WebRTC
         /// calls to <see cref="SendMessage(byte[])"/> do not fail. Internally the data channel contains
         /// a buffer of messages to send that could not be sent immediately, for example due to
         /// congestion control. Once this buffer is full, any further call to <see cref="SendMessage(byte[])"/>
-        /// will fail until some mesages are processed and removed to make space.
+        /// will fail until some messages are processed and removed to make space.
         /// </summary>
         /// <remarks>
         /// The code handling this event should unwind the stack before
@@ -142,6 +158,13 @@ namespace Microsoft.MixedReality.WebRTC
         /// </summary>
         /// <seealso cref="SendMessage(byte[])"/>
         public event Action<byte[]> MessageReceived;
+
+        /// <summary>
+        /// Event triggered when a message is received through the data channel,
+        /// includes information if the message is binary or text.
+        /// </summary>
+        /// <seealso cref="SendMessage(byte[])"/>
+        public event Action<MessageKind, byte[]> MessageReceivedEx;
 
         /// <summary>
         /// Event fires when a message is received through the data channel.
@@ -229,6 +252,32 @@ namespace Microsoft.MixedReality.WebRTC
             Utils.ThrowOnErrorCode(res);
         }
 
+        /// <summary>
+        /// Send a message through the data channel with the specified kind. If the message cannot be sent, for example because of congestion
+        /// control, it is buffered internally. If this buffer gets full, an exception is thrown and this call is aborted.
+        /// The internal buffering is monitored via the <see cref="BufferingChanged"/> event.
+        /// </summary>
+        /// <param name="messageKind">The kind of message to send to the remote peer.</param>
+        /// <param name="message">The message to send to the remote peer.</param>
+        /// <param name="size">The size of the message to send in octects.</param>
+        /// <exception xref="System.InvalidOperationException">The native data channel is not initialized.</exception>
+        /// <exception xref="System.Exception">The internal buffer is full.</exception>
+        /// <exception cref="DataChannelNotOpenException">The data channel is not open yet.</exception>
+        /// <seealso cref="PeerConnection.InitializeAsync"/>
+        /// <seealso cref="BufferingChanged"/>
+        public void SendMessageEx(MessageKind messageKind, IntPtr message, ulong size)
+        {
+            MainEventSource.Log.DataChannelSendMessage(ID, (int)size);
+            // Check channel state before doing a P/Invoke call which would anyway fail
+            if (State != ChannelState.Open)
+            {
+                throw new DataChannelNotOpenException();
+            }
+
+            uint res = DataChannelInterop.DataChannel_SendMessageEx(_nativeHandle, messageKind, message, size);
+            Utils.ThrowOnErrorCode(res);
+        }
+
         internal void OnMessageReceived(IntPtr data, ulong size)
         {
             MainEventSource.Log.DataChannelMessageReceived(ID, (int)size);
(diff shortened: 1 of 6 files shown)


djee-ms:
Hi Tore, thanks for your contribution. Unfortunately I don't think we'll be able to take that change as is:

- I have yet to understand what's the story with this binary flag. First, very practically, nobody ever complained about it, everyone as I know is passing data around as binary even if the content is some text. I don't immediately see what would be the benefit of telling the protocol that your data is a string, as opposed to leaving it as an opaque binary blob (which can perfectly hold a string as far as I can tell; I know some users are passing JSON data without any issue). Second, more theoretically, I have looked at the RFC and as far as I can tell the protocol allows you to tell it that you send text, but doesn't seem to otherwise care about it (except for the fact this guarantees UTF-8), and therefore I don't see why one would bother informing the protocol about this if it doesn't change anything. Really the only case I see for this (and incidentally this is the same story with the WebSocket protocol) is to send and receive data between peers that have zero priori knowledge nor implicit agreement on the content, and use the protocol to enforce UTF-8 encoding. But as soon as you assume any encoding, then my understanding is that binary is completely equivalent.

- This is a breaking change that essentially breaks the entire data channel API, and so far I have not seen compelling reason to do so. And you mentioned yourself a non-API-breaking alternative.

If you can explain why this is so critical and why sending text in a binary blob is impractical for your scenario, then we can consider a change. Your issue #673 states that there's no other way, but as explained I am fairly confident encoding the string as UTF-8 (or anything else, really) yourself and sending it as binary should work, unless you have a specific remote endpoint implementation with other restrictions (and if so, which one?). And if we take a change anyway, which I am open to do if that can help unblock you, I would really want to avoid an API breaking and instead would strongly prefer having an incremental change with a new overload specifically for text data, as you suggested.

torepaulsson:
Hi Jerome, thanks for your quick reply!
I agree that this cannot be included if the feature create a breaking change! I will update the draft with non-breaking changes. I will keep the nuget package as is and only add overloads+one event for the text data. I will also add methods and events in the mr-webrtc so that also does not get any breaking changes.

This feature is really not a real use case for this library I guess. The issue I have is that I use it for communicating with another library, webrtc in gstreamer, in the implementation I have the commands are sent using the text messages and the data is sent with the binary flag.

I agree that practically it makes little sense to have this flag, bytes are bytes, and as long as the two parties that communicate agree on a protocol everything is fine, but I cannot change the other party in this instance :)

libs/mrwebrtc/src/data_channel.cpp
```diff
@@ -96,11 +112,16 @@ void DataChannel::OnStateChange() noexcept {
...
   std::lock_guard<std::mutex> lock(mutex_);
-  if (message_callback_) {
-    message_callback_(buffer.data.data(), buffer.data.size());
+  if (buffer.binary) {
+    if (message_callback_) {
+      message_callback_(buffer.data.data(), buffer.data.size());
+    }
+  }else if (text_message_callback_) {
+    text_message_callback_(buffer.data.data(), buffer.data.size());
   }
```
torepaulsson, line 121: This is where we change how the library works.
Before we would always fire the message_callback_, but now it is only done when we have binary data.

I am thinking, if we want this to work just as it was before, we should maybe add the following logic,
if text_message_callback_ is not registered, always fire the message_callback. This would ensure that old implementations are still working.

if text_message_callback_ is registered, then we can split on binary data and fire the different events.

I feel it is some strange logic, but it would keep the library working just as it had before my change. What do you think @djee-ms ?

djee-ms:
Hi @taurepaulsson,

Sorry for the delay. We have been talking internally with @fibann about this change, as I had mixed feelings about it and how it's currently implemented. And I think we agree that we'd want instead to leave the current behavior as is, and add an overload (say, `SendMessageEx` / `MessageReceivedEx`) that has the binary/text flag exposed. That way users could still use the current API as is, or could use this new API to send/receive a mix of text and binary messages, and know about it. There are 2 related reasons for wanting this:

- Your current implementation actually does change the behavior, because the `MessageReceived` callback is made to only be called for binary message, whereas it was previously called for all messages. To that end, this is a breaking change. For an existing user upgrading to a version with your change, they would suddenly lose some messages unless they also subscribe to the new `TextMessageReceived` callback.

- More generally, adding a functionality via an overload / separate method without touching any of the existing API feels like a safer approach to ensure this is an incremental feature and not a breaking change.

Would you mind giving it a try with that approach? I looked at your implementation otherwise and there was nothing else really to change in terms of code quality or missing comment or anything I think, just this change in behavior of `MessageReceived` that we'd like to avoid.

Thanks!

torepaulsson:
Hi @djee-ms 
No worries, we all have other things to do as well, I'm very glad that you help me out with this. I agree about your proposal regarding the MessageReceivedEx event.

Regarding the change in the implementation I've tried to add it that the event is always fired, unless you register to the text message callback. So it should not be a breaking change now But I will change the event to function as you say instead, create a extended version and always fire both events if registered, a lot easier to understand!

Regarding the SendMessage I feel that it is nice that it takes a string instead of a byte array? What do you think? It wont be breaking if I just add a SendMessage(string message) overload to the existing SendMessage method. If you feel strong about the SendMessageEx I can create that as well! Just wanted to ask.

djee-ms:
> I've tried to add it that the event is always fired, unless you register to the text message callback. So it should not be a breaking change 

You're right, I missed that sorry. I still prefer the separate overload which avoids a behind-the-scene change of behavior of the event depending on some other state, I think it's easier to understand for users.

> Regarding the SendMessage I feel that it is nice that it takes a string instead of a byte array?

Sorry this was a bit confusing. I was talking about the C interop API, where you cannot/shouldn't use overloads, _e.g._ `mrsDataChannelSendMessageEx()`. Similarly we want I think an event `mrsDataChannelMessageExCallback` similar to the existing one but with an extra parameter for the text/binary kind. With those two, we can implement all the overloads we want in any language.

Now, for the C# library there are already several `Send` overloads:
- `SendMessage(byte[] message)` is the historical one, most convenient for binary data.
- `SendMessage(IntPtr message, ulong size)` is an optimized overload that prevent allocation, because the `byte[]` array above was forcing developers to allocate an array in some circumstances.

My proposal is to add:
- `SendMessageEx(MessageKind messageKind, IntPtr message, ulong size)` which would be the "raw" overload directly mapping to `mrsDataChannelSendMessageEx()`.
- `SendMessage(string textMessage)` which I agree with you is nice to have, but should clearly document it sends a _text_ message only, not binary.

For the receive event, we should also add an `Ex` handler I guess. I don't see a better way:
```diff
  public event Action<byte[]> MessageReceived;
+ public event Action<MessageKind, byte[]> MessageReceivedEx;
```

Side note too, as shown above I'd avoid `bool` and use an enum to be 100% explicit about the message kind, something like:
```cs
enum MessageKind
{
    Binary,
    Text
}
```

How does that sound @torepaulsson? Also ping @fibann if he wants to chime in.

torepaulsson:
Hi @djee-ms 
Thanks :) I'll give it a go!

torepaulsson:
Hi @djee-ms,
Lot of meetings today but I've changed the implementation to what you suggested. I wondered a bit if you wanted an enum also in the C++ mrwebrtc project, I included one but maybe it would be easier to implement interops by using the boolean flag.

Hopefully I've gotten it the way you wanted :)

libs/Microsoft.MixedReality.WebRTC/DataChannel.cs
```diff
@@ -58,6 +58,22 @@ public enum ChannelState
...
+        {
+            /// <summary>
+            /// The message is a binary message.
+            /// </summary>
+            Binary = 1,
+
+            /// <summary>
+            /// The message is a text message.
+            /// </summary>
+            Text = 2
```
fibann, line 74: Should specify what "text message" means (UTF-8 string).

torepaulsson: I've updated the summary, see 70af869ce60cc2bda6068c0abfc4af5fded8c348

fibann (approved):

djee-ms (approved):

djee-ms:
All good for me. @torepaulsson if you want to add your name to the [CONTRIBUTORS](https://github.com/microsoft/MixedReality-WebRTC/blob/master/CONTRIBUTORS) list you can update this PR with it. Otherwise I'll merge it as is.

torepaulsson:
> All good for me. @torepaulsson if you want to add your name to the [CONTRIBUTORS](https://github.com/microsoft/MixedReality-WebRTC/blob/master/CONTRIBUTORS) list you can update this PR with it. Otherwise I'll merge it as is.

Hi, I've rebased so the fixup commit is included in the first commit. It is OK, no need to put me as a contributor :)

djee-ms:
Merged, thanks for your contribution @torepaulsson !!

Qualcuno:
Good job. Sending text is very important if a peer is on another system such as an SFU which distinguishes between text messages and binary messages. When will this feature be available in the nuget package? https://www.nuget.org/packages/Microsoft.MixedReality.WebRTC.UWP

torepaulsson:
@Qualcuno, I don't know, I'm curious as well though!
@djee-ms, do you know the plans for a future release?

PR #680, opened 2021-02-22
Update helloworld-cs-connection-core3.md

ThaiVan:
Take a note for the issue 678.


docs/manual/cs/helloworld-cs-connection-core3.md
```diff
@@ -63,3 +63,6 @@ Run the two instances of the application again. This time both terminals print a
 If launched with the audio or video capture flags, the capturer instance records those media and send them via the network to the other instance, which invokes the remote frame callback and print a message every 60 frames. After that, you can press any key to stop each instance. The signaler and peer connection will close and the program will terminate.
 
 ![Peer connections connecting to each other](cs7.png)
+
+Note: From the version 2.0, there is some changes in the way a transceiver is added. Therefore, according to the tutorial context, the client which creates an offer PeerConnection.CreateOffer() should be starting with Video capture, -a or -audio argument, then you can see video datas are sent from the client to the server. Otherwise, no video datas are transfered from the client to the server, or viceversa. This is disscused in the issue id 678.
+This
```
djee-ms, line 68: Thanks for the contribution. There are a few issues though:
- This is partially incorrect: there's no transceiver before v2.x so there's no "change in the way a transceiver is added", at least not from the point of view of users. So I would rephrase or avoid altogether calling that a change.
- Typo empty paragraph "This" at the end
- More generally the English wording is not ideal and the form has some issues: "Video" shouldn't be capitalized; audio is mentioned via its arguments but video is not, for some reason, creating an imbalance; the sentence starts in passive form then switch to active form; grammar issue at the end and missing issue link.

Please consider this alternative:
```
_Note_: Starting from v2.0 and the introduction of explicit transceiver objects and support for multiple tracks, users have to be careful to add transceivers on the offering side alone (they will be created on the answering side automatically), as there is no automated pairing of existing transceivers in WebRTC. Therefore in this tutorial the offering peer, which is the instance started last (client/caller), needs to be the one which will send audio and/or video (`-a`/`--audio` and `-v`/`--video` flags). Otherwise the caller will not add any transceiver, and the callee (server; the instance started first) will not be able to send anything. This is only an issue if the callee wants to send some media but the caller doesn't. If they both send, then the caller will add the transceiver and the callee will use it.
```

...which formats to:

> _Note_: Starting from v2.0 and the introduction of explicit transceiver objects and support for multiple tracks, users have to be careful to add transceivers on the offering side alone (they will be created on the answering side automatically), as there is no automated pairing of existing transceivers in WebRTC. Therefore in this tutorial the offering peer, which is the instance started last (client/caller), needs to be the one which will send audio and/or video (`-a`/`--audio` and `-v`/`--video` flags). Otherwise the caller will not add any transceiver, and the callee (server; the instance started first) will not be able to send anything. This is only an issue if the callee wants to send some media but the caller doesn't. If they both send, then the caller will add the transceiver and the callee will use it.

djee-ms (changes requested):

docs/manual/cs/helloworld-cs-connection-core3.md
ThaiVan: Thank you. I learned from your review.

djee-ms (approved):

PR #698, opened 2021-03-09
NativeRender Improvements.

rygo6-MS:
* Fixes multiple crashes in NativeRenderer native library.
* Enables a stream to change resolution mid-stream.
* Configure VS project to build UnityPlugin for ARM UWP.


libs/mrwebrtc-unityplugin/include/api.h
```diff
@@ -64,6 +69,9 @@ mrsNativeRenderer_GetVideoUpdateMethod() noexcept;
 /// Pipe log entries to Unity's log.
 MR_UNITYPLUGIN__API void MR_UNITYPLUGIN__CALL
 mrsNativeRenderer_SetLoggingFunctions(UnityLogger::LogFunction logDebugFunc,
-                                      UnityLogger::UnityLogger::LogFunction logErrorFunc,
+                                      UnityLogger::LogFunction logErrorFunc,
                                       UnityLogger::LogFunction logWarningFunc);
+
+MR_UNITYPLUGIN__API void MR_UNITYPLUGIN__CALL
+mrsNativeRenderer_SetTextureSizeChangeCallback(TextureSizeChangeCallback callback) noexcept;
```
djee-ms, line 76: I'd rename to `TextureSizeChanged` (both function and type), I think event names generally use the past tense to indicate the event is raised after the underlying action occurred.

libs/mrwebrtc-unityplugin/src/api.cpp
```diff
@@ -87,8 +87,16 @@ mrsNativeRenderer_EnableRemoteVideo(mrsNativeVideoHandle nativeVideoHandle,
...
-  nativeVideo->EnableRemoteVideo(format, textures, textureCount);
+  nativeVideo->SetRemoteVideoTextures(format, textures, textureCount);
+  return Result::kSuccess;
+}
+
+mrsResult MR_UNITYPLUGIN__CALL
+mrsNativeRenderer_UpdateTextures(mrsNativeVideoHandle nativeVideoHandle,
+                                 TextureDesc textures[]) noexcept {
+  UnityLogger::LogDebug("mrsNativeRenderer_UpdateTextures");
+  NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(nativeVideoHandle);
```
djee-ms, line 98: Can we use the common error checking pattern here?
```cpp
if (NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(nativeVideoHandle)) {
  // do something
  return Result::kSuccess;
}
return Result::kInvalidNativeHandle;
```

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -39,12 +40,19 @@ void I420VideoFrame::CopyFrame(const mrsI420AVideoFrame& frame) {
   size_t ysize = (size_t)ystride * height;
   size_t usize = (size_t)ustride * height / 2;
```
djee-ms, line 41: Should we use `(height + 1) / 2` to round up and avoid truncating chroma for the last line here?

libs/mrwebrtc-unityplugin/src/api.cpp
```diff
@@ -87,8 +87,16 @@ mrsNativeRenderer_EnableRemoteVideo(mrsNativeVideoHandle nativeVideoHandle,
                                     int textureCount) noexcept {
   UnityLogger::LogDebug("mrsNativeRenderer_EnableRemoteVideo");
   NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(nativeVideoHandle);
-  //NativeRenderer* nativeVideo = (NativeRenderer*)nativeVideoHandle;
-  nativeVideo->EnableRemoteVideo(format, textures, textureCount);
+  nativeVideo->SetRemoteVideoTextures(format, textures, textureCount);
```
djee-ms, line 90: This can throw due to `resize()`. Either it should be caught inside the method, and the method marked as `noexcept`, or caught here, to avoid propagating exceptions and for correctness since this function is marked as `noexcept` itself.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -79,17 +87,22 @@ NativeRenderer::~NativeRenderer() {
   // Log_Debug("NativeRenderer::~NativeRenderer");
 }
 
+void NativeRenderer::SetTextureSizeChangeCallback(TextureSizeChangeCallback textureSizeChangeCallback) {
```
djee-ms, line 90: Add `noexcept`

libs/mrwebrtc-unityplugin/src/api.cpp
```diff
@@ -87,8 +87,16 @@ mrsNativeRenderer_EnableRemoteVideo(mrsNativeVideoHandle nativeVideoHandle,
...
+  return Result::kSuccess;
+}
+
+mrsResult MR_UNITYPLUGIN__CALL
+mrsNativeRenderer_UpdateTextures(mrsNativeVideoHandle nativeVideoHandle,
+                                 TextureDesc textures[]) noexcept {
+  UnityLogger::LogDebug("mrsNativeRenderer_UpdateTextures");
+  NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(nativeVideoHandle);
+  nativeVideo->UpdateTextures(textures);
```
djee-ms, line 99: This can throw. See other similar comment.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
 
 void NativeRenderer::I420ARemoteVideoFrameCallback(
-    void* user_data,
-    const mrsI420AVideoFrame& frame) {
+  void* user_data,
+  const mrsI420AVideoFrame& frame) {
 
-    NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(user_data);
+  // It is possible for one buffer to be empty, each buffer must be checked.
+  if (frame.ydata_ == nullptr || frame.udata_ == nullptr || frame.vdata_ == nullptr) return;
```
djee-ms, line 163: Avoid single-line `if`, always use braces.
```suggestion
  if (frame.ydata_ == nullptr || frame.udata_ == nullptr || frame.vdata_ == nullptr) {
    return;
  }
```

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
 
-    // RESEARCH: Do we need to keep a frame queue or is it fine to just render
-    // the most recent frame?
+  NativeRenderer* nativeVideo = static_cast<NativeRenderer*>(user_data);
+ 
+  {
+    // Instance lock in case they get updated from UpdateTextures
+    std::lock_guard guard(nativeVideo->m_lock);
+    if ((int)frame.width_ != nativeVideo->m_remoteTextures[0].width || (int)frame.height_ != nativeVideo->m_remoteTextures[0].height) {
+      Log_Warning("NativeRenderer: I420 frame resolution changed from what it was initialized with. Resizing.");
```
djee-ms, line 171: Do we need a warning here since we have a callback to inform the user? WebRTC changing the video resolution mid-call is a feature and a common action, so it feels weird to log a warning like if something bad or unexpected had happened.
Also, we should check that `g_textureSizeChangeCallback` is not `nullptr`.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
-    // copy operation could be done inside g_lock scope. This would result in a
-    // single g_lock, but the lock would be held for a longer period of time
-    // every video frame, for every video stream. RESEARCH: Which is better?
+  // Acquire a video frame buffer from the free list with the global lock.
+  std::shared_ptr<I420VideoFrame> remoteI420Frame = nullptr;
+  {
+    // Global lock
+    std::lock_guard guard(g_lock);
+    if (g_freeI420VideoFrames.size()) {
+        remoteI420Frame = g_freeI420VideoFrames.back();
```
djee-ms, line 183: nit: move to save 2 ref-count updates
```suggestion
        remoteI420Frame = std::move(g_freeI420VideoFrames.back());
```

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
+        g_freeI420VideoFrames.pop_back();
+    } else {
+        remoteI420Frame = std::make_shared<I420VideoFrame>();
+    }
+  }
 
-    // Acquire a video frame buffer from the free list.
+  // I don't see how this could happen, but it was here so I will leave it.
+  if (remoteI420Frame == nullptr) {
```
djee-ms, line 191: I guess the intention was to check a failed alloc? But then we should catch `std::bad_alloc` instead rather than this.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
+  }
+
+  // Copy frame outside of any lock for perf.
+  // This may consume more time on this thread unnecessarily, but
+  // will block the render update thread for less time.
+  bool frameCopySuccess = remoteI420Frame->CopyFrame(frame);
+
+  bool shouldUpdateFrame = false;
+  if (frameCopySuccess) {
     {
```
djee-ms, line 203: Why the double `{ }` block here? The `if { }` block already scopes the `std::lock_guard`.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +156,78 @@ void NativeRenderer::DisableRemoteVideo() {
...
+      // Global lock
+      std::lock_guard guard(g_lock);
+      g_nativeVideoUpdateQueue.emplace(nativeVideo);
     }
+  } else {
+    // If frame copy failed or was unnecessary, recycle frame
+    {
+      // Global lock
+      std::lock_guard guard(g_lock);
+      g_freeI420VideoFrames.push_back(remoteI420Frame);
```
djee-ms, line 228: nit: same as before
```suggestion
      g_freeI420VideoFrames.push_back(std::move(remoteI420Frame));
```

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
```diff
@@ -27,7 +29,15 @@ public class TextureDesc
...
     {
-        // private IntPtr _videoHandle;
+        /// <summary>
+        /// Callback for when the texture size of stream changes in a NativeVideo.
+        /// This callback will not occur on the main thread.
+        /// Unity must handle this resize, it cannot be done entirely in DX11.
+        /// </summary>
+        public event TextureSizeChangeCallback TextureSizeChanged;
+
+        private static Dictionary<IntPtr, NativeVideo> lookupDictionary = new Dictionary<IntPtr, NativeVideo>();
```
djee-ms, line 39: `_lookupDictionary` for private variables (`_` prefix)

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
```diff
@@ -36,7 +46,9 @@ internal class NativeVideo : IDisposable
         /// <param name="peerConnection"></param>
         public NativeVideo(IntPtr videoHandle)
         {
-	        _nativeVideoHandle = NativeVideoInterop.Create(videoHandle);
+            _nativeVideoHandle = NativeVideoInterop.Create(videoHandle);
+            if (lookupDictionary.ContainsKey(videoHandle)) lookupDictionary[videoHandle] = this;
+            else lookupDictionary.Add(videoHandle, this);
```
djee-ms, line 51: This is enough, [if the key doesn't exist it gets created](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.item?view=netstandard-2.0):
```suggestion
            lookupDictionary[videoHandle] = this;
```

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
```diff
@@ -84,13 +96,26 @@ public void EnableRemoteVideo(VideoKind format, TextureDesc[] textures)
...
+            NativeVideoInterop.UpdateTextures(_nativeVideoHandle, interopTextures);
+        }
 
         /// <summary>
         /// Stops rendering the remote video stream.
         /// </summary>
         public void DisableRemoteVideo()
         {
             NativeVideoInterop.DisableRemoteVideo(_nativeVideoHandle);
+            if (lookupDictionary.ContainsKey(_nativeVideoHandle)) lookupDictionary.Remove(_nativeVideoHandle);
```
djee-ms, line 118: No need to check for existing key, [`Remove()` will fail returning `false` if it doesn't exist](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove?view=netstandard-2.0).
```suggestion
            lookupDictionary.Remove(_nativeVideoHandle);
```

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
```diff
@@ -101,6 +126,20 @@ public static IntPtr GetVideoUpdateMethod()
         {
             return NativeVideoInterop.GetVideoUpdateMethod();
         }
+        
+        /// Sets callback handlers for the logging of debug, warning, and error messages.
```
djee-ms, line 130: Wrong copy/pasted comment

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -42,12 +47,15 @@ private void OnDestroy()
...
-                {
-                    StartNativeRendering((int)frame.Width, (int)frame.Height);
-                }
+                CreateEmptyVideoTextures(dirtyWidth, dirtyHeight, 128);
+                _nativeVideo.UpdateTextures(GetTextureDescArray());
+            }
+            
+            if (_nativeVideo == null && _i420AFrameQueue.TryDequeue(out I420AVideoFrameStorage frame))
+            {
+                StartNativeRendering((int)frame.Width, (int)frame.Height);
```
djee-ms, line 58: We're discarding the frame here?

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -75,19 +83,29 @@ public void StartRendering(IVideoSource source)
 
         private void I420AVideoFrameReady(I420AVideoFrame frame)
         {
-            _i420aFrameQueue?.Enqueue(frame);
+            _i420AFrameQueue?.Enqueue(frame);
```
djee-ms, line 86: It's never NULL anymore, is it? Remove `?`?

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -116,41 +134,51 @@ private void TearDown()
             }
             _source = null;
         }
-
+
```
djee-ms, line 137: nit: remove spaces

tools/build/mrwebrtc-unityplugin/uwp/mrwebrtc-unityplugin-uwp.vcxproj
```diff
@@ -0,0 +1,182 @@
...
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\render_api.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\targetver.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityGraphics.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityGraphicsD3D11.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityGraphicsD3D12.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityGraphicsD3D9.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityGraphicsMetal.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\Unity\IUnityInterface.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\video_types.h" />
+    <ClInclude Include="..\..\..\..\libs\mrwebrtc-unityplugin\include\export.h" />
```
djee-ms, line 176: Path should use `$(MRWebRTCProjectRoot)`

tools/build/mrwebrtc-unityplugin/uwp/mrwebrtc-unityplugin-uwp.vcxproj.filters
```diff
@@ -0,0 +1,54 @@
...
+    </ClInclude>
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\targetver.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\include\api.h">
+      <Filter>interop</Filter>
+    </ClInclude>
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\log_helpers.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\native_renderer.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\render_api.h" />
+    <ClInclude Include="$(MRWebRTCProjectRoot)libs\mrwebrtc-unityplugin\src\PlatformBase.h" />
+    <ClInclude Include="..\..\..\..\libs\mrwebrtc-unityplugin\include\export.h" />
```
djee-ms, line 44: Same, use `$(MRWebRTCProjectRoot)`

djee-ms (changes requested):

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
rygo6-MS: Not sure, it seems to have been working like this without issue?

rygo6-MS: I just tried adding the height+1, but noticed at the top of the frame it adds a 1 pixel horizontal line of distorted pixels on 1080p. I assume that means it is correct how it currently is.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
djee-ms: Can't we start the renderer AND also enqueue the frame, to not lose it? Or is it not worth the complication?

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
djee-ms: That sounds more like a double bug to me. If the height is odd, the last Y line still needs UV chroma. As is, we basically only have luminance for that last line. Sounds more to me like changing to `(height + 1) / 2` uncovered another bug somewhere else.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: The issue is that unity must know the texture resolution to then create the texture and send it to the native side, and that texture creation has to be done from the unity main thread. 

So it gets kind of odd as you'd have to wait for the unity thread to respond after the first frame.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
rygo6-MS: Or maybe the line of pixels at the top is correct and the issue is in my stream? It's just copying the I420AVideoFrame struct directly.

Your saying it should be like this, correct?
```
  size_t ysize = (size_t)ystride * height;
  size_t usize = (size_t)ustride * (height + 1) / 2;
  size_t vsize = (size_t)vstride * (height + 1) / 2;
```

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
rygo6-MS: I could make it just rely on the UpdateTextures method to resize it, would make the C# simpler.

rygo6-MS:
Sorry didn't notice your inline comments before, just pushed those fixes as well.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
rygo6-MS: So I did some more comparisons on some different streams. The difference may actually be fairly negligible, see attached photo. So if +1 is more technically correct I think that is fine. I have committed it with height + 1.
compare

libs/mrwebrtc-unityplugin/include/api.h
```diff
@@ -12,6 +12,7 @@ extern "C" {
 
 using mrsResult = Microsoft::MixedReality::WebRTC::Result;
 using mrsNativeVideoHandle = mrsObjectHandle;
+typedef void (*TextureSizeChangeCallback)(int width, int height, mrsRemoteVideoTrackHandle handle); // would like to know how to put this in native_renderer.h
```
djee-ms, line 15: Can you please rename that one, at least with `mrs` prefix, and ideally also with `Changed` instead of `Change` for consistency with callback naming? Thanks!

libs/mrwebrtc-unityplugin/include/api.h
```diff
@@ -49,6 +50,10 @@ mrsNativeRenderer_EnableRemoteVideo(mrsNativeVideoHandle nativeVideoHandle,
                                     TextureDesc textures[],
                                     int textureCount) noexcept;
 
+MR_UNITYPLUGIN__API mrsResult MR_UNITYPLUGIN__CALL
+mrsNativeRenderer_UpdateTextures(mrsNativeVideoHandle nativeVideoHandle,
+                                 TextureDesc textures[]) noexcept;
```
djee-ms, line 55: Is it worth passing an array size, or are we expected always exactly 3 textures? If the latter, can you please add a comment?

djee-ms: Ah that's even more confusing because `mrsNativeRenderer_EnableRemoteVideo()` does have a `textureCount`.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -29,22 +29,30 @@ static std::set<mrsNativeVideoHandle> g_nativeVideoUpdateQueue;
 static std::vector<std::shared_ptr<I420VideoFrame>> g_freeI420VideoFrames;
 static std::vector<std::shared_ptr<ArgbVideoFrame>> g_freeArgbVideoFrames;
 std::set<mrsNativeVideoHandle> NativeRenderer::g_nativeVideos;
+NativeRenderer::TextureSizeChangeCallback NativeRenderer::g_textureSizeChangeCallback = nullptr;
 
-void I420VideoFrame::CopyFrame(const mrsI420AVideoFrame& frame) {
+bool I420VideoFrame::CopyFrame(const mrsI420AVideoFrame& frame) {
```
djee-ms, line 34: This should probably be `noexcept`

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +167,74 @@ void NativeRenderer::DisableRemoteVideo() {
...
+  // Acquire a video frame buffer from the free list with the global lock.
+  std::shared_ptr<I420VideoFrame> remoteI420Frame = nullptr;
+  {
+    // Global lock
+    std::lock_guard guard(g_lock);
+    if (g_freeI420VideoFrames.size()) {
+        remoteI420Frame = std::move(g_freeI420VideoFrames.back());
+        g_freeI420VideoFrames.pop_back();
+    } else {
+      try {
```
djee-ms, line 200: nit: move that block outside the lock, there's no need for holding it while doing some expensive allocation.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
```diff
@@ -133,45 +167,74 @@ void NativeRenderer::DisableRemoteVideo() {
...
+  bool frameCopySuccess = remoteI420Frame->CopyFrame(frame);
+
+  bool shouldUpdateFrame = false;
+  if (frameCopySuccess) {
+    // Set new copied frame on nativeVideo while in the instance lock of the
+    // NativeVideo otherwise the render update loop may grab it pre-emptively.
+    std::lock_guard guard(nativeVideo->m_lock);
+    // Only update if null, otherwise this means the frame previously
+    // added has not yet been processed.
+    if (nativeVideo->m_nextI420RemoteVideoFrame == nullptr) {
```
djee-ms, line 221: Are we saying here that we drop the current frame in favor of keeping an old frame? Shouldn't we do the opposite?

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
```diff
@@ -84,23 +95,50 @@ public void EnableRemoteVideo(VideoKind format, TextureDesc[] textures)
 
             NativeVideoInterop.EnableRemoteVideo(_nativeVideoHandle, format, interopTextures, interopTextures.Length);
         }
+        
+        public void UpdateTextures(TextureDesc[] textures)
+        {
+            var interopTextures = textures.Select(item => new NativeVideoInterop.TextureDesc
```
djee-ms, line 101: Related to other comment, should we validate the size of `textures[]` here? If it's smaller than 3 we will have an out-of-bound access in native.

tools/build/mrwebrtc-unityplugin/uwp/mrwebrtc-unityplugin-uwp.vcxproj
```diff
@@ -0,0 +1,182 @@
...
+  </ItemDefinitionGroup>
+  <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
+    <ClCompile>
+      <Optimization>Disabled</Optimization>
+      <PreprocessorDefinitions>_DEBUG;_CRTDBG_MAP_ALLOC;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
+    </ClCompile>
+    <Link>
+      <GenerateDebugInformation>true</GenerateDebugInformation>
+      <OutputFile Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(OutDir)$(TargetName)$(TargetExt)</OutputFile>
```
djee-ms, line 130: Remove `Condition` here?

djee-ms (changes requested):

djee-ms:
A few more minor points but should be good to go after that, thanks!

libs/mrwebrtc-unityplugin/include/api.h
rygo6-MS: I believe EnableRemoteVideo was set up like this in anticipation of RGBA implementation having a different texture count. UpdateTextures will probably need textureCount added when RGBA is supported.

libs/mrwebrtc-unityplugin/src/native_renderer.cpp
rygo6-MS: Yes, that is probably more correct.

libs/mrwebrtc-unityplugin/include/api.h
rygo6-MS: I ended up working this a bit so EnableRemoteVideo native method no longer deals with setting textures it only subscribes and now only UpdateRemoteTextures sets textures. The functionality was essentially a duplication.

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideo.cs
rygo6-MS: Added validation method.

tools/build/mrwebrtc-unityplugin/uwp/mrwebrtc-unityplugin-uwp.vcxproj
rygo6-MS: Note sure myself. Does allow debug information in a release build impact performance in anyway?

djee-ms: What?

I mean replace:
```suggestion
      <OutputFile>$(OutDir)$(TargetName)$(TargetExt)</OutputFile>
```

I don't see why the output path would be only set in 1 config and not the others.

djee-ms (approved):

libs/mrwebrtc-unityplugin/include/api.h
```diff
@@ -45,9 +46,13 @@ mrsNativeRenderer_Destroy(mrsNativeVideoHandle nativeVideoHandle) noexcept;
 //// subscribed to the FrameReady call back on the VideoTrack.
 MR_UNITYPLUGIN__API mrsResult MR_UNITYPLUGIN__CALL
 mrsNativeRenderer_EnableRemoteVideo(mrsNativeVideoHandle nativeVideoHandle,
-                                    VideoKind format,
-                                    TextureDesc textures[],
-                                    int textureCount) noexcept;
+                                    VideoKind format) noexcept;
+
+MR_UNITYPLUGIN__API mrsResult MR_UNITYPLUGIN__CALL
+mrsNativeRenderer_UpdateRemoteTextures(mrsNativeVideoHandle nativeVideoHandle,
```
fibann, line 52: Can you add a comment to this and SetTextureSizeChanged below?

libs/unity/library/Runtime/Scripts/NativeRender/NativeVideoRenderer.cs
```diff
@@ -24,8 +26,9 @@ public class NativeVideoRenderer : MonoBehaviour
         private Texture2D _textureU;
         private Texture2D _textureV;
 
-        private VideoFrameQueue<I420AVideoFrameStorage> _i420aFrameQueue;
-
+        private int dirtyWidth;
+        private int dirtyHeight;
```
fibann, line 30: Should use the `_` prefix for private fields, so `_dirtyWidth`, `_dirtyHeight`. Also might benefit from a comment saying what they are, "dirty" implies that it's used when res changes but it's not clear if this is the old res or the new res.

fibann (approved):
Just a couple of minor things to add/fix.

tools/build/mrwebrtc-unityplugin/uwp/mrwebrtc-unityplugin-uwp.vcxproj
rygo6-MS: Oh, I thought you meant the whole line

rygo6-MS:
I believe everything is done. Is there anything I missed?

djee-ms:
Merged, thanks @rygo6-MS for the work!

PR #759, opened 2021-05-18
NodeDssSignaler: Ensure that the local peer ID is a valid SDP token

nmlgc:
If no local peer ID is set, NodeDssSignaler derives it from the device ID. Apparently, it's quite common for them on Android to contain spaces, which aren't allowed in SDP tokens. If we fix this, we might as well go all the way.


fibann (approved):
Thanks for the contribution @nmlgc!

Issue #722, opened 2021-03-26
Removing DataChannels causes InvalidOperationException


torepaulsson:
**Describe the bug**
When calling PeerConnection.RemoveDataChannel an InvalidOperationException is thrown.

**To Reproduce**
1. Go to 'TestNetCoreConsole' project
3. Add the following code after the pc.Initialize, after line 49 in Program.cs
```csharp
Console.WriteLine("Adding data channel.");
await pc.AddDataChannelAsync("the_data_channel", true, true, CancellationToken.None);
```
4. Add the following code when stopping the recording, after line 128 in Program.cs
```csharp
   Console.WriteLine("Removing data channels...");
   foreach (var dataChannel in pc.DataChannels.ToImmutableArray())
   {
      pc.RemoveDataChannel(dataChannel);
   }
```
5.  See error thrown is caught in the general exception handler, line ~143. 
6. Stack trace shows:
```
   at System.ThrowHelper.ThrowInvalidOperationException_HandleIsNotInitialized()
   at System.Runtime.InteropServices.GCHandle.FromIntPtr(IntPtr value)
   at Microsoft.MixedReality.WebRTC.Interop.Utils.ReleaseWrapperRef(IntPtr wrapperRef) in C:\code\MixedReality-WebRTC\libs\Microsoft.MixedReality.WebRTC\Interop\InteropUtils.cs:line 135
   at Microsoft.MixedReality.WebRTC.DataChannel.DestroyNative() in C:\code\MixedReality-WebRTC\libs\Microsoft.MixedReality.WebRTC\DataChannel.cs:line 210
   at Microsoft.MixedReality.WebRTC.PeerConnection.RemoveDataChannel(DataChannel dataChannel) in C:\code\MixedReality-WebRTC\libs\Microsoft.MixedReality.WebRTC\PeerConnection.cs:line 1476
   at TestNetCoreConsole.Program.<Main>d__0.MoveNext() in C:\code\MixedReality-WebRTC\examples\TestNetCoreConsole\Program.cs:line 133
```

**Expected behavior**
Data channel connection removed and resources cleared from memory 

**Environment**
 - MR-WebRTC version (package version, or specific commit if building from source): master
 - Platform: Windows Desktop
 - Architecture: x64
 - Target device: Windows Desktop

**Additional context**
I noticed this issue when I was using data channels and wanted to remove them from the peer connection. It always crashed, thought I would try to reproduce it in a easier context.

[...]

PR #799, opened 2021-08-09
fix: do not attempt to release data channel memory twice

torepaulsson:
PR for issue described in https://github.com/microsoft/MixedReality-WebRTC/issues/722

The data channel memory was freed twice, once by the caller and then once again by an event registration, from what I can tell the event registration is there to handle when a remote peer removes a data channel.

Also added some code so the NamedPipeSignaler demo doesn't crash as much.

### DIFF
--- a/examples/TestNetCoreConsole/NamedPipeSignaler.cs
+++ b/examples/TestNetCoreConsole/NamedPipeSignaler.cs
@@ -2,10 +2,11 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Concurrent;
 using System.IO;
 using System.IO.Pipes;
+using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
 using Microsoft.MixedReality.WebRTC;
 
 namespace NamedPipeSignaler
@@ -19,7 +20,7 @@ namespace NamedPipeSignaler
     /// <remarks>
     /// This is a simple implementation for debugging and testing; this is not a production-ready solution.
     /// </remarks>
-    public class NamedPipeSignaler
+    public class NamedPipeSignaler : IDisposable
     {
         /// <summary>
         /// Peer connection this signaler is associated with.
@@ -83,7 +84,7 @@ namespace NamedPipeSignaler
         /// <summary>
         /// Thread-safe collection of outgoing message, with automatic blocking read.
         /// </summary>
-        private readonly BlockingCollection<string> _outgoingMessages = new BlockingCollection<string>(new ConcurrentQueue<string>());
+        private readonly BufferBlock<string> _outgoingMessages = new BufferBlock<string>();
 
         /// <summary>
         /// Construct a signaler instance for the given peer connection, with an explicit pipe name.
@@ -96,6 +97,11 @@ namespace NamedPipeSignaler
         {
             PeerConnection = peerConnection;
             _basePipeName = pipeName;
+
+            PeerConnection.DataChannelAdded += PeerConnection_DataChannelAdded;
+            PeerConnection.DataChannelRemoved += PeerConnection_DataChannelRemoved;
+            PeerConnection.LocalSdpReadytoSend += PeerConnection_LocalSdpReadytoSend;
+            PeerConnection.IceCandidateReadytoSend += PeerConnection_IceCandidateReadytoSend;
 
             // Try to create the server, if not already existing
             IsClient = false;
@@ -151,26 +157,8 @@ namespace NamedPipeSignaler
             _sendStream = new StreamWriter(_clientPipe);
             _sendStream.AutoFlush = true;
             _recvStream = new StreamReader(_serverPipe);
-            PeerConnection.LocalSdpReadytoSend += PeerConnection_LocalSdpReadytoSend;
-            PeerConnection.IceCandidateReadytoSend += PeerConnection_IceCandidateReadytoSend;
             _ = Task.Factory.StartNew(ProcessIncomingMessages, TaskCreationOptions.LongRunning);
-            _ = Task.Factory.StartNew(WriteOutgoingMessages, TaskCreationOptions.LongRunning);
-        }
-
-        /// <summary>
-        /// Stop the signaler background tasks, and dispose of all native resources.
-        /// </summary>
-        public void Stop()
-        {
-            _recvStream.Close();
-            _outgoingMessages.CompleteAdding();
-            _outgoingMessages.Dispose();
-            PeerConnection.LocalSdpReadytoSend -= PeerConnection_LocalSdpReadytoSend;
-            PeerConnection.IceCandidateReadytoSend -= PeerConnection_IceCandidateReadytoSend;
-            _sendStream.Dispose();
-            _recvStream.Dispose();
-            _clientPipe.Dispose();
-            _serverPipe.Dispose();
+            _ = Task.Run(() => WriteOutgoingMessagesAsync(CancellationToken.None));
         }
 
         /// <summary>
@@ -238,12 +226,14 @@ namespace NamedPipeSignaler
         /// Entry point for the writing task dequeuing outgoing messages and
         /// writing them to the sending pipe.
         /// </summary>
-        private void WriteOutgoingMessages()
+        private async Task WriteOutgoingMessagesAsync(CancellationToken ct)
         {
-            // GetConsumingEnumerable() will block when no message is available,
-            // until CompleteAdding() is called from Stop().
-            foreach (var msg in _outgoingMessages.GetConsumingEnumerable())
+            // OutputAvailableAsync() will wait until messages are available or
+            // until Complete() is called from Dispose().
+            while (await _outgoingMessages.OutputAvailableAsync(ct))
             {
+                var msg = await _outgoingMessages.ReceiveAsync(ct);
+
                 // Write the message and wait for the stream to be ready again
                 // for the next Write() call.
                 _sendStream.Write(msg);
@@ -262,7 +252,7 @@ namespace NamedPipeSignaler
                 // WebRTC signaler thread which is typically invoking this method through
                 // the PeerConnection signaling callbacks.
                 Console.WriteLine($"[->] {msg}");
-                _outgoingMessages.Add(msg);
+                _outgoingMessages.Post(msg);
             }
             catch (Exception e)
             {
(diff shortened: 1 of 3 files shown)


examples/TestNetCoreConsole/NamedPipeSignaler.cs
```diff
@@ -238,12 +248,14 @@ private void ProcessIncomingMessages()
...
         /// </summary>
-        private void WriteOutgoingMessages()
+        private async Task WriteOutgoingMessagesAsync(CancellationToken ct)
         {
-            // GetConsumingEnumerable() will block when no message is available,
-            // until CompleteAdding() is called from Stop().
-            foreach (var msg in _outgoingMessages.GetConsumingEnumerable())
+            while (await _outgoingMessages.OutputAvailableAsync(ct))
             {
+                // GetConsumingEnumerable() will wait until messages are available when no message is available,
```
djee-ms, line 255: Outdated comment? `GetConsumingEnumerable()` was removed, no?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1470,10 +1470,6 @@ public void RemoveDataChannel(DataChannel dataChannel)
                 Utils.ThrowOnErrorCode(res);
 
                 DataChannelRemoved?.Invoke(dataChannel);
```
djee-ms, line 1472: If we're relying on `PeerConnection_RemoveDataChannel` to always call internally the interop callback `PeerConnectionInterop.DataChannelRemovedCallback`, we're assuming `OnDataChannelRemoved()` is called which will raise the `DataChannelRemoved` event. So isn't raising the event here too a duplicate?

torepaulsson: Yes I agree, I wonder why it is done in this manner though. Could it be that the event should always fire independent if it is the peer or the remote that removes it, but that should already be hooked up from the interop callback PerConnectionInterop.DataChannelRemovedCallback?

I've tested some more with the TestNetCoreConsole, starting two instances and trying to close one connection. I added a handler for RemovedDataChannel which just logs to console. Seems like the remote does not notice that the data channels are removed. Seems like there are more problems here!

examples/TestNetCoreConsole/NamedPipeSignaler.cs
torepaulsson: Good catch! fixed in 64f463e65cf4ab4e6294e35fce2b2f0ee4602d44

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
torepaulsson: An update, my setup seems to not be working properly, the data channel never goes to Connected. Probably why the client side never knows about it... I will continue investigating tomorrow. It might be a better fix to remove the invokation of the event. That seems to be how the AddDataChannel method is implemented, it does not fire the add event right?

torepaulsson: Ok, I thought that the DataChannelRemoved event was going to fire when the data channel state changed to closed, but of course that is not the case, that would be up to the client to decide how to handle the data channels.

I'll continue tomorrow!

torepaulsson: Okay I need help. I'm unsure what this comment means, is it that the removed event is fired only on the local instance of the peer connection, or that it is fired also when the data channel is removed on the remote?
```csharp
        /// <summary>
        /// Event fired when a data channel is removed from the peer connection.
        /// This event is always fired, whatever its creation method (negotiated or not)
        /// and original creator (local or remote peer).
        /// </summary>
        public event DataChannelRemovedDelegate DataChannelRemoved;
```
There might be some hooks missing for this?

djee-ms: > I've tried moving the data channel open method and it seems that it works just when I have it after initialize, but not afterwards.

This is a known Google limitation (or maybe even from the WebRTC standard) that data channels need SRTP handshake during initial connection to secure them, and if you don't have any data channel before that initializing then the SRTP handshake is not performed. Later when you add a data channel, it can't be initialized because SRTP is not ready. So you always have to create at least 1 data channel _before_ you initialize, if you ever want to use _any_ data channel during your peer connection lifetime.

> Okay I need help. I'm unsure what this comment means

This is a comment on the C# code so only documents the intended behavior of the C# API itself, unrelated to how the internal underlying Google implementation works. Unfortunately that C++ low-level implementation (`libwebrtc`) is difficult to use and inconsistent at times. I cannot remember on top of my head the details, but I know some events are not fired consistently, so we had to manually craft some in MixedReality-WebRTC to make them more consistent to the end users.

The comment says that we aim at a behavior where the `DataChannelRemoved` event is raised when any data channel is removed, whether it was removed manually by the local peer connection (by an API call), or removed remotely by the other remote peer connection (via a message transmitted over the network). It also says that this behavior is to be consistent between data channels negotiated or not, that is either of:
- the data channels whose ID is pre-agreed by both peers and not transmitted over the network (they both create a data channel with the same ID, that should be available) -- those are out-of-band data channels because the ID is never exchanged, it's pre-agreed upon (_e.g._ "for my app, data channel ID 24 is the text message"). Note that here both peers need to create via API call the data channel for it to be useable.
- the data channels created by one peer without an explicit ID, and the implementation picked an ID and transmitted it over the network to the other peer that then reactively created a data channel with the same ID -- those are in-band data channels because the ID is sent in-band through the peer connection. Note that here only 1 peer creates the data channel via API call; the other peer will auto-create it internally after negotiation, and raise a `DataChannelAdded` event to the app.

All of that however is not very important, the point here is that `libwebrtc` only raise add/remove events in _some_ cases (in general when not from a user-initiated API call), but we aim at having a more consistent API in C# where events are always raised in all cases, so we manually craft the missing ones.

In short, this comment doesn't give you any info on what the internal implementation does, nor on what/where events should be raised from the C# API.

torepaulsson: Thanks for through explanation!

I dug a bit and found the peerconnectioninterface.h, as well as the datachannelinterface.h, as well as the documentation for webrtclib here https://w3c.github.io/webrtc-pc/#interface-definition.

It looks like there is no implementation for DataChannelRemoved which is called when a remote removes the data channel. Only the state of the data channel will be changed. So what I thought was a bug was just not supported, and doesn't need to be since you can check the data channel state instead.

This said. I've updated the code to take this into consideration. I've changed so we don't call the event like you commented, since it is called by the c wrapper. I've also removed the DataChannels.Remove(dataChannel) call in OnDataChannelRemoved, since that is already done in the remove method.

I will give this a try tomorrow!

torepaulsson: Ok, I've gone through this once again, when a PeerConnection is closed the data channels are actually removed, and the event is fired, so we need to have the DataChannels.Remove(dataChannel) call in the event handler. It will just return false if the data channel is already removed so no worries.

Can you have another look @djee-ms?

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
```diff
@@ -1468,12 +1468,6 @@ public void RemoveDataChannel(DataChannel dataChannel)
 
                 var res = PeerConnectionInterop.PeerConnection_RemoveDataChannel(_nativePeerhandle, dataChannel._nativeHandle);
                 Utils.ThrowOnErrorCode(res);
-
-                DataChannelRemoved?.Invoke(dataChannel);
-
-                // PeerConnection is owning the data channel, and all internal states have been
-                // updated and events triggered, so notify the data channel to clean its internal state.
-                dataChannel.DestroyNative();
```
torepaulsson, line 1476: I noticed this had some comments about PeerConnection owning this instance, but I get the feeling it is more owned by the peer connection interop, since that is the one that destroys the native implementation, and also actually triggers events that this PeerConnection class only wrapps in a way.

djee-ms: Yes. But the interop is kind of internal, so conceptually from the end-user point of view using the public API, the ownership is with the `PeerConnection`. But you're right, that derives from the fact the native data channel is owned by the native peer connection.

libs/Microsoft.MixedReality.WebRTC/PeerConnection.cs
djee-ms: I won't have time today sorry, and I'll be away next week unfortunately. Maybe if @fibann has some spare time, otherwise this will have to wait for my return.

torepaulsson: No worries, I can continue testing and working with packages built from this, so no stress!

torepaulsson: @djee-ms, I've used this for a while now, seems to be working fine! Hopefully you have some time to look at this! Always feels better to build a nugetpackage from main than a branch:)

djee-ms: Ack, and many thanks for your change. I'll try to find some time this week to have a look again at this PR. Sorry for the inconvenience, we're light on time allocated to this project.

examples/TestNetCoreConsole/Program.cs
```diff
@@ -123,7 +129,14 @@ static async Task Main(string[] args)
...
 
-                signaler.Stop();
+                Console.WriteLine("Removing data channels...");
+                foreach (var dataChannel in pc.DataChannels.ToImmutableArray())
+                {
+                    Console.WriteLine($"Removing DataChannel {dataChannel.Label}: State '{dataChannel.State}'");
+                    pc.RemoveDataChannel(dataChannel);
+                }
+
+                await Task.Delay(TimeSpan.FromSeconds(10));
```
djee-ms, line 139: Why the wait here?

djee-ms (commented):
Actually looks all good, now with very minimal change to the core library, which I think is correct. Can you just clarify the wait? After that I think we can merge. Thanks again for the contribution, this is much appreciated!!

examples/TestNetCoreConsole/Program.cs
torepaulsson: Seems like a big leftover actually. I believe it was added to give me some time to check that the data channels were removed before disposing the entire peer connection, but I'm not sure! I'll remove the wait

torepaulsson: made a fixup, just let me know and i'll rebase the commits so the fixup does not go on main
d328c3c5ce5d7ce04e3fe6e5ad670bfddafc52ad

djee-ms: I'll squash anyway, by policy. We rarely merge, unless there's a very compelling reason to keep the history.

djee-ms (approved):

